PR #1147, opened 2012-09-28
NDHotKeyEvent fixes

### LINKED ISSUE #1134: Switching keyboard layouts causes issues with . (text entry) functionality
I frequently switch between QWERTY and Dvorak keyboard layouts. If I launch QS in one layout and switch to another while it's running, the trigger for text entry (what should be the period key) gets stuck on whatever physical key used to be the period when I launched the program. 

**If launched in QWERTY then switched to Dvorak**
Attempting to type an 'e' starts the text entry field. Typing a period just types a period like any other character. 

**If launched in Dvorak then switched to QWERTY**
Attempting to type a 'v' starts the text entry field. Typing a period just types a period like any other character.

tiennou:
This imports NDHotKeyEvent as a submodule and fixes #1134 by listening for the correct notification. 

See https://github.com/nathanday/ndhotkeyevent/pull/2 for the NDHotKeyEvent diff, since it's now separate.


skurfer:
Any hope for this one?

tiennou:
Rebased.

From now on, I'll refer to my *Apple Extended AZERTY* physical keys like this `p`, because otherwise you get lost pwetty qwckliy.

- Switching keyboard layout works, shortcuts don't stop working — If you had ⌥⌘L in AZERTY, it's still the same in DVORAK (eg. the hotkey is "moved" to `⌥⌘P`).

- ب غخع سظهفزا 😆  If you switch to a non-latin based layout, the physical keys stay where they are بقخک  😆  w.r.t to the last selected layout (so AZERTY -> DVORAK -> Afghan Dari, shortcut is `⌥⌘p`, AZERTY -> Afghan Dari, shortcut is `⌥⌘L`. That one is debatable though.

- It looks broken for some keys. The thing is that IIRC `NDHotKeysEvent` stores keychars, not keycodes, so some keys (eg, I have ⌃⇧⎋ and ⌃⌥⇧$) turn into various other keys depending on the layout, and stop working until I get back to AZERTY.

skurfer:
I haven’t followed this one that closely, but I thought I would at least test it out to make sure everything still works as expected with US QWERTY.

Do we need the changes in the other repo to get merged, or is the submodule pointing to a point in history that already has those things?

The only problem I’ve run into so far is that a trigger bound to `⌃;` just beeps at me. Now that I look at the Console, it’s flooded with “enable failed”, so it probably has something to do with switching apps and checking scope.

tiennou:
The submodule points to somewhere sane (apart from the 3th bugpoint above, which I'm in the process of fixing).

It shouldn't be doing that, the point of that PR is to make Text Input changes preserve a shortcut's character — which currently works for every printable char, but not for the rest (3 above).

skurfer:
Before I get too far down this road, what should I be testing? The `rework` branch on our fork seems to have trouble watching anything, and the submodule appears to point to `master`.

So running the test app from master, I tried watching a couple of random keystrokes: ⌃I and ⌃L

They mostly kept up with layout changes, with one exception: When going from U.S. to Colemak, where the `I` lives on the physical `L` key, ⌃I stops working. If I switch to Dvorak first, then to Colemak, both keystrokes are recognized. My guess is that some code somewhere says “I don’t need to watch this `L` key any more”, but then nothing comes along and flips it back on when `L` becomes `I`.

tiennou:
Really ? `rework` is the branch that should fix the layout changing. `master` does not cope with that...

There might effectively be a "contention" issue, in that I don't think you can register 2 hotkeys at the same time, so depending on the order they were created, "I" tries to move to `L`, and fails because there's a hotkey there that's still active. It means I'll have to split the deactivate step from the actual reregister and reactivate dance I'm doing when the notification fires.

So, to be clear, `master` will not follow keyboard layout changes. `rework` should, but it you say it has trouble watching anything then that's a bug... Does it log an error when you register a hotkey, when you "enable all" ?

skurfer:
I’ve tried to test a bit more. Part of the problem is I’ve never used the Test App before, so I’m not always sure what’s normal and what’s broken. On `master`, I can mostly assign keys manually with the various “Hot Key…” buttons, but they don’t seem to do anything on `rework`. At best, they’ll say I’ve added `⌃A`, no matter what I was adding.

Anyway, the Add button seems to work on `rework`, which is probably the important thing for what we’re testing.

I think the short version is that `⌃I` is a good test case. I once again watched `⌃I` and `⌃L`. `L` seemed to keep working as I switched layouts, but `I` was not so lucky. Switching back to the default US layout doesn’t revive `⌃I`. It no longer appears with “List All”. What else should I try?

tiennou:
I made a short clip of me adding `⌘⌥M` and `⌘⎋`, then switching keyboard layouts from AZERTY to QWERTY to Dvorak. Hope this helps making sense of how the test app works :

https://www.dropbox.com/s/punldyzkrhgnvc3/HotKeys.mov?dl=0

tiennou:
I just pushed a fix for transient duplicates to rework, so you could test with `A`/`Q` while switching AZERTY/QWERTY.

skurfer:
OK, after a `git submodule update` on this branch, the test app seems to be able to follow ⌃I and ⌃L around through various layout changes, so I think you’ve got that part nailed down.

So now what? Should we put out a test build for some of the Dvorak people to poke at?

tiennou:
Glad to hear that. This means that our hotkey code should now work correctly — I'll do some tests tonight.

skurfer:
Going to do a pre-release soon. Did you get a chance to test any more? Should I include this?

skurfer:
I’ve been running this locally for a while. I haven’t seen any problems with functionality (though I also haven’t changed keyboard layouts at all), but I still see a large number of “enable failed” messages. Can you reproduce or do I need to dig into it?

skurfer:
Found one small problem. I had a trigger for ⌥⌘⇥ (switch to previous application, for those times when QS leaves focus on “nothing”) but it quit working. In the prefs, it shows as ⌥⌘0. I tried redefining the shortcut but it still shows 0 instead of Tab (and hitting ⌥⌘0 doesn’t fire the trigger).

tiennou:
Sorry, haven't had the time to look at this yet...

> a large number of “enable failed” messages.

Are you using scoped hotkeys somehow ? I think it's one of the few things that are QS-sugar-coating so I didn't really check how they would behave.

> it shows as ⌥⌘0

Was this after a layout change ? IIRC this can happen on "keyboard invariant" keys, if they aren't handled as such (NDHotKeyEvent checks what's the new key under the new layout, get back a non-ascii value, caps it to ASCII erroneously, then you get either 'A' or '0').

skurfer:
> Are you using scoped hotkeys somehow ?

Yes, I have a few triggers that are scoped to specific apps.

> > it shows as ⌥⌘0
> 
> Was this after a layout change ?

I haven’t changed layouts since September when I was testing this out. For what it’s worth, I’m back on `master` and the trigger now displays and works correctly. (As noted earlier, I tried redefining the shortcut, which didn’t seem to work, but based on the current state of things, it at least captured and stored “⇥” in the trigger.)

tiennou:
> Yes, I have a few triggers that are scoped to specific apps.

I'll have to check that then. I think that's the problem.

> the trigger now displays and works correctly.

Was it working as ⌥⌘⇥ at the time then, even tough it displayed as something else (my tests using ⌘⎋ had the same failure, shows as ⌘A/Q, but works as ⌘⎋).

skurfer:
> Was it working as ⌥⌘⇥ at the time then, even tough it displayed as something else

Nope. Just a beep, which is how I discovered the 0 in the preferences. Here are the relevant settings from `Triggers.plist`.

```plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>ID</key>
	<string>A3BA0483-792B-417F-B529-D78D3035706D</string>
	<key>characters</key>
	<string>	</string>
	<key>command</key>
	<dict>
		<key>actionID</key>
		<string>FileOpenAction</string>
		<key>directID</key>
		<string>QSPreviousApplicationProxy</string>
	</dict>
	<key>enabled</key>
	<true/>
	<key>keyCode</key>
	<integer>48</integer>
	<key>modifiers</key>
	<integer>1573160</integer>
	<key>name</key>
	<string>Focus Previous Application</string>
	<key>onPress</key>
	<false/>
	<key>onRelease</key>
	<true/>
	<key>type</key>
	<string>QSHotKeyTrigger</string>
</dict>
</plist>
```

hmelman:
> Are you using scoped hotkeys somehow ?

I wonder if it's possible to at a category in the left pane of the trigger preferences called Scoped that showed those triggers with scopes to make seeing this easier? Maybe also an Unscoped or Global category?

skurfer:
@tiennou via Slack:

> the version I pushed on the PR didn't have the submodule update I expected

So can we just point to a different commit and this will be good to merge?

tiennou:
It's not. I've tried building QS against it (since I was trying to make it work correctly "standalone" first), but now it's failing hard against the working version. Haven't had the time to take a closer look though...

pjrobertson:
Closing this in favour of #2594 

There's probably more 'useful' stuff in this PR, but let's close this and fix any outlying issues separately.

Hi @tiennou  :-)

Issue #1277, opened 2012-12-12
Recreating objects


skurfer:
If a `QSObject` isn't in the catalog, it doesn't exist after a restart. This has been a problem with triggers for a long time, and will soon be a problem for Synonyms.

I think in the majority of cases, the various object sources could probably recreate the object knowing only the identifier. So, what I propose is that in `+[QSObject objectWithIdentifier:]`, we continue to return an existing object if possible, but if it doesn't exist, do something like this:

``` objective-c
SEL recreate = @selector(recreateObjectWithIdentifier:);
id handler = [self handlerForSelector:selector];
if (handler) {
    return [handler recreateObjectWithIdentifier:anIdentifier];
}
```

And then of course we'd have to implement `recreateObjectWithIdentifier:` in various object classes, and make sure `objectWithIdentifier:` gets called at some point to trigger the behavior.

Places to test:
- [ ] Triggers made with internal objects (e.g. Pasteboard ⇥ Paste) should still work once the catalog entry's been disabled (see the 'Internal Objects' catalog entry). See #1373 for more info

pjrobertson:
This is a big job, but could it be a task for the sprint?

skurfer:
Maybe. I’ve already done it and still have the branch, but I’m sure it needs more work at this point.

If you could look over #1363 and see if you have any comments/concerns about my chosen implementation, that would help.

tiennou:
Wall of text incoming, apologies in advance. I'll comment here speaking about #1363 because let's say that this is the *meta issue* (*metapullrequest incoming* 😉).

> Rather than try to preserve objects in dictionary form, we just ask the class that created them in the first place to create them again.

I can think of a bunch of cases where the identifier* is not sufficient to recreate something. For example, menu items from the Accessibility plugin (which I think people use to rebind Triggers).

\* unless you craft it carefully, which I'm not fond of. That would need identifiers like bundle-id->menu->menu->... My personal point of view is that identifiers should be mandatory `NSUUID`s and be done with it. They don't convey any meaning, they're just *identifiers* used to check if we already have that object "in cache" (e.g. the Catalog + the set of runtime objects). With maybe a few wrinkles because filesystem objects must be kept fast.

All in all, I'd really much prefer the mechanism to use the object's `-dictionaryRepresentation` in some way, because that's the only general way I can think of that would work for any case.

[...]

PR #1363, opened 2013-01-31
recreate objects that aren't in the catalog (demo)

skurfer:
Here's a proof-of-concept for what I was proposing in #1277. Rather than try to preserve objects in dictionary form, we just ask the class that created them in the first place to create them again.
## Notes
- I don't really like that there's a class method on `QSObject` and an instance method on `QSObjectSource` with nearly identical signatures. They aren't exactly the same thing, but then again, the name makes sense for both. Suggestions welcome.
- I don't know of a specific style rule on this, but you'll see that the type is optional for `+[QSObject recreateObjectOfType:withIdentifier:]`. I can't think of another method where the first argument is optional, while later ones are required. So maybe it should change, but to my English speaking ears, it sounded better the way it is.
- We pass the object source both identifier _and_ type to make things easier on the object source. Some of them provide multiple types of objects and while you could probably sniff the identifier to figure out which you have, this is much cleaner.
- By not storing dictionaries, we don't have to worry about non-plist objects being present. This also means that if plug-ins change the metadata they store in objects, we don't care because they will always be freshly created, not pulled from an archive.
- Because this is implemented at the command level, it also allows you to use non-catalog objects in saved commands, droplets, etc. in addition to triggers.

I've implemented the "recreate" method in the filesystem object source and in the iTunes plug-in (see the [diff for the "resurrection" branch](https://github.com/quicksilver/iTunes-qsplugin/compare/resurrection)) for demonstration purposes.

(FYI, through some mysterious magic, triggers would already work for files not in the catalog. That is, `objectWithIdentier:` would always find it. But the same method in the synonyms code would come up empty. It was driving me crazy, but I quit trying to figure it out since my goal was to fix the situation across the board.)
## Things to try

Or things I tried anyway. All of the following used to stop working after a restart.
- Create a synonym for a file that isn't in the catalog.
- Create a trigger that sets an EQ preset for iTunes, then disable the EQ presets catalog entry.
- Create a trigger that uses an Album, Artist, or Track. (I just used "Select in Command Window" to keep it quiet.)

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject.h
@@ -105,6 +105,7 @@ extern NSSize QSMaxIconSize;
 
 + (id)objectWithName:(NSString *)aName;
 + (id)objectWithIdentifier:(NSString *)anIdentifier;
++ (id)recreateObjectOfType:(NSString *)aType withIdentifier:(NSString *)anIdentifier;
 + (id)makeObjectWithIdentifier:(NSString *)anIdentifier;
 + (id)objectByMergingObjects:(NSArray *)objects;
 + (id)objectByMergingObjects:(NSArray *)objects withObject:(QSObject *)object;
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -206,6 +206,24 @@ NSSize QSMaxIconSize;
 	return [objectDictionary objectForKey:anIdentifier];
 }
 
++ (id)recreateObjectOfType:(NSString *)aType withIdentifier:(NSString *)anIdentifier
+{
+    if (!anIdentifier) {
+        return nil;
+    }
+    if ([QSObject objectWithIdentifier:anIdentifier]) {
+        return [QSObject objectWithIdentifier:anIdentifier];
+    }
+    if (!aType) {
+        return nil;
+    }
+    id handler = [[QSReg objectHandlers] objectForKey:aType];
+    if (handler && [handler respondsToSelector:@selector(recreateObjectOfType:withIdentifier:)]) {
+        return [handler recreateObjectOfType:aType withIdentifier:anIdentifier];
+    }
+    return nil;
+}
+
 + (id)objectByMergingObjects:(NSArray *)objects withObject:(QSObject *)object {
 	if ([objects containsObject:object] || !object)
 		return [self objectByMergingObjects:objects];
--- a/Quicksilver/Code-QuickStepCore/QSCommand.m
+++ b/Quicksilver/Code-QuickStepCore/QSCommand.m
@@ -243,9 +243,10 @@ NSTimeInterval QSTimeIntervalForString(NSString *intervalString) {
         dObject = [newObject retain];
         
         id rep = [dObject identifier];
-        if(rep != nil)
+        if(rep != nil) {
             [[self commandDict] setObject:rep forKey:@"directID"];
-        else {
+            [[self commandDict] setObject:[dObject primaryType] forKey:@"directType"];
+        } else {
             rep = [dObject dictionaryRepresentation];
             if(rep)
                 [[self commandDict] setObject:rep forKey:@"directArchive"];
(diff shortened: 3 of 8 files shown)


Quicksilver/Code-QuickStepCore/QSObject.h
```diff
@@ -105,6 +105,7 @@ extern NSSize QSMaxIconSize;
 
 + (id)objectWithName:(NSString *)aName;
 + (id)objectWithIdentifier:(NSString *)anIdentifier;
++ (id)recreateObjectOfType:(NSString *)aType withIdentifier:(NSString *)anIdentifier;
```
pjrobertson: RE your naming convention, I don't like this because it jerks from the command after it. Something clearer could be:

```
+ (id)remakeObjectWithIdentifier:(NSString *)anIdentifier type:(NSString *)aType;
+ (id)makeObjectWithIdentifier:(NSString *)anIdentifier;
```

pjrobertson: @skurfer - this was more than 3 years ago, not sure if you'll still get a pig or not!?

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -206,6 +206,24 @@ + (id)objectWithIdentifier:(NSString *)anIdentifier {
...
+    if (!anIdentifier) {
+        return nil;
+    }
+    if ([QSObject objectWithIdentifier:anIdentifier]) {
+        return [QSObject objectWithIdentifier:anIdentifier];
+    }
+    if (!aType) {
+        return nil;
+    }
+    id handler = [[QSReg objectHandlers] objectForKey:aType];
```
pjrobertson: One question:

Will this `recreateObject...` method ever get called *before* QSReg is properly populated/plugins loaded?

As Etienne pointed out in the original issue thread. What happens if a plugin is disabled?

pjrobertson:
> I don't really like that there's a class method on QSObject and an instance method on QSObjectSource with nearly identical signatures. They aren't exactly the same thing, but then again, the name makes sense for both. Suggestions welcome.

I think your new method on `QSObject` just makes things more confusing. `makeObjectWithIdentifier:` already tries to look for existing objects (as `remakeObject...` does). I think the whole make/remake thing is confusing. Why does the developer need to know the difference? And then, should `makeObject...` try and call `remakeObject...` incase it's possible to recreate the object?

I'd say you're on the right track, but to simplify things I propose:

```objective-c
 + (id)makeObjectWithIdentifier:(NSString *)anIdentifier {
    return [self makeObjectWithIdentifier:(NSString *)anIdentifier type:nil];
}

 + (id)makeObjectWithIdentifier:(NSString *)anIdentifier type:(NSString *type) {
    id object = [self objectWithIdentifier:anIdentifier];
    
    if (!object && type != nil) {
        id handler = [[QSReg objectHandlers] objectForKey:aType];
        if (handler && [handler respondsToSelector:@selector(recreateObjectOfType:withIdentifier:)] {
            object = [handler recreateObjectOfType:aType withIdentifier:anIdentifier];
        }
    }
    if (!object) {
    		object = [[[self alloc] init] autorelease];
	 		[object setIdentifier:anIdentifier];
                 if (type) {
                       [object setPrimaryType:type];
                 }
    }
}
```

One other thing your code doesn't address (or perhaps it was by design): If `remakeObject...` turns a blank, then you're returning `nil` which means an empty object right? Is this better than creating a boring object with just an identifier (and possibly type), or is an object with identifier still more meaningful than `nil`?

pjrobertson:
So just looking at some of my triggers and thinking about my `nil` comment at the end.
Right now I have triggers for non existent files:
screen shot 2016-07-20 at 20 25 32

With your changes this'd be `nil` right? (goes and checks... actually `fileObjectWithPath` would never return `nil`, so in this case it's fine). But... I'm sure *some* providers may well return `nil` if they can't figure out how to make the object from the identifier. Or are you saying it's up to the provider to make sure `nil` is never returned? I think that's a bit risky, so it's better to keep it in `QSObject`

One thing to note:
Oh wait, thank god. I thought I'd uncovered a can of worms. You moved the object dictionary to QSLib in dbe8512c3181, so any new code you wrote now would have to go to QSLib to look for the object I guess. I thought for a minute we had two object dictionaries in QSLib **and** QSObject

skurfer:
> You moved the object dictionary to QSLib 

Yes, and for Christmas, I’d like to get rid of `+[QSObject makeObjectWithIdentifier:]` and use `+[QSObject objectWithIdentifier:]` to create (not fetch) objects. That will make it consistent with `objectWithName:`, `objectWithString:`, etc.

That would break almost every plug-in, I think.

tiennou:
My personal view on object initialization would be :
``` obj-c
QSObject *obj = [[QSObject alloc] init];
obj.identifier = @"whatever"; // It might even be set in init and not settable !
[obj setData:@"stuff" forType:QSSomeType];
[QSLib registerObject:obj]; // or stashing in in an array in the case of a scan
```

PR #1373, opened 2013-02-04
Attempt to add internal objs/messages to QS permanently

pjrobertson:
Internal Objs/messages currently only work if the catalog entries are enabled. Otherwise triggers etc. that use them just fail miserably.

This is an attempt to fix that, although it may not be the best. Open on the quicksilver repo for collab :)

`+[QSObject cleanObjectDictionary]` may well clear the items straight away, but testing seems to show that adding them to the catalog gives triggers etc. enough time to use the objects and hold on to them.

### DIFF
--- a/Quicksilver/Code-App/QSController.m
+++ b/Quicksilver/Code-App/QSController.m
@@ -899,8 +899,9 @@ static QSController *defaultController = nil;
 	if (DEBUG_STARTUP)
 		NSLog(@"Catalog loaded");
 #endif
-
+    
 	[QSObject purgeIdentifiers];
+    [QSObject addInternalObjectsToObjectDictionary];
 
 #ifndef DEBUG
 	if (newVersion) {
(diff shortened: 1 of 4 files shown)


skurfer:
Can we close this? Between #1648 and #1886, is the problem this was supposed to fix even there?

pjrobertson:
...no idea! Is what this was trying to fix actually fixed?
I'm kind of for the point that you shouldn't *have* to enable the catalog entry to use an object in a trigger.
If I want to add some file object that isn't in the catalog entry to a trigger, I can. If I want to use a URL that isn't in the catalog in a trigger, I can. If I wanted to use an iTunes artist in a trigger (but not keep it in the catalog, because that's too messy, then I *should* be able to)

Just checked - the answer is no. If I enable the 'internal objects' catalog, create a Trigger for 'Clipboard ⇥ Paste' then disable the catalog entry the object immediately disappears.

Again, this is probably an issue with how the object is stored in the trigger, which *might* be addressed in the "archiving objects" discussion. I agree this implementation is messy

pjrobertson:
Basically, the crux of it is that the object has an ID of `QSClipboardObject` and it can't be found/recreated with that ID ;-)

skurfer:
OK, then I think the general fix for #1277 should cover this, too.

pjrobertson:
OK, I've created a 'places to test' list on that issue to keep a note of this problem. Whenever we come across somewhere that breaks because of this we can add it to that list to test them all (and perhaps create unit tests) before we merge the final PR

PR #1395, opened 2013-02-17
Add an option to remove the last character only when pressing ⌫. Fixes #177

### LINKED ISSUE #177: Option for delete key to delete last character instead of entire entry

pjrobertson:
I know this has been a long sought after feature, especially for newcomers. tbh since we now have ⌃U for clearing the current pane I'll probably enable this as well.

Makes QS behave more like a normal search

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSPreferenceKeys.h
+++ b/Quicksilver/Code-QuickStepCore/QSPreferenceKeys.h
@@ -21,6 +21,7 @@ Following lines are no longer used in this project.
 #define kLastUsedLocation @"Last Used Location"
 
 #define kDoubleDeleteClearsObject @"Double Delete Clears Object"
+#define kDeleteRemovesLastCharacter @"deleteRemovesLastCharacter"
 #define kBrowseMode @"Browse Mode"
 #define kResultWindowBehavior @"Result Window Behavior"
 #define kSuppressHotKeysInCommand @"Suppress HotKeys in Command"
--- a/Quicksilver/Code-QuickStepCore/QSPreferenceKeys.h
+++ b/Quicksilver/Code-QuickStepCore/QSPreferenceKeys.h
@@ -20,7 +20,6 @@ Following lines are no longer used in this project.
 #define kLastUsedVersion @"Last Used Version"
 #define kLastUsedLocation @"Last Used Location"
 
-#define kDoubleDeleteClearsObject @"Double Delete Clears Object"
 #define kDeleteRemovesLastCharacter @"deleteRemovesLastCharacter"
 #define kBrowseMode @"Browse Mode"
 #define kResultWindowBehavior @"Result Window Behavior"
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
@@ -1407,8 +1407,17 @@ NSMutableDictionary *bindingsDict = nil;
 #pragma mark -
 #pragma mark NSResponder Key Bindings
 - (void)deleteBackward:(id)sender {
-    if(defaultBool(kDoubleDeleteClearsObject) && [self matchedString] == nil) {
-        
+    if (defaultBool(kDeleteRemovesLastCharacter) && [self partialString].length > 1) {
+        // reset the seaarch array (search the entire catalog)
+        [self setSearchArray:sourceArray];
+        validSearch = YES;
+        [[self partialString] deleteCharactersInRange:NSMakeRange(partialString.length-1, 1)];
+        [self partialStringChanged];
+        if (validMnemonic) {
+            // some objects found, change the colour of the results string
+            [resultController.searchStringField setTextColor:[NSColor blackColor]];
+        }
+    } else if(defaultBool(kDoubleDeleteClearsObject) && [self matchedString] == nil) {
         [super delete:sender];
     } else {
         [self clearSearch];
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
@@ -1417,7 +1417,7 @@ NSMutableDictionary *bindingsDict = nil;
             // some objects found, change the colour of the results string
             [resultController.searchStringField setTextColor:[NSColor blackColor]];
         }
-    } else if(defaultBool(kDoubleDeleteClearsObject) && [self matchedString] == nil) {
+    } else if([self matchedString] == nil) {
         [super delete:sender];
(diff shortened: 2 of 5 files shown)


pjrobertson:
Did we ever get to a consensus on this?

> I’m still opposed to resetting the search scope, but I no longer feel so strongly that I would refuse to merge something if it included that change.

What is your use case for clearing the object but *not* the search scope?

tiennou:
One use case I found is that, when you're testing things (like symlinks), it's nice to be able to preserve the scope while clearing out the search string ;-). I think it can be extended to pretty much every workflow where you browse a little, then try a search which gives nothing back (e.g. Keychain items).

skurfer:
What he said.

Think of a trigger like Search Artists. If you screw up, do you want to search the whole catalog, or just search for another artist? (Especially important for things that aren’t even in the catalog, like artists or the children of some other object.)

skurfer:
So… What are we doing with this guy?

pjrobertson:
...still debating about it in its entirety?! ;-)

I'm still inclined to say double delete clears the search field and single delete are still mutually exclusive. We don't want to alienate users too much with weird behaviour. I've just re-read this PR and the proposed solutions all seem so confusing. All I wanted to do was make things *easier* for people so searching works like it does anywhere else.

Here's what I propose (for the time being, I count hold ⌫ to clear as a 'new' feature and can come in another PR)

* Enable delete removes a single character by default for new users.
* Rename 'double delete clears the search pane' to 'delete clears the entire search string'. If you enable this you cannot use delete to remove a single character
* keep pressing delete when there is no search string clear the search object.

Whilst writing this comment I realised that double delete is actually quite a misnomer, if you look at this in a different way, you could say the only thing being changed is:

* If there's a search string, delete currently clears the whole thing
* I want to change it to delete clears on letter from the search string by default, leaving the other option.

The 'double delete' preference is a bit of a misnomer, since it's not actually the fact that you're pressing ⌫⌫ twice in succession that's causing the object to clear. It's just that the first delete clears the search string, the second the object (since there's no search string).
You can see how the 'double delete' name isn't quite right if you, for example:

* Search for 'Desktop' 
* Press ⌫ once (which clears the search string)
* Press ↓↑ etc to scroll through items (or even dismiss the interface)
* Press ⌫ again (which clears the item)

You're not pressing delete twice in succession, just clearing the search string at some point in time, then clearing the object

skurfer:
I find nothing to disagree with in all of that. Just one concern:

> Enable delete removes a single character by default for new users.

How will you pull that off? Existing users won’t have a preference set for the current behavior, because there is no such preference.

pjrobertson:
So the current (default) behaviour is to clear the search string on ⌫ and then pressing ⌫ does nothing. I propose making the default to clear one character, and then rename the 'Extras' preference from 'pressing the delete key twice clears the current object' (this will also be enabled by default) to 'pressing the delete key clears the whole search string'

skurfer:
> I propose making the default to clear one character

OK, so you meant for *all* users, not just new ones. I don’t know about that. Typically, when we’ve added a new preference for something, its default value preserves old behavior.

I think the single character option might be more intuitive for new users, as you say, and is a reasonable default for them, but changing it for *everyone*…?

pjrobertson:
> OK, so you meant for all users, not just new ones.

Yep. I was swayed towards this thinking from @tiennou ’s comment above:

> Also, I think this should default to enabled, with a hidden default to disable, because it's not a surprising behavior (as opposed to the double-delete), and users actually expect ⌫ to remove the last entered character.

Just because people might be used to one thing, doesn't mean it's right. I do understand/agree that there will be some users who see that and are like "hey, what happened, this is annoying!" but I'm sure there will also be users who see that and go "oh cool, delete finally does what I expect it to do. Great bug fix!"

For the first group of users, they can read the release notes and see where it stated "Add an option under the extras preferences to make delete clear the whole search string"

skurfer:
> Just because people might be used to one thing, doesn't mean it's right.

True. In this case it’s right because *it’s faster to just start over!* 😉

Anyway, if you’re willing to deal with the surprised and annoyed, I’m OK with the change. But can we please make reverting it a visible preference in the UI?

tiennou:
> But I really need a rationale for the double-delete removal, especially since I wrote it, I use it all the time, and I put it under Extras for a reason.

Ref: 414f626ecc0d9975ee97f8ec8b062dcc6b43a8b9, which sadly doesn't explain what was *my* rationale at the time 😨 ... I think it had something to do with not losing the currently-browsed-to location while looking for a file whose name I didn't exactly remember, but that's a shot in the dark.

hmelman:
I think I'm new to this. I have long been bothered that ⌫ in the first pane clears the entry rather than just delete the last character.
- How is ⌫⌫ supposed to work? What if I want to delete the last two characters?
- Yes ^U works to clear now (both the search text and the object), but my fingers haven't learned it (my fault). I find that ⎋ is much easier. ⎋ clears the entry in the first pane (though not the object which is fine) and ⎋⎋ clears the command window. 
- I think ⌫ should just delete the last character of search text. ^U and ⎋ and ⎋⎋ should behave as they currently do. ⌫⌫ should delete two characters. If you want a (hidden or not) pref to do something else with it, fine.

skurfer:
> I have long been bothered that ⌫ in the first pane clears the entry rather than just delete the last character.

Well, then you’re a *monster*! 😜 (See [my comment][rant] on the original issue.)

> How is ⌫⌫ supposed to work? What if I want to delete the last two characters?

With the proposed change, that’s what it will do (by default).

> ⎋ clears the entry in the first pane

Huh? <kbd>⎋</kbd> dismisses the entire interface, but leaves the search string and selection in place.

Your last bullet point is exactly what’s proposed. But to clarify, we don’t need to do any magic to capture ⌫⌫. We just need the preference to determine behavior for the *first* press of ⌫. If it erases a character, the next press will erase another character (until there are none). If it erases the entire search string, the next press will clear the selection from the pane (which is the current ⌫⌫ behavior when you enable that extras pref).

[rant]: https://github.com/quicksilver/Quicksilver/issues/177#issuecomment-235481728

skurfer:
Something to keep in mind @pjrobertson… Make sure ⌫ still works as expected when “managing” a collection.

hmelman:
> > ⎋ clears the entry in the first pane
> Huh? ⎋ dismisses the entire interface, but leaves the search string and selection in place.

I'm using Bezel and ⎋ in first pane, after typing, definitely leaves object shown and text shown (and match underlined) though it closes the results list. ⎋⎋ closes the command window entirely. I'd expect this to be the same in all interfaces.

QS 1.4.2 (401A) 10.11.6, 3.4 GHz i7 iMac, 24GB RAM, 1TB Fusion

PR #1761, opened 2014-01-30
Proof of concept showing simplicity of NSCoding

pjrobertson:
So yeah, I got annoyed with all the bugs we have saving/loading triggers and stuff, so thought I'd do a proof-of-concept on using `NSCoding`. Who'd have thought it would only take me 40 minutes, and result in fewer lines of code? (when we're properly done, we can remove all the stupid `dictionaryRepresentation` stuff, which will remove hundreds of lines)

With this branch, only triggers are archived at the moment. Of course the goal would be to archive everything we write to disk (shelves, QSCommands, catalogs) but this is just a proof of concept at the moment.
I haven't tested too much, but it seems to work nicely. The other side effect of this working is that quicksilver/elements.trigger.event-qsplugin#2 is now way less complicated (no need to use value transformers - we can just save QSObjects! :)
### The downsides...

The reason I never did this before was because it's completely backwards incompatible (without a lot of extra work on our part). After seeing quicksilver/elements.trigger.event-qsplugin#2 though and just how much sidestepping we've been doing over the years I got fed up and figured that backwards incompatibility isn't a problem. We just bump QS to v2.0 and be done with it.
So timescale... I'm kind of lost as to where v1.2.0-dev is, but does it include the UTI stuff? If so, then that's pretty backwards incompatible... so maybe we should just bump to v2.0 now?

**warning:** back up your `triggers.plist` before trying

### DIFF
--- a/Quicksilver/Code-App/QSTriggersPrefPane.m
+++ b/Quicksilver/Code-App/QSTriggersPrefPane.m
@@ -167,7 +167,7 @@
     
 	[triggerTable registerForDraggedTypes:[NSArray arrayWithObjects:NSFilenamesPboardType, QSTriggerDragType, nil]];
 	[triggerTable setVerticalMotionCanBeginDrag: TRUE];
-	[triggerTable setOutlineTableColumn:[triggerTable tableColumnWithIdentifier:@"command"]];
+	[triggerTable setOutlineTableColumn:[triggerTable tableColumnWithIdentifier:kCommand]];
 	[[[triggerTable tableColumnWithIdentifier:@"type"] dataCell] setArrowPosition:NSPopUpNoArrow];
 
 	NSColor *color = [triggerSetsTable backgroundColor];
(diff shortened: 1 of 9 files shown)


pjrobertson:
To solve the backwards-incompatible thing, we could just have a check on startup that looks at whether the trigger is in the old format. If so, then save a new file called "triggers-v1.plist" and then update the current triggers file

pjrobertson:
Closing this for now. I implemented NSCoding today. It doesn't do anything yet, but the next step would be to switch triggers to use that. To answer my 6 year younger self:

> To solve the backwards-incompatible thing, we could just have a check on startup that looks at whether the trigger is in the old format. If so, then save a new file called "triggers-v1.plist" and then update the current triggers file

No. What we should do is keep the triggers.plist the same, but make a *new* triggers-v2.plist. That won't break backwards compatibility. On startup, we check for triggers-v2.plist, if it doesn't exist, we read from triggers.plist using the obsolete method, then just save to triggers-v2.plist going forwards.

PR #1897, opened 2014-08-02
always resolve symlinks when loading children

skurfer:
Is there any reason not to do this? I can’t think of one.

(I’ll give it a couple of days, and if there’s no resolution, I’ll do another 1.2.0 pre-release without this.)


pjrobertson:
Does this still need to be merged, what with Etienne's changes?

skurfer:
Nope.

PR #1935, opened 2014-09-30
Add preference to restart automatically without crash reporter

clarkewd:
An update with just the hidden default per the conversation at https://github.com/quicksilver/Quicksilver/pull/1837


pjrobertson:
Thanks for this @clarkewd I've created a new pull request adding in Rob's suggestion but building on your work.
If you have any more things you want to change, then feel free to submit another pull request

PR #1943, opened 2014-10-09
Travis Unit Testing

pjrobertson:
This adds a `.travis.yml` file to the project which is needed by [Travis CI](https://travis-ci.org) to do automatic tests.

Note: The branch in this pull request is the #1939 branch with one extra commit to add a `.travis.yml` file. I'm only creating this PR to test Travis' testing skills :S

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,2 @@
 language: objective-c
-xcode_project: Quicksilver/Quicksilver.xcodeproj # path to your xcodeproj folder
-xcode_scheme: "Quicksilver Distribution"
+script: xctool -project Quicksilver/Quicksilver.xcodeproj -scheme "Quicksilver Distribution" test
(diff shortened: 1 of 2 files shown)


skurfer:
Just watched a presentation on Travis, which renewed my interest in this. Should it just be

```
language: objective-c
xcode_project: Quicksilver/Quicksilver.xcodeproj
xcode_scheme: Quicksilver Distribution
```

Not sure what to do about that test, though. Can we just disable it so it won’t get run by automatic tools, but a dev could still enable it and run it locally?

pjrobertson:
Yes, I think for that one test we’re going to have to disable it from Travis testing. I still can’t find any reliable method of working it.

My initial PR was for exactly what you’ve typed (look at the first commit). I changed things to try and get that silly unit test working.
I agree what you’ve put is best, if now we can just disable that test… :)

pjrobertson:
P.S. It should probably be part of the qsrelease tool to make sure the test *does* get run before QS is released

Perhaps a bit late to catch it, but at least it will be run!

PR #1975, opened 2014-11-03
Make the large type window mouse-selectable

### LINKED ISSUE #1903: Large Type can get stuck
If Large Type runs when the screen saver’s active (via an Event Trigger of Run After Delay), the text stays on screen until you restart Quicksilver.

pjrobertson:
Means it can be selected (and closed) if it was never made key (e.g. when opened during screensaver)
Fixes #1903

Also - various tidyups


pjrobertson:
No! I'll leave this open to remind me to fix it in a few days... :-)

PR #2005, opened 2014-12-17
URL Extractor updates

skurfer:
Nothing huge. I was writing a blog post for work on Beautiful Soup and was looking over this script, finding things I now don’t like. :smiley:

The only behavioral difference is that it now includes links to places on the same page (like `#this`). I can’t remember why they were ever skipped. Anyone?


pjrobertson:
This is just python and it's all pretty simple. I think this can be rebased and merged no problem.

skurfer:
Shouldn’t need a rebase, should it?

pjrobertson:
Nope, cool

PR #2012, opened 2015-01-02
only clean up short strings

skurfer:
I guess passing a full path as the image name was a valid use case after all. AppleScript actions get their icons this way.

In all the crash reports, there was never an attempt to get a range outside `{0, 10}`. If I understand what’s happening, only small strings can be stored as `NSTaggedPointerString`, so there’s no need to “clean” the longer ones.

An absolute path is likely to be longer, so this indirectly works. To be 100% correct, we’d need to test `fileExistsAtPath:`, but I don’t like the idea of going to the disk that frequently. Maybe `isAbsolutePath` would work, but are we certain that doesn’t touch the disk at all?

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSResourceManager.m
+++ b/Quicksilver/Code-QuickStepCore/QSResourceManager.m
@@ -64,7 +64,7 @@ QSResourceManager * QSRez;
 - (NSString *)safeImageName:(NSString *)rawName
 {
     // strings with a slash lead to a crash - issue #2002
-    if ([rawName containsString:@"/"]) {
+    if ([rawName length] <= 10 && [rawName containsString:@"/"]) {
         if (![slashNames containsObject:rawName]) {
             NSLog(@"sanitizing image name: %@", rawName);
             [slashNames addObject:rawName];


pjrobertson:
Sorry I've been useless at replying.

So - as we guessed this is a stupid OS X bug with NSTaggedPointer. Do we have any other solutions? Is it fixed in Yosemite 10.10.2? (If so I say we revert these changes and just forget about it)

Otherwise - how about my [previous idea](https://github.com/quicksilver/Quicksilver/pull/2003#issuecomment-66958753) of using `isAbsolutePath` and checking if an icon exists at that path?

P.S. [you're right](https://www.mikeash.com/pyblog/friday-qa-2012-07-27-lets-build-tagged-pointers.html) only small strings can be put in `NSTaggedPointerString`

skurfer:
> Is it fixed in Yosemite 10.10.2? (If so I say we revert these changes and just forget about it)

Well, good question. I can’t reproduce it any more, but I don’t see any mention of it in the release notes. It caused so many crashes, I’m scared to remove the fix, but I also hate the fix, so… remove it?

pjrobertson:
I'm in the process if upgrading to Yosemite (clean install). I'll give it a play without the fix and see if I get any crashes.
Wednesday, 04 February 2015, 11:56PM +0000 from Rob McBroom <notifications@github.com>:
>Is it fixed in Yosemite 10.10.2? (If so I say we revert these changes and just forget about it)
Well, good question. I can’t reproduce it any more, but I don’t see any mention of it in the release notes. It caused so many crashes, I’m scared to remove the fix, but I also hate the fix, so… remove it?

skurfer:
> I'm in the process if upgrading to Yosemite (clean install). I'll give it a play without the fix and see if I get any crashes.

Yeah, you could try just running 1.2.0. But I never once saw the crash personally, so I don’t know if that’s a good test. (I had to create a quick command-line project that just called `+[NSImage imageNamed:]` on a bunch of strings.

pjrobertson:
I’ve been running 1.2.1 all of today, and have only had one (unrelated) crash.
(correct me if I’m wrong, but the ‘bug' was fixed in 1.2.2 right?)

Do you still have your test project for reproducing the crash?

skurfer:
> the ‘bug' was fixed in 1.2.2 right?

Yeah.

> Do you still have your test project for reproducing the crash?

No, but it was just an array of strings and a `for` loop that called `imageNamed:` on each one.

I’m inclined to remove our work-around. If you agree, feel free to close this.

pjrobertson:
I’ve been trying to reproduce the bug on 10.10.2, but can’t so yeah - I saw we close this and revert the change :)

PR #2017, opened 2015-01-18
Updating menu bar icon for Yosemite

skeeJay:
Very simple PR that replaces two images so the menu bar icon fits visually in Yosemite.


HenningJ:

pjrobertson:
What's our practice on things like this in terms of which OS do we design for? I think these icons look cleaner anyway (on 10.9) so I say we should just go with them.

@skeeJay are all icons slightly taller (like these ones) on Yosemite?

skeeJay:
@pjrobertson yeah, in addition to having a thinner stroke, there's less whitespace around the icons; it's why the old icon has looked somewhat out of alignment will all the system icons. Have attached a screenshot with the new icon in context to illustrate.

screen shot 2015-01-18 at 4 56 32 pm

rprimus:
Does the new image also handle dark mode?

skeeJay:
@rprimus yes, Yosemite handles dark mode and pressed states from the same single image resource.

screen shot 2015-01-19 at 7 13 56 am

screen shot 2015-01-19 at 7 14 11 am

screen shot 2015-01-19 at 7 14 18 am

pjrobertson:
Alright, cool. Thanks for this :)

skurfer:
Yeah, been using this for a while and it looks good. Was going to merge it eventually. 😀

PR #2021, opened 2015-01-31
Use travis testing with QS

pjrobertson:
A rebuild of #1943

Basically the UI tests have been ignored via the use of an environment variable (hopefully!)

Ignore UI tests when running in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
 language: objective-c
-script: ./.travis_test.sh
-notifications:
-  email: false
+xcode_project: Quicksilver/Quicksilver.xcodeproj # path to your xcodeproj folder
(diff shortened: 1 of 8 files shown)


pjrobertson:
Alright. I've finally got this to work. In the end I had to write write a script that would just run the all tests *except* the Quicksilver Tests (UI tests)

It seems to work fine... finally!

skurfer:
Eh, personally, I think I’d rather get rid of that test so we can keep the config simple. :smiley:

Doing it this way, we’d have to maintain the shell script from time to time (if anyone even know or remembers to do it).

And we should set up a shared “quicksilver” account in Travis so we can all get in and adjust settings. (But that doesn’t have to hold up this PR.)

PR #2023, opened 2015-02-07
Draw a background on the text mode editor

### LINKED ISSUE #2004: Blurring on the re-entry of text
If you hit "." and enter plain text, the appearance of any text that is then entered is 'correct' (see, e.g., http://i.imgur.com/wBAe1xT.png). However, if you then hit tab, shift+tab, and ".", the text then becomes very subtly blurred (see, e.g., http://i.imgur.com/6UhKuDq.png).

This is on OS X 10.10.1, running on a MacBook Pro (Retina, 13", mid-2014).

pjrobertson:
Stops the text anti-aliasing from messing up.
Fixes #2004 

In the process I tidied things up a bit by moving lots of the setter methods to the view's `awakeFromNib` - instead of calling them every time the text view is set up. (may cause bugs, so look around!)

I can't for the life of me figure out how to fix another annoying thing - every time a letter is typed the whole text in the view moves ±1px up/down :(


skurfer:
I found at least one problem. 😀

screen shot 2015-02-08 at 12 35 11 pm

skurfer:
I don’t see this issue any more. I think it may have been fixed with the release of 10.11, or possibly by #2152. Close?

pjrobertson:
I haven't seen this, this branch is old and stale, so I agree. Should we keep the branch around just incase, so we don't lose the history?

skurfer:
> Should we keep the branch around just incase, so we don't lose the history?

I’d rather keep the list of mystery branches down. It doesn’t look like it would be that hard to recreate if we ever needed to, but I’ll leave it up to you.

pjrobertson:
>  It doesn’t look like it would be that hard to recreate if we ever needed to, but I’ll leave it up to you.

Just looked at the commit: true. Who knows if this is actually a 'fix' anyway?

Closed and deleted

Issue #683, opened 2012-02-05
Finder 'Sidebar Items' not being cataloged properly


pjrobertson:
If I have Catalog Pref Pane > Modules > Finder > Finder Sidebar Items checked, not all my sidebar seem to be indexed. Only 6 out of the eight.

Also, the 'Finder Sidebar Items' catalog entry doesn't have an icon

bfulop:
For me it's `favoriteitems.CustomListItems.*.Alias` that worked (March 2015) – it lists stuff that I put under the Favorites section in Finder.

PR #2038, opened 2015-03-12
Sidebar items

skurfer:
…thanks to a recent comment on #683.

Apple clearly can’t decide where they want to store these things.

Should we just get rid of the Toolbar Items preset? It doesn’t work either and I’m not sure if it’s even a thing you can do now.

### DIFF
--- a/Quicksilver/PlugIns-Main/Finder/Info.plist
+++ b/Quicksilver/PlugIns-Main/Finder/Info.plist
@@ -66,28 +66,6 @@
 					<key>source</key>
 					<string>QSDefaultsObjectSource</string>
 				</dict>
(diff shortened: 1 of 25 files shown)


pjrobertson:
I'm happy to just get rid of it - since this change would break things for those on older OSes right?

skurfer:
Yeah, I think so. If it ever worked. :wink:

skurfer:
In the future, if we just remove the English string, will Transifex remove it from the others on the next update?

pjrobertson:
I assume so. We don't ever 'push' the other languages to Transifex, only pull so I guess it'd delete them then..

PR #2039, opened 2015-03-12
Make sure directTypes *and* directFileTypes are filled for AS actions

pjrobertson:
This is a hack - the end goal is to not distinguish between the two.

You can test it with this AppleScript:

``` applescript
using terms from application "Quicksilver"
    on get direct types
        return {"public.directory"}
    end get direct types

    on open files _items_
        repeat with _item in _items_
            tell application "Finder" to reveal _item
        end repeat
    end open files
end using terms from
```

Verify that the AS only shows for folder types


pjrobertson:
This is not yet ready. The code fails for:

```
using terms from application "Quicksilver"
	on process text todo_text with todofile
		tell application "Quicksilver"
			show large type todo_text
		end tell
		--		do shell script "echo \"" & todo_text & "\" >> " & todofile
	end process text
	
	on get argument count
		(* Use "return 1" (or omit this entire handler) to never show the third pane. 
    Use "return 2" to force the third pane to show. 
    Use "return 3" to make the third pane optional. *)
		return 2
	end get argument count
	
	--This handler may be omitted if the action accepts all direct object types.
	on get direct types
		return {"NSStringPboardType"}
	end get direct types
	
	--This handler may be omitted if the action accepts all indirect object types.
	on get indirect types
		return {"NSFilenamesPboardType"}
	end get indirect types
	
end using terms from
```

because `NSStringPboardType` > `public.utf8-plain-text` > conforms to `public.item` and means text actions are classed as file actions

PR #2049, opened 2015-04-14
Cleanup catalog entry API

tiennou:
This is a rebase of #1601.

-  [x] Fix the auto-reenabling presets after a restart

### DIFF
--- a/Quicksilver/Code-App/QSCatalogPrefPane.m
+++ b/Quicksilver/Code-App/QSCatalogPrefPane.m
@@ -169,8 +169,8 @@ static id _sharedInstance;
 	[[[itemContentsTable tableColumnWithIdentifier:kItemPath] dataCell] setFont:[NSFont labelFontOfSize:9]];
 	[[[itemContentsTable tableColumnWithIdentifier:kItemPath] dataCell] setWraps:NO];
 
-	[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(catalogChanged:) name:QSCatalogEntryChanged object:nil];
-	[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(catalogIndexed:) name:QSCatalogEntryIndexed object:nil];
+	[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(catalogChanged:) name:QSCatalogEntryChangedNotification object:nil];
+	[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(catalogIndexed:) name:QSCatalogEntryIndexedNotification object:nil];
 	[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(updateEntrySelection) name:NSOutlineViewSelectionDidChangeNotification object:nil];
 
 	[itemTable reloadData];
@@ -284,7 +284,7 @@ static id _sharedInstance;
 		[self showOptionsDrawer];
 
 	[[NSNotificationCenter defaultCenter] postNotificationName:QSCatalogStructureChanged object:nil];
-	[[NSNotificationCenter defaultCenter] postNotificationName:QSCatalogEntryChanged object:childEntry];
+	[[NSNotificationCenter defaultCenter] postNotificationName:QSCatalogEntryChangedNotification object:childEntry];
 }
 
 - (IBAction)saveItem:(id)sender {
@@ -574,7 +574,7 @@ static id _sharedInstance;
 }
 
 - (IBAction)applySettings:(id)sender {
-	[[NSNotificationCenter defaultCenter] postNotificationName:QSCatalogEntryChanged object:nil];
+	[[NSNotificationCenter defaultCenter] postNotificationName:QSCatalogEntryChangedNotification object:nil];
 	[(QSController *)[NSApp delegate] rescanItems:sender];
 }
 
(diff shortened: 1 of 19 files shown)


skurfer:
Awesome. I guess that checkbox is there because you haven’t fixed it yet?

tiennou:
Yeah, now it is checked ! I also fixed one of your... own ;-).

tiennou:
Commit Of The Day : 1e7a5d4.

Take a good look at it, and tell me that's the only occurrence :-S. I literally ran QS 3 times in a row just while intently looking at this dealloc method before The Big Picture struck me.

skurfer:
Other things I found going through my comments. If you don’t want to deal with these, I can do another PR after this is merged.

- [ ] get rid of `shouldIndex` and bind NIB to `canBeIndexed`
- [ ] replace `canBeIndexed` checkbox with a label that's invisible when false
- [ ] rename “Include in Global Catalog” to “Entry Enabled”
- [ ] call `[loopItem scanForced:YES]` in `-[QSLibrarian reloadSource:]`

skurfer:
> Take a good look at it, and tell me that's the only occurrence :-S.

Heh. Nice. I haven’t checked, but it wouldn’t surprise me if the registry, plug-in manager, trigger manager, etc. all had a pointless `dealloc`.

tiennou:
I'll try to take a look at those. Since I'm in the process of rebasing the bunch of stuff I have lying around that, what would be the best way to go ? I might have a few ideas on how to calm down that -writeCatalog: method, as well as a pending Librarian rewrite I'm currently rebasing. I suppose this one PR is *mostly tested*, so I'll do another for the Lib stuff, and then continue piling on this one for more fixes. Does that suit you ?

tiennou:
I'm not sure about some of your points above so let's discuss them ;-).
> get rid of `shouldIndex` and bind NIB to `canBeIndexed`

`shouldIndex` and `canBeIndexed` seem to have different semantics (the first is plain-ol' `enabled` (eg "should be scanned", the other is relative to whether an entry can get its index saved).

> replace `canBeIndexed` checkbox with a label that's invisible when false

As per above, I don't think it means what you think it means ;-).

> rename “Include in Global Catalog” to “Entry Enabled”

I'm mostly against... for now. Since I plan on inflicting the same kind of heavy rewrite to the Librarian, I had a plan to maybe handle those inner... catalogs or something. I guess we'll see what comes out.

tiennou:
Also — because my eyes hurt of looking at that messy indentation and I already stepped all over that file — what was the project's Most Preferable Indentation Character already ? ;-)

skurfer:
> what was the project's Most Preferable Indentation Character already ?

4 spaces, which you will helpfully find **not** documented at http://qsapp.com/wiki/Github 😐

skurfer:
> I'm not sure about some of your points above so let's discuss them ;-).

Neither am I. :-)

It's been a long time. I was just looking at comments, but don't remember the reasons behind them all. I'll review and comment here.

> `shouldIndex` and `canBeIndexed` seem to have different semantics (the first is plain-ol' `enabled` (eg "should be scanned", the other is relative to whether an entry can get its index saved).

Yes, they are different, which is why we should use `canBeIndexed` instead.

The checkbox control on the Attributes tab labelled "Maintain an index of contents" (in the English translation) is bound to `shouldIndex` which, as you said, is just `isEnabled` so it's always exactly the same value as the checkbox above. I think “Maintain an index…” should be bound to `canBeIndexed` instead.

> > replace `canBeIndexed` checkbox with a label that's invisible when false
> 
> As per above, I don't think it means what you think it means ;-).

Referring to the same “Maintain an index” checkbox… Since it’s **always** disabled, why does it need to be a checkbox at all. Why not make it a label that says "Indexed" and make it visible when `canBeIndexed` is true?

> > rename “Include in Global Catalog” to “Entry Enabled”
> 
> I'm mostly against... for now. Since I plan on inflicting the same kind of heavy rewrite to the Librarian, I had a plan to maybe handle those inner... catalogs or something. I guess we'll see what comes out.

OK. We can wait on that.

skurfer:
> I suppose this one PR is *mostly* tested, so I'll do another for the Lib stuff, and then continue piling on this one for more fixes. Does that suit you ?

Yeah, that sounds good.

tiennou:
Again, I don't think `canBeIndexed` means what you think. It's a (optional) property of an object source that controls whether that source supports being serialized to a qsindex file.

Now that I reread your 2 points, I understand they aren't actually 2 points : you want a way to tell if a given entry supports being indexed, which actually makes sense. I'm not too fond of that, as it gives insight into something almost no user will care about.

skurfer:
> Again, I don't think `canBeIndexed` means what you think. It's a (optional) property of an object source that controls whether that source supports being serialized to a qsindex file.

I know what it is. Just look at the iTunes plug-in. :stuck_out_tongue:

> Now that I reread your 2 points, I understand they aren't actually 2 points : you want a way to tell if a given entry supports being indexed, which actually makes sense. I'm not too fond of that, as it gives insight into something almost no user will care about.

I had just noticed that the two checkboxes ultimately showed the same value. What I want in the end is to stop doing that. I tried to figure out what the checkboxes’ original intent was and change them to reflect that, but getting rid of one entirely would also work. I agree that knowing if something supports indexing provides no practical benefit to users.

skurfer:
Are you planning to remove the “Maintain an index of contents” checkbox here? If so, I’ll wait. Otherwise, I think this is good to merge.

PR #2070, opened 2015-07-03
use standard OS X term for keyboard shortcuts

skurfer:
I tried to get all the user-visible strings and left the rest alone. Let me know if I missed anything.


pjrobertson:
A pretty big change for QS veterans! Was there something that prompted you to do this?
I agree that this is probably best, but I'm worried a backlash from existing users might make it more of a headache than we'd hope...!

skurfer:
> Was there something that prompted you to do this?

It was always wrong. Don't you think? I just accepted it, but @hmelman called us out in it.

> I agree that this is probably best, but I'm worried a backlash from existing users might make it more of a headache than we'd hope...!

I hope they have bigger priorities than this. 😀 And they're as able to submit pull requests as we are.

pjrobertson:
> It was always wrong. Don't you think?

Yeah you're definitely right, but I just kinda think 'hotkey' has become a Quicksilver standard. Let's see how this goes

Do any of the locale files need changing? Do we need to update anything on Transifex?
If not, go ahead and merge :)

skurfer:
> Do any of the locale files need changing?

I left the left-hand side the same, so a lot of those still say HotKey, but the user visible right-hand side is changed in the English translation. There are a few localizations that are just using the English version. Should we change the user-visible strings there, too? Should I change the left-hand side? Nothing will break if we do it everywhere, right?

> Do we need to update anything on Transifex?

I don’t think so, since the “keys” used to look up translations haven’t changed. Though if it shows the English version as a guide, we should probably push the latest up there. Or if we change the keys like I asked about above.

skurfer:
Going ahead with this, since the changes in `QSSearchPrefPane.xib` are holding up something else. We can always tweak it later.

Issue #925, opened 2012-06-08
"Smarter" Spacebar behavior


sinewave:
I think it would be really awesome if you could add another option for a "smarter" search bar behavior to the preferences. A lot times i find myself wishing i could combine all the other options instead of having to choose between them. What i would suggest is for the behavior to be context sensitive based on the current selection in the first pane. 

For example pressing the spacebar would perform the following actions in order:

folders/files         - navigate into folder or file (ie show apps recent documents, txt files contents, etc..)
urls                       - jump to argument field 
files                       - quickview for any file that you can't typically navigate into

Other definitions could be made for other file types with the default behavior being show contents.  

It might sound a little confusing on paper but coming for Launchbar which has a similar behavior it actually feels *very* intuitive and definitely subscribes to the quicksilver philosophy :)

skurfer:
This was implemented in #1803 and #2072

Issue #1766, opened 2014-02-04
Asynchronous object sources


skurfer:
Currently, Quicksilver calls an object source’s `objectsForEntry:` and expects it to return results. This doesn’t work well with asynchronous APIs like Spotlight and many HTTP-based services.

It should also be possible for Quicksilver to tell an object source to _start_ scanning, but not expect results. The source will be responsible for setting the results when they become available and informing Quicksilver that the scan is complete.

This was attempted in #1455, but the world wasn’t ready. :smiley:

skurfer:
@tiennou said:
> I'm making @protocol QSObjectSource

Nice.

> fixing the delegate methods to pass entries instead of dictionaries (that's the breaking change)

That was my approach in #1455. It’s much more useful that way. And that would allow object sources to rescan an entry directly without bothering with notifications.

> and I found no easy way of knowing beforehand if an entry expects a dict or a real entry, which is why it's breaking.

My solution was to create a different method that could live along-side `objectsForEntry:` (called `loadObjectsForEntry:`). If you did that, there’s your way to test for compatibility.

tiennou:
> My solution was to create a different method that could live along-side `objectsForEntry:` (called `loadObjectsForEntry:`). If you did that, there’s your way to test for compatibility.

That's as far as I went (renaming `entryCanBeIndexed:` instead). Except I can't add methods to `QSObjectSource` (the class) without it trickling down to all subclasses, thus breaking my `respondToSelector:` test...

> That was my approach in #1455. It’s much more useful that way. And that would allow object sources to rescan an entry directly without bothering with notifications.

Thanks for the refresher ! I'm impressed at how much we already rebuilt with words actually ;-). I don't think I'm going to tackle that right now tough, because I think I'll first make a go at cleaning up QSLib, the scanning process and the object gathering, because that's what I was aiming at when I started. Now I'm getting grumpy with all those mismatched notification names :-S.

What I was thinking was to use the entry's `-contents` as the canonical source of objects (instead of the objectDictionary currently in QSLib), and just use observation to keep track of changes to those contents — so async loading would be handled automatically. But *then*, we would need async ranking ;-).

PR #2072, opened 2015-07-08
allow type-specific smart spacebar behavior

skurfer:
A small enhancement to address some of the concerns raised in the discussion on #925.

Basically, for any type in `QSTypeDefinitions`, in addition to the name and icon, you can also specify an integer for one of the existing spacebar behaviors.

This made the special-case code for URLs unnecessary, so I removed it. I’ve added the necessary entry for URLs to the plist in the core plug-in. If this is merged, I’ll also add something for search URLs to the web search plug-ins, and iTunes tracks in that plug-in.

I’ll document the new key and its possible values in the plug-in ref, too.

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
@@ -32,6 +32,7 @@
 #define pUserKeyBindingsPath QSApplicationSupportSubPath(@"KeyBindings.qskeys", NO)
 #define MAX_HISTORY_COUNT 20
 #define SEARCH_RESULT_DELAY 0.05f
+#define kQSSmartSpace @"smartspace"
 
 NSMutableDictionary *bindingsDict = nil;
 
@@ -1337,7 +1338,7 @@ NSMutableDictionary *bindingsDict = nil;
     QSAction *action = [[self actionSelector] objectValue];
     if (behavior == 7) {
         // override smart defaults with type-specific behavior (if defined)
-        NSNumber *typeBehavior = [[[QSReg tableNamed:@"QSTypeDefinitions"] objectForKey:[newSelectedObject primaryType]] objectForKey:@"smartspace"];
+        NSNumber *typeBehavior = [[[QSReg tableNamed:@"QSTypeDefinitions"] objectForKey:[newSelectedObject primaryType]] objectForKey:kQSSmartSpace];
         if (typeBehavior) {
             behavior = [typeBehavior integerValue];
         }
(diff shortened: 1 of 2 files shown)


pjrobertson:
All pretty simple and nice and clean. Go ahead and merge once you've added the consts (if you want to)

skurfer:
Thanks for the quick merge. I’ve updated the plug-in ref and put out a new Web Search plug-in. No one will see it because it requires build 4012. I did that because it also contains an incompatible change to work with @tiennou’s catalog entry updates (that I wasn’t aware of until I tried to build the plug-in).

I’ve also updated iTunes, but there are more changes planned there, so I haven’t released it.

tiennou:
About that, I'm currently cleaning up `QSObjectSource`, so there will be movement in that section (like backward-incompatible movement). I'm trying to keep the incompatible changes in a branch on top of my cleanup so we can decide when we want all plugins to be rebuild (unless I find a workaround that allows me to detect new-style source vs the current ones, which I tried a few times and failed at).

skurfer:
>  I'm currently cleaning up `QSObjectSource`

And doing something about #1766, I hope. :wink:

> I'm trying to keep the incompatible changes in a branch on top of my cleanup so we can decide when we want all plugins to be rebuild

Sounds good. I’ve got a list of plug-ins that were using `QSCatalogEntryChanged` and will need to be updated for the next release. Do you know of any other changes we need to make related to the catalog entry clean-up?

> (unless I find a workaround that allows me to detect new-style source vs the current ones, which I tried a few times and failed at)

Are there any methods being added/renamed/removed? If so, `respondsToSelector` seems like a good candidate. But if upgrading the plug-ins won’t be too bad, I’d rather just do that instead of having a bunch of branching code in the core app to support legacy stuff.

tiennou:
> And doing something about #1766, I hope. :wink:

In fact, there's already the beginning of this, in the form of `-enableEntry:/disableEntry:`. It's just that the rest of the code expect notifications, and will brainlessly reload the entry's contents. I'll see what can be done about that, because it's sorely needed.

>> I'm trying to keep the incompatible changes in a branch on top of my cleanup so we can decide when we want all plugins to be rebuild

> Sounds good. I’ve got a list of plug-ins that were using `QSCatalogEntryChanged` and will need to be updated for the next release. Do you know of any other changes we need to make related to the catalog entry clean-up?

>> (unless I find a workaround that allows me to detect new-style source vs the current ones, which I tried a few times and failed at)

> Are there any methods being added/renamed/removed? If so, `respondsToSelector` seems like a good candidate. But if upgrading the plug-ins won’t be too bad, I’d rather just do that instead of having a bunch of branching code in the core app to support legacy stuff.

I'm making `@protocol QSObjectSource`, fixing the delegate methods to pass entries instead of dictionaries (that's the breaking change), and I found no easy way of knowing beforehand if an entry expects a dict or a real entry, which is why it's breaking. The ugly way was adding a -(BOOL)newStuff; method in the protocol that returns `NO` and that the entry can use to know what to pass to its source (new source would override it to `YES`). But then we'd have to cleanup that sometime later, and given the `-selection/currentEntry` mess there's already in I'd be in favor of just breaking it ;-).

tiennou:
(Bonus question. What's the difference between `-lastScanDate` and an entry's indexation date ? What happens to non-indexable entries ? :weary:)

skurfer:
Moving this discussion to #1766

PR #2073, opened 2015-07-08
update the results window to use AutoLayout

skurfer:
A redo of #1924 

I wanted to get it out the door before it was completely forgotten.

### DIFF
--- a/Quicksilver/Code-App/QSMainPreferencePanes.h
+++ b/Quicksilver/Code-App/QSMainPreferencePanes.h
@@ -12,8 +12,6 @@
 @interface QSSearchPrefPane : QSPreferencePane {
     IBOutlet NSPopUpButton *keyboardPopUp;
 }
-- (BOOL)showChildrenInSplitView;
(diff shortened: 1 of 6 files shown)


pjrobertson:
So we're totally getting rid of the child results view? I like it :-)

I'm assuming you've been running with this for a while, and you've tested it on El Capitan? How about 10.9 Mavericks? It looks great for me on 10.10

skurfer:
I can try on El Capitan. I seem to have ditched my Mavericks VM, but I can get @claytron to try it out.

skurfer:
Looks good on El Capitan.

pjrobertson:
I actually think I have a Mavericks VM. I can test it when I get hold of my
hard drive - it's in your car ;-)

pjrobertson:
There are quite a few problems with AutoLayout. I looked into fixing it (mainly the resize button) but it was all a mess. Should we just stick with the old way and scrap AutoLayout (but still remove the child split view)?

PR #2075, opened 2015-07-09
Deprecation cleanup

tiennou:
Here's an attempt at bringing down the count of warnings. I even *tried* to kill the ND classes ;-).

### DIFF
--- a/Quicksilver/Code-App/QSSimpleWebWindowController.m
+++ b/Quicksilver/Code-App/QSSimpleWebWindowController.m
@@ -8,9 +8,11 @@
 
 #import "QSSimpleWebWindowController.h"
 #import "QSWindow.h"
-
-
 #import <WebKit/WebKit.h>
+
+@interface QSSimpleWebWindowController () <NSFileManagerDelegate>
+@end
+
 @implementation QSSimpleWebWindowController
 
 - (id)initWithWindow:(id)window {
--- a/Quicksilver/Code-External/CGSPrivate.h
+++ b/Quicksilver/Code-External/CGSPrivate.h
@@ -101,7 +101,8 @@ typedef enum {
 	CGSWarpFade,	// Warp old and fade out revealing new.
 	CGSSwap,		// Swap desktops over graphically.
 	CGSCube,		// The well-known cube effect.
-	CGSWarpSwitch   // Warp old, switch and un-warp.
+	CGSWarpSwitch,	// Warp old, switch and un-warp.
+	CGSFlip,
 } CGSTransitionType;
 
 typedef enum {
@@ -221,3 +222,25 @@ extern OSStatus CGSConnectionGetPID(const CGSConnection cid, pid_t *pid, const C
 extern CGSValue CGSCreateCStringNoCopy(const char *str);
 extern char* CGSCStringValue(CGSValue string);
 extern int CGSIntegerValue(CGSValue intVal);
+
+//#define CGSTransparentBackgroundMask (1 << 7)
+
+// Window wrapping
+typedef struct CGPointWarp CGPointWarp;
+
+struct CGPointWarp {
+	CGPoint local;
+	CGPoint global;
+};
+
+extern CGError CGSSetWindowWarp(const CGSConnection cid, const CGSWindow wid, uint32_t w, uint32_t h, CGPointWarp *mesh);
+extern OSStatus CGSSetWindowListBrightness(const CGSConnection cid, CGSWindow *wids, float *brightness, int count);
+
+// Global Hotkeys
+typedef enum {
+	CGSGlobalHotKeyEnable = 0,
+	CGSGlobalHotKeyDisable = 1,
+} CGSGlobalHotKeyOperatingMode;
+
+extern CGError CGSGetGlobalHotKeyOperatingMode(CGSConnection connection, CGSGlobalHotKeyOperatingMode *mode);
+extern CGError CGSSetGlobalHotKeyOperatingMode(CGSConnection connection, CGSGlobalHotKeyOperatingMode mode);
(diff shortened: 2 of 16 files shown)


pjrobertson:
Nice looking cleanup here. How did you update CGSPrivate? Did you manually do it yourself, or copy/paste from an updated copy online?
I've tried to do it countless times in the past, but it's always caused crashes (e.g. with things like the cube animation effect etc.)

pjrobertson:
> e.g. with things like the cube animation effect etc.

Actually, since we've killed the cube interface and we've removed the effect from the setup screen it shouldn't be a problem. I'm running with this now

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSDirectoryParser.m
```diff
@@ -49,31 +46,28 @@ - (NSArray *)objectsFromPath:(NSString *)path depth:(NSInteger)depth types:(NSAr
...
-            if (targetFile) {
-                aliasSource = [NDAlias aliasWithContentsOfFile:file];
-                aliasFile = file;
-                file = targetFile;
-                [manager fileExistsAtPath:file isDirectory:&isDirectory];
-                [[NSURL fileURLWithPath:file] getResourceValue:&type forKey:NSURLTypeIdentifierKey error:nil
-                ];
+            NSURL *targetURL = [NSURL URLByResolvingBookmarkAtURL:currentURL options:NSURLBookmarkResolutionWithoutUI bookmarkDataIsStale:NULL error:&err];
+            if (!targetURL) {
+                NSLog(@"Error resolving alias: %@", err);
```
skurfer: I’m seeing this logged hundreds of times. Most of them say because the thing is a directory, but testing `NSURLIsDirectoryKey` first doesn’t seem to return `YES` so I’m not sure what the issue is.

skurfer:
Besides the comment on the diff, I also saw at least one double free after running this for a while. I didn’t have time to find a way to reproduce, but if I do, I’ll let you know.

tiennou:
@pjrobertson I copied the "new" BSD-licenced version and readded the missing things on top of it.

@skurfer I'll take another look at that. To be frank, I was trying to kill `NDAlias` because of the deprecation warnings so I didn't run much. I'm also concerned that, since aliases are (to me) opaque OS X things, Apple could have changed something between the old Carbon API (used by NDAlias) and the new NSURL bookmarks. And then, maybe thinking some more, maybe those pre-OS X days are sufficiently behind us that we shouldn't care too much ;-).

tiennou:
Rebased against master, without the switch-to-NSURL-bookmarks thing in because @skurfer said so (sorry your comment got lost since I rebased...). We'll change to bookmarks later, so at least this can clear the warning queue a little...

pjrobertson:
>without the switch-to-NSURL-bookmarks thing 

Remind me, do you mean the removal of NDAlias and stuff?

tiennou:
That's something else, but those NSURL changes are required before we can drop NDAlias. Agreed, those are just one-step methods around the 2-step API (make bookmark+write data / read data+resolve bookmark). I haven't been able to see the hundreds of logs @skurfer reports seeing, but I don't have that many aliases around.

skurfer:
I ran with this merged in for a while. It was mostly fine, but I saw the same crash about once per day when QS was in the background.

    Crashed Thread:        6  Dispatch queue: com.apple.root.utility-qos
    
    Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
    Exception Codes:       KERN_INVALID_ADDRESS at 0x0000248c0d31bec0
    
    VM Regions Near 0x248c0d31bec0:
        Memory Tag 250         0000000118752000-0000000118b52000 [ 4096K] r--/r-- SM=COW  
    --> 
        MALLOC_NANO            0000600000000000-0000600000400000 [ 4096K] rw-/rwx SM=COW  
    
    Application Specific Information:
    objc_msgSend() selector name: retain
    
    …snip…
    
    Thread 6 Crashed:: Dispatch queue: com.apple.root.utility-qos
    0   libobjc.A.dylib               	0x00007fff8518f0dd objc_msgSend + 29
    1   com.apple.Foundation          	0x00007fff8629d9ff NSKeyValueWillChange + 465
    2   com.apple.Foundation          	0x00007fff86261801 -[NSObject(NSKeyValueObserverNotification) willChangeValueForKey:] + 216
    3   com.apple.Foundation          	0x00007fff863055be _NSSetDoubleValueAndNotify + 235
    4   com.blacktree.QSCore          	0x000000010863440c __42-[QSLibrarian scanCatalogIgnoringIndexes:]_block_invoke + 473
    5   libdispatch.dylib             	0x00007fff88421323 _dispatch_call_block_and_release + 12
    6   libdispatch.dylib             	0x00007fff8841cc13 _dispatch_client_callout + 8
    7   libdispatch.dylib             	0x00007fff8841f88f _dispatch_root_queue_drain + 935
    8   libdispatch.dylib             	0x00007fff8842dfe4 _dispatch_worker_thread3 + 91
    9   libsystem_pthread.dylib       	0x00007fff8b8a6a9d _pthread_wqthread + 729
    10  libsystem_pthread.dylib       	0x00007fff8b8a43dd start_wqthread + 13

There aren’t any meaningful changes to `QSLib` here, but I ran without these changes for a few days and never saw the crash.

I’ve been running off this branch with the Debug configuration all day hoping to catch it and get more detail, but no luck yet.

skurfer:
I was trying to test a bunch of stuff at once, and I noticed this conflicts with your changes in https://github.com/quicksilver/Quicksilver/tree/t/notification-rewrite

Do you want to just include those changes here so we don’t run into the conflict later?

tiennou:
Just rebased this one and the notification-rewrite thing on top so there should be no conflicts.

skurfer:
> Just rebased this one and the notification-rewrite thing on top so there should be no conflicts.

Thanks! I think I’ve gone over this a couple of times already, so I’ll give it another quick look and probably merge tomorrow. Then it’s on to the aliases.

tiennou:
That would be good. I'm basing most of my other work on this one, so if it can go in quickly without too much regressions, I'd be happy ;-).

skurfer:
One down!

PR #2078, opened 2015-07-24
Fix typing / quickly. Fixes #2077

### LINKED ISSUE #2077: Slash character doesn't navigate into folder if I type too quickly
I have a folder called `src` in my index that contains the folder `net`. If I quickly type `src/net` into Quicksilver, it simply drops the slash:

screen shot 2015-07-23 at 13 32 58

If I wait for the `src` folder result to pop up before typing `/`, the correct sequence takes place:
screen shot 2015-07-23 at 13 33 53
screen shot 2015-07-23 at 13 34 00

This is a bit annoying and surprising, because it doesn't happen if I use the right arrow key instead of `/`. I like `/` because it's nicer to type, but the fact that it fails if I don't pause makes Quicksilver feel unreliable.

pjrobertson:
Basically if anything is typed immediately before the slash (not just a slash), then it's OK to drill down (I think).

I also got rid of one `if()` for the case where `/` is long pressed. I didn't like the fact that it would immediately drill down into 'Macintosh HD' if you held `/` for long enough. I'm happy for this to be changed back though if there's a case for it


skurfer:
> I didn't like the fact that it would immediately drill down into 'Macintosh HD' if you held `/` for long enough.

Same here. I’ll give this a look.

PR #2080, opened 2015-08-03
provide a nicer fallback behavior for stringValue

skurfer:
No more `NSAssert()`.


tiennou:
Also, some explanation of why you're removing it would be nice ;-). I'm not fond of seeing some catalog tree search involved...

pjrobertson:
@tiennou - get on to IRC. Both Rob and I are sitting next to each other - quite a novelty! :D


I prefer to use encrypted email. You can find my public key here <http://patjack.uk/pk.asc>.
Learn how to encrypt your email with the Email Self Defense Guide <https://emailselfdefense.fsf.org/en/>

skurfer:
> Also, some explanation of why you're removing it would be nice

It hangs the interface. Good reason? :smiley:

PR #1611, opened 2013-09-18
Rework QSTask.

tiennou:
This supersedes #1603 and #1265.

Major surgery on QSTask, this makes QSTask a *real* model-only object, kills the deadlock issues between `QSTaskController` and `QSTaskViewer` and fixes the auto-show/hide behavior once and for all (at least it seems to behave right, now ;-)) :

- It will show if there's a new task and the prefs says it (it was).
- It won't show if during the time the task was created and the time the viewer reacts the task is done. This should fix #1040, and stop me wanting to bash it each time I open a new tab.
- It won't try to auto-close the window if you open it manually and a task finishes (this fixes a never reported issue ;-)).

Known issues/discussion:

- The *first* time it opens (set it to automatically open, quit and restart so you get the catalog startup scan), the view is messed up because it extends all the way to the right, which looks weird.
- I've setup minimal code to correctly to handle subtasks, but there's no viewer suppport, and I'm not sure it's worth it. It was *mostly* present, but the only thing I can think that would use that is the catalog scanning, and maybe the update system.
- I'm tempted to push up the `HIDE_DELAY` by a few millisec, so you can see the task completed. Right now it just closes before you can even see the task has ended (which looks the same to the user, but could be nicer).
- I've killed the `target`/`action` mechanism it had for task cancellation, instead relying on a `cancelBlock`.
- Documentation. I really want to do this one, but I need some more time.
- This *could* break plugins that used tasks (there was a `-(QSObject *)result` method I've killed, plus the effects of the major surgery). But I don't think that many plugins did that, and I think the API as it is is sufficient for "general user information". But if you find I broke some plugin, I'll happily revise and provide at least a *private* runtime equivalent.
- Maybe the preferences settings should be also specified in the Extras pref pane ? I think it's pretty hidden atm, IIRC by default it won't autoshow, and you can't change those settings unless you open it first and fiddle with the gear menu. #119 talks about that a little.

All in all, eyes are welcome so I'm breaking anything, and if you stop a thing I didn't point out in the list above, the

Issue #2016, opened 2015-01-16
Check index validity on the respective scan queue


skurfer:
Got a hang the other day (which is rare for me). It looks like the main thread got stuck on `indexIsValid` for one of the entries.

The title of this issue might be too specific. Generally, we need to move more of the catalog related stuff out of the main thread. (There are many cases where checking for validity requires hitting the disk or network.)

skurfer:
For the curious:

```
Call graph:
    2618 Thread_6830359   DispatchQueue_1: com.apple.main-thread  (serial)
    + 2618 start  (in libdyld.dylib) + 1  [0x7fff898e75c9]
    +   2618 NSApplicationMain  (in AppKit) + 1832  [0x7fff865b52d4]
    +     2618 -[NSApplication run]  (in AppKit) + 594  [0x7fff865c9e23]
    +       2618 -[NSApplication nextEventMatchingMask:untilDate:inMode:dequeue:]  (in AppKit) + 194  [0x7fff865d5e80]
    +         2618 _DPSNextEvent  (in AppKit) + 964  [0x7fff865d66d1]
    +           2618 _BlockUntilNextEventMatchingListInModeWithFilter  (in HIToolbox) + 71  [0x7fff8b96cffb]
    +             2618 ReceiveNextEventCommon  (in HIToolbox) + 179  [0x7fff8b96d0be]
    +               2618 RunCurrentEventLoopInMode  (in HIToolbox) + 235  [0x7fff8b96d43f]
    +                 2618 CFRunLoopRunSpecific  (in CoreFoundation) + 296  [0x7fff8c6b9838]
    +                   2618 __CFRunLoopRun  (in CoreFoundation) + 910  [0x7fff8c6b9e0e]
    +                     2618 __CFRunLoopDoBlocks  (in CoreFoundation) + 341  [0x7fff8c6ba655]
    +                       2618 __CFRUNLOOP_IS_CALLING_OUT_TO_A_BLOCK__  (in CoreFoundation) + 12  [0x7fff8c6c854c]
    +                         2618 __44-[NSApplication _reactToScreenInvalidation:]_block_invoke  (in AppKit) + 75  [0x7fff86914c2d]
    +                           2618 -[NSApplication _reactToScreenInvalidationImmediately:]  (in AppKit) + 379  [0x7fff86914dee]
    +                             2618 -[NSNotificationCenter postNotificationName:object:userInfo:]  (in Foundation) + 66  [0x7fff93e95ea1]
[...]

PR #2091, opened 2015-08-31
don't lock up the main thread to set properties

skurfer:
Setting these things doesn’t count as “updating the UI”, so no need to do it on the main thread.

It looks like #1611 does something similar, but there seems to be a lot to work out before we can merge that.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSTask.m
+++ b/Quicksilver/Code-QuickStepCore/QSTask.m
@@ -182,11 +182,9 @@ static NSMutableDictionary *tasksDictionary = nil;
 }
 
 - (void)setName:(NSString *)value {
-    QSGCDMainSync(^{
-        if (name != value) {
-            name = [value copy];
-        }
-    });
+    if (name != value) {
+        name = [value copy];
+    }
 }
 
 - (NSString *)status {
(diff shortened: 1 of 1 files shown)


pjrobertson:
Hmm... looking at git blame, it seems like I did that as KVO changes should
all be done on the main thread. See my original pull for the explanations:

https://github.com/quicksilver/Quicksilver/pull/1319

As for what our alternatives are, I'm not really sure. Find out where the
blocking is happening and fix that? It's inherently a problem with our
multi-threading complexities as opposed to that property setting I guess,
but tracking it down is the difficult part.

skurfer:
> it seems like I did that as KVO changes should all be done on the main thread

Ah, glad you could find the original reference. I want to find something more concrete than an e-mail thread, though. The guys here where I work that do Cocoa all day every day told me these changes should be fine.

I find the idea that we have to change these properties on the main thread hard to swallow. It seems reasonable in this case, because we control the class that manipulates the properties, but KVO wouldn’t be practical if you could only ever use it with classes you wrote yourself. I’ll look for an official answer.

> Find out where the blocking is happening and fix that?

I did that, too. I just haven’t created the pull request yet. :smiley:

tiennou:
Are you talking about the scan-update-related locking we're seeing ? (I got one of these by "accidentally" selecting the Volumes preset). I didn't even see `QSTask` involved here, my gut feeling was more along the lines of "why the hell are we still scanning/indexing/whatever on the main thread ?"...

I would actually prefer #1611 to be merged, because it's supposed to be cleaner ;-). I may try to devote some time to get it rebased one of these evenings.

skurfer:
I’ve never encountered this lock myself, but you can see it in #2090. It looks typing in the UI updates the object history, which triggers a rescan, which creates a task, and that task gets updated on the main thread where the user is attempting to type.

I think catalog scanning is *mostly* kept off the main thread, except for this task stuff, and #2016.

> I would actually prefer #1611 to be merged, because it's supposed to be cleaner ;-)

Me too.

tiennou:
Yeah, so we're talking about the same deadlock. My personal preference would be to make either `reloadEntry:` or `scanForced:` go do its thing in the background instead of on the main thread.

PR #2093, opened 2015-09-01
Indentation

tiennou:
This changes Xcode indent setting to always use tabs, as well as some TextMate-specific setup.

When this gets merges, I *may* do "Code Cleanup 2" ;-)


pjrobertson:
I actually can't remember what we said we preferred. Tabs or spaces. I don't really mind though, maybe spaces is nicer.

* What happens when a new file is created? Will it be set to some 'default' or will it still go on the user's preferences
* You say you've made changes to "always use tabs" but a diff of the code shows each file ref has `usesTabs=0` appended to it in the Xcodeproj. I'd assume that means "don't use tabs"?

tiennou:
IIRC I based the choice by running this [gist](https://gist.github.com/tiennou/1d2eab37b64a44d33947). You need to pass it a glob pattern, like "*.[hm]". It's quick and dirty but does the thing. Aaand it seems tabs are a majority, so obviously this is reversed... I'll fix that then ;-).

tiennou:
And yes, I think that's what `usesTabs` means. I actually didn't expect Xcode to have a per-file setting, and recently worked on another project where whatever I did to the indent setting wouldn't register in the xcodeproj, so there goes consistency...

skurfer:
I used tabs in Xcode for a long time because most existing QS code was done that way. Then we decided on spaces, so I said “Oh, thank God” and changed my preferences. (To answer the question of what we decided.)

It’s all a mess now, so I don’t know if there’s any clean way out other than to decide (again) and enforce it. Then the code will eventually all get to one or the other. Since we’re closer to “all tabs”, we might get there quicker, so I can see an argument for that.

I’m of the opinion that it will *always* be a mess and we shouldn’t worry about it that much. :smiley:

Note that you also told TextMate to use spaces.

And what about all the plug-ins?

tiennou:
Yeah, I don't know what I was up to when I changed those settings. I'll fix it, once the plan gets more thoroughly fleshed out.

Thinking about it, I think I'll enhance the script to report on mismatching indent settings per file — here it is :
```
562 files, 29251 tab lines, 17426 spaces line
225 files have mismatching indent :
[snip list of files]
169 mismatching files have more tabs, 48 have more spaces
```

Since those mismatching files are the ones I'm actually bothered with (because I know of no editor that handles mismatching indent settings the correct way, and even TM per-project settings are useless), the plan can be "Define tab as the normal indentation, switch the funky ones to use tabs, keep the rest how they are". Or maybe just switch everything to tabs and be done with it...

pjrobertson:
@tiennou your commit changes are still saying we're using spaces... but I thought we'd agreed on tabs?! :s

FYI if we *do* want to change everything to using tabs, here's a simple command to run:
```
find . -type f -name "*.[mh]" -exec sh -c "unexpand -t 4 \"{}\" > \"{}.tmp\" && mv \"{}.tmp\" \"{}\"" \;
```

pjrobertson:
@skurfer said:
> And what about all the plug-ins?

Should we add the .travis file to all the plugins and do the same `check_indent.rb` check? Only thing is we'd either need to include `check_indent.rb` in all the plugin project or do something like this in travis:

```sh
wget https://raw.githubusercontent.com/quicksilver/Quicksilver/ci/Quicksilver/Tools/travis/check_indent.rb -O check_indent.rb && ruby check_indent.rb
```

pjrobertson:
Can we merge this already and finally end the tab/spaces battle? ;-)

I think the only thing that needs changing is that @tiennou needs to update his commit message to say "tabs" instead of "spaces"

skurfer:
> Can we merge this already and finally end the tab/spaces battle? ;-)

Besides the commit message, TextMate is still set up to use spaces instead of tabs. And there’s no new-line at the end of that file. :smiley: Maybe we should just redo this in another branch.

tiennou:
Rebase (at last !). Please Merge Quickly, it's "painful" to be editing like crazy and having to fix Xcode incorrect settings after the fact :wink:.

pjrobertson:
From Rob:
> And there’s no new-line at the end of that file.

:D
Haha

tiennou:
You're making me suspicious now, isn't `softTabs = !hardTabs = false` ?

pjrobertson:
:confused: what's your question?! Everything looks fine to me. I was just joking about Rob's requirement for a new line at the end of the file.

tiennou:
My bad, the joke flew right above my head :laughing:. I was misled by the starting "And".

skurfer:
> Please Merge Quickly

Won’t this conflict with #2177 now?

tiennou:
I'll rebase if it does. I'm actually more concerned about the others PRs that have changes to the project file :cry:.

tiennou:
check_indent now reports the file and hunk headers, so you can more quickly get to where the problematic lines are (instead of having to go through the diff for context).

Also, I don't think it will actually be problematic when merging since #2177 uses those commits directly.

PR #2094, opened 2015-09-02
make Quicksilver available on launch

skurfer:
Unless you’re the sort of person that likes answering this question once per week…


pjrobertson:
aha! you figured out how to fix it?! :D

PR #2096, opened 2015-09-08
Warn users when trying to run the app on 10.8

pjrobertson:
Turns out we made 1.3.0 10.9+ only, but forgot to change the warning in QSController

PR #2104, opened 2015-09-18
Notification Center

### LINKED ISSUE #1669: Too many dialogs when an update is available.
I counted 4 separate dialogs today. One asks if the user would like to install an update. A second one installs the update. A third one asks if I want to relaunch. And the fourth and completely unnecessary one asks if I want to view release notes.

Please fix this to be like Chrome, where updates are automatically downloaded and installed. At best, there should be one final dialog that asks about relaunching the app. Given that Quicksilver is a background task that doesn't hurt if killed and restarted (i.e. there is no user-created state that might be lost), I'm happy even if it was killed and relaunched automatically with no prompt. And if somebody really wants to read release notes, they'll find a way to get to them. No point asking the other 99.99% about it for every single update.

Thanks for listening!

### LINKED ISSUE #1859: Re-implement confirmation dialogs using Notification Center
Assuming that we are going to drop support for 10.7, we should switch dialogs for updates, release notes, etc. to use Notification Center so the user can have more control over them and so they won’t steal focus.

skurfer:
As described in #1859, this replaces pop-up alerts with Notification Center notifications where appropriate.

I did not convert:

* older messages that are unlikely to be shown
* messages that really *should* interrupt the user or result from explicit user actions
* messages that are too long to fit in a notification (like those that display `localizedDescription`)

All subjective, of course, so let me know if I didn’t go far enough (or went too far).

As part of this, I switched the default notification style for the entire application to “alert”. This is necessary to get the action buttons, but it’s probably not appropriate for the places we were previously using Notification Center (where you’d use Growl or the built-in notifier) because now they *all* have action buttons and require manual dismissal. I don’t see a way around it, so I’m inclined to change the default for those back to the built-in notifier.

Also, all the strings are localized now (as in available to translators but not actually translated), which is an improvement.

This also does the same thing as #2096. You wanted to know if my “interesting” project was small, Paddy? Let’s ask instead if it’s worth it. :grimacing:

fixes #1859
fixes #1669


tiennou:
Here's my take on that : https://github.com/quicksilver/Quicksilver/tree/t/notification-rewrite

It planned to 1) kill deprecated stuff 2) handle OS X versions that don't have Notification Center (but we might be way after that actually) 3) hide the modal/notification choice 4) ~~profit~~. I think the actual Notification Center thing is actually in a stash somewhere, but if that inspires you...

The only thing I don't like (but that's an "Apple API problem") is how you can't use blocks to handle notification replies and have to go through delegation. So you litter code everywhere.

Also, indentation ;-).

skurfer:
> Here's my take on that : https://github.com/quicksilver/Quicksilver/tree/t/notification-rewrite

Cool. Want me to merge that into this branch?

> The only thing I don't like (but that's an "Apple API problem") is how you can't use blocks to handle notification replies and have to go through delegation. So you litter code everywhere.

Agreed. I tried to have just one class be the delegate to keep it contained, but that’s messy, too.

> Also, indentation ;-).

Yeah, OK. Should I do anything about that for this pull? FWIW, I changed my Xcode settings (to tabs) a few days ago.

Quicksilver/Code-App/QSController.m
```diff
@@ -645,16 +648,24 @@ - (void)checkForFirstRun {
...
 			if (lastLocation && ![bundlePath isEqualToString:[lastLocation stringByStandardizingPath]]) {
 				//New version in new location.
-				[NSApp activateIgnoringOtherApps:YES];
-				NSInteger selection = NSRunAlertPanel(NSLocalizedString(@"Running from a new location",nil), NSLocalizedString(@"The previous version of Quicksilver was located in \"%@\". Would you like to move this new version to that location?",nil), NSLocalizedString(@"Move and Relaunch",nil), NSLocalizedString(@"Don't Move",nil), nil, [[lastLocation stringByDeletingLastPathComponent] lastPathComponent]);
-				if (selection)
-					[NSApp relaunchAtPath:lastLocation movedFromPath:bundlePath];
+                NSUserNotification *locationAlert = [[NSUserNotification alloc] init];
+                [locationAlert setIdentifier:QSLocationChangedUserNotification];
+                [locationAlert setTitle:NSLocalizedString(@"Running from a new location", nil)];
+                [locationAlert setInformativeText:NSLocalizedString(@"The previous version of Quicksilver was located in America. Would you like to move this new version to that location?", nil)];
```
tiennou: America ? Really ?? :laughing:

tiennou:
Note, I'm currently working on that one. I'm not fond of the hardcoded delegate stuff, so I'm moving the notification-managing to `QSAlertManager`, so we can use blocks elsewhere and get a nice "modal" alert if we're frontmost.

Quicksilver/Code-App/QSController.m
pjrobertson: Haha! It's just Rob testing our code-review quality I guess :P

skurfer: Ha! Yeah, uh… I was testing. You passed.

Seriously, no idea what that’s about. Find/Replace gone wrong maybe?

PR #2105, opened 2015-09-23
Add a new CI test to make sure we use tabs and not spaces. Fixes #2050

### LINKED ISSUE #2050: Code style
I'm filing a new issue so we can discuss that ;-). I'm investigating some style guide I'd like us to all use when coding. @skurfer told me we preferred 4 spaces, so I went along and built a quick Ruby script so I could be sure we were going the right way.

Here's the script : https://gist.github.com/tiennou/4a4561e55050234b0031
The results of `./indent-stat.rb "*.[hm]"` are `{:files=>562, :tabs=>29451, :spaces=>16996}`.

That's 63% tabs. This means I'd like to switch to tabs because there's almost 2 times more lines that begin with those.

And then I'd propose going setting that [style](https://github.com/raywenderlich/objective-c-style-guide) as "canonical", with an exemption w.r.t to us using tabs for indent.

Some insight on the tab/space wars : [1](http://boredzo.org/blog/archives/2007-05-07/tabs-vs-spaces) + [followup](http://boredzo.org/blog/archives/2008-11-05/tabs-vs-spaces-redux).

There's also a bunch of links at the end of the style guide above which points to more style guides. Feel free to take a look at them, but I just took the first one I found that looked like something I'd want to enforce.

Fun fact: our "Initial commit to GC" is `{:files=>713, :tabs=>35909, :spaces=>19675}`. That's 64% tabs. I'll let you guess which commit caused a drop of almost 200 files ;-).

pjrobertson:
So, based on Etienne's recent PR on tabs/spaces, and after watching [this](http://www.bbc.co.uk/academy/technology/article/art20150206154333467) video where a GitHub guy was saying they're now using CI for much more than just tests (CSS validation etc. - watch the video!) I thought it would be interesting if we did the same.

I've butchered Etienne's gist ruby script (and probably written very bad ruby code) and have attempted to write a script that will check all newly added lines in .m/.h files and check to make sure they start with tabs and not spaces


pjrobertson:
P.S. I know @skurfer was [against](https://github.com/quicksilver/Quicksilver/pull/2021#issuecomment-72402365) using 'custom' scripts originally, but maybe I can persuade him here... ;-)

There's definitely scope for us to expand this out in the future - adding in more scripts and stuff, and it's not exactly 'difficult' to remember (although maybe we should have a directory for travis stuff kept separate).
Perhaps some day we could even have some kind of 'real' CI that meant when plugin PRs are merged they are automatically built and pushed to the plugin repo :-)

tiennou:
God, I feel like the pesky guy ;-). I'm not too fond of dot-hiding the script, could it be moved under `/Scripts` (or wherever we have those, I don't remember the name offhand) ?

pjrobertson:
...probably a good idea :-)
I've given it its own folder now, and have removed my test commits. It works as expected (fails when there are lines with spaces) so that's all good :-)

skurfer:
Man, you guys are really going ahead with this tab stuff, eh?

* Why is the Map stuff in here? It should already be on `master`.
* I was against using a custom script to do something Travis should handle natively. This doesn’t bother me too much.
* My first thought was “What a bullshit reason to fail tests!” but if you were forcing spaces instead of tabs, I’d think “Oh, yeah. We need that!” so… my argument is probably not objective. 😀
* I see the new Ruby script is indented using spaces, which is hilarious. At least it shows we’re not *complete* animals. 🐗
* Do you want to say that this “fixes” #2050?

pjrobertson:
> Why is the Map stuff in here? It should already be on master.
No reason except for sloppy git skills. Fixed

> I was against using a custom script to do something Travis should handle natively. This doesn’t bother me too much.

Cool

> My first thought was “What a bullshit reason to fail tests!” but if you were forcing spaces instead of tabs, I’d think “Oh, yeah. We need that!” so… my argument is probably not objective. 😀

Personally I'm still pro spaces, but since the QS codebase is primarily tabs as Etienne's discovered, I thought it was best we go down that route

> I see the new Ruby script is indented using spaces, which is hilarious. At least it shows we’re not complete animals. 🐗

Hahaha, well that's just hilarious! Maybe we need to decide what 'coding style' we have for ruby though... what if we want to use spaces? ;-)

skurfer:
Better. Were the tests not getting called via `xctool` before? I thought they were running somehow.

pjrobertson:
> Better. Were the tests not getting called via xctool before? I thought they were running somehow.

I basically copied the exact same command that [Travis was calling](https://travis-ci.org/quicksilver/Quicksilver/builds/78277014#L122) before I made these changes.
If you look at the logs for [tests on this branch](https://travis-ci.org/quicksilver/Quicksilver/builds/81773128), you'll see exactly the same thing is run/log output is generated.

pjrobertson:
...I've also switched check_indent to use tabs...!

pjrobertson:
Can this be merged? Every day that passes, we run the risk of ruining our code with spaces.. ;-)

PR #2107, opened 2015-09-27
nl.lproj string-files translated into Dutch

gvmelle:
I saw that the files where not translated, so I did that. Hiope it helps


pjrobertson:
Thanks for doing this! We manage our translations through [Transifex](https://www.transifex.com/quicksilver/quicksilver/) - so would it be possible for you to add your translations there? (to make sure we don't lose any translations that somebody has already done on Transifex)

Thanks very much for your effort!

gvmelle:
OK, even better!

PR #2141, opened 2015-10-27
Handle the $ variables in Spotlight queries

### LINKED ISSUE #2140: Spotlight plugin > date queries
Time variables don't seem to work in the Spotlight plugin query.

This doesn't work: `kMDItemLastUsedDate >= $time.this_week`
This does: `kMDItemKind == 'Folder'`

tiennou:
Fixes #2140

### DIFF
--- a/Quicksilver/Code-QuickStepFoundation/NSMetadataQuery+Synchronous.m
+++ b/Quicksilver/Code-QuickStepFoundation/NSMetadataQuery+Synchronous.m
@@ -17,7 +17,10 @@
 - (NSArray *)resultsForSearchString:(NSString *)searchString inFolders:(NSSet *)paths
 {
 	[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(doneSearching:) name:NSMetadataQueryDidFinishGatheringNotification object:nil];
-	NSPredicate *search = [NSPredicate predicateFromMetadataQueryString:searchString];
+    NSPredicate *search = [NSPredicate predicateWithFormat:searchString];
+    if (!search) {
+        search = [NSPredicate predicateFromMetadataQueryString:searchString];
+    }
 	[self setPredicate:search];
 	if (![paths isEqual:nil]) {
 		NSMutableArray *pathURLs = [NSMutableArray array];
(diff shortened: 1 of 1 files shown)


pjrobertson:
Latest QS is 10.9+ right?
In that case we need to change [this line](https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/Code-App/QSController.m#L53)

Don't see anything wrong with this though, as long as it doesn't break existing Spotlight Query stuff

pjrobertson:
...is it worth adding unit tests for doing a query with the two examples given in #2140 ? :-)

skurfer:
> Latest QS is 10.9+ right?
> In that case we need to change this line

That’s already changed in #2096 and #2104. No need to do it a third time. :smiley:

skurfer:
I wonder why this doesn’t like `*` now…

```
2015/10/29 12:11:33.956 AM Quicksilver[605]: An error ocurred while scanning "Comment Tags": Syntax error in Metadata query string: kMDItemFinderComment LIKE[cd] '*#*'
2015/10/29 12:11:33.959 AM Quicksilver[605]: An error ocurred while scanning "Yojimbo Items and Tags": Syntax error in Metadata query string: kMDItemKind LIKE 'Yojimbo*'
```

pjrobertson:
Looks like NSPredicate and Spotlight use different syntax :(

https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Predicates/Articles/pSpotlightComparison.html

tiennou:
@pjrobertson Thanks, I *knew* there was some differences between the 2 syntaxes but wasn't able to find the actual documentation.

@skurfer Was this running the fix or the `master` version ?

skurfer:
> that doesn't explain why it did work before, and not now.

I don’t see them using single quotes in any of the examples. Maybe that’s the problem, and not the asterisk.

> Was this running the fix or the `master` version ?

The queries work with `master`, but fail with this change.

tiennou:
Differences are that NSPredicate mandates quoting constants and its LIKE clauses are not the same. We can fix that by making it try both syntaxes, but meh.

skurfer:
> is it worth adding unit tests for doing a query with the two examples given

Along with a few others? Yes. I put it on my Remember the Milk Quicksilver list.

Issue #2064, opened 2015-06-04
Recent QS 1.2.2 (4011) will hang on OSX 10.9.4 often


JurekRaben:
Hi,

the recent version of QS quite often starts to hang as an unresponsive process and doesn't react on user actions anymore. Have to to force quit it using activity monitor. After reloading it works again... 

OSX 10.9.4

skurfer:
This sounds similar to #1977.

Do you have any aliases or symbolic links in your home directory to files on a network drive?

JurekRaben:
Yes, I have links to network drives on my desktop. These links are not accessible sometimes, if the network drive is offline.

skurfer:
OK, thanks. I’ll close this as a duplicate. All these symbolic link issues are high on the list.

Issue #2036, opened 2015-03-10
Relaunch action missing for applications installed by `brew cask`


timvisher:
`brew cask` installs applications by symlinking them to `~/Applications`. This is fine and works to open them or to activate them, but it doesn't work to relaunch them.

I've discovered that 𝑄uıcĸsılⅴεʀ will Relaunch the full path to the application, which will often be the next entry in the results panel when searching for the app.

I'm guessing this is because of how qs is handling the symlink. It would be awesome if it could detect that the symlink is an application in some way and allow me to relaunch it via the symlink.

BartlomiejSkwira:
+1, same situation here. Launching the manually helped but would be cool if Quicksilver would do that automagically

izaakm:
+1

Issue #2030, opened 2015-02-20
Symlinked applications catalog error (1.2.2)


leoj3n:
Encountering the issue #1882 was supposed to fix.

OS X 10.8.5
Quicksilver 1.2.2

I have tried rebuilding the catalog, tried creating a copy of the "Applications (User)" preset with depth of 1 and no type, and tried disabling Finder/Path Finder plugins to no avail.

Regression?

skurfer:
I just tried `brew linkups python3` to create some symlinked applications in `/Applications` and they get indexed correctly. Are your symlinks in `~/Applications`? (Not that is should matter.)

leoj3n:
Do you mean `brew linkapps`? Yes my symlinks are in `~/Applications`.

skurfer:
> Do you mean `brew linkapps`?

Yeah. Autocorrected.

> Yes my symlinks are in `~/Applications`.

I tried moving one there and it worked just as well. All the code involved in that should be OK on 10.7+.

Are you getting useless objects in the catalog, or are the linked apps not appearing at all? Do you see them in the Catalog prefs if you select “Applications (User)” and open the Info panel, then go to Contents?

leoj3n:
No useless objects. The symlinked application are appearing. Same problem regarding "Applications" as "Applications (User)".

![catalog](https://cloud.githubusercontent.com/assets/990216/6281038/a1ca6e5e-b87d-11e4-84fc-d81132681789.png)

Note: Same result if selecting "Applications" or "Applications (Catalog)".
![applications](https://cloud.githubusercontent.com/assets/990216/6281048/c592eac8-b87d-11e4-8d16-ef7a8d162a2e.png)

![macvim](https://cloud.githubusercontent.com/assets/990216/6281049/c7e53b14-b87d-11e4-98d6-0e0451004137.png)

I should also mention Quicksilver has been crashing every now and then. Maybe I need to obliterate the catalog by hand and let it rebuild it? Maybe a plist? Dunno. I'll try reinstalling when you give me the go-ahead unless you want me to troubleshoot further.

skurfer:
> Maybe I need to obliterate the catalog by hand and let it rebuild it? Maybe a plist? Dunno. I'll try reinstalling when you give me the go-ahead unless you want me to troubleshoot further.

About the only thing useful is clearing the caches. I usually do it Terminal. Quit Quicksilver, then run

    rm -r ~/Library/Caches/*Quicksilver

and start it again.

[...]

Issue #1977, opened 2014-11-03
occasional hangs 


heikohaller:
since upgrading to 1.2.0 (400F), QS becomes unresponsive every couple of days.
Task Manager does not show it a unresponsive though – It simply does not activate on shourtcut, also application triggers do not work. 
Workaround: Quitting from Activity Monitor and restarting QS helps.

OS: Mavericks (10.9.5)

Installed Plugins:
- contacts
- evernote
- cyberduck
- Keychain
- Safari
- Shelf
- Terminal

(Disabled all plugins now, will report back)

1-61803:
See also this thread https://groups.google.com/forum/#!topic/blacktree-quicksilver/Xo2YXiAFiB4

skurfer:
So I don’t forget… Look into `NSURLBookmarkResolutionWithoutMounting`

tiennou:
(Note there's a [commit](663ffb4c9a758427424cdc5382ed699c0421b788) in #2075 that tries to make the NSURL stuff a little easier to use).

ethanbeyer:
Hey there, I have been having a very similar issue to what @heikohaller was describing, and my ticket, #2132, was closed. No biggy, but @skurfer said to me in that thread:

"Until then, check your catalog for links or aliases to files on network drives that aren’t available."

I have none in my catalog. I've even reset the preferences and such. Seemed to help for about a day, but the problem has sporadically persisted.
I'll take a sample from the Activity Monitor next time it happens and will report back.

Thanks!

ethanbeyer:
Follow-up: I was wrong.

![screen shot 2015-10-29 at 09 35 41](https://cloud.githubusercontent.com/assets/6864684/10819797/7162c342-7e20-11e5-9a58-f2a11d483251.png)

I didn't realize that "Recent Items" was caching the `smb://` server I connect to for work.

But *now* on to what I find really odd: some of the most recent occurrences of this problem have been while I am connected to this server.

Scratching my head, yo.

skurfer:
Fixed by #2142

PR #2142, opened 2015-10-28
Modern aliases

tiennou:
Sorry for the name, *you* know that's a hard problem ;-).

This resurrects the switch to NSURL bookmarks I tried to do (#2075), since I wanted to fix the class of network-alias-stalls bugs we have and the main QS method we had actually doesn't handle all the bells and whistles of bookmark resolution. Maybe I should have added a comment on `-[NSFileManager resolveAliasAtPath:]` and friends but, well, I didn't. Consider yourselves warned :smirk:.

This one fixes the issue Rob had with the hundreds of warning messages had because *symlinks*. Since I wasn't keen on basing the new resolver because of [shaky grounds](http://digitalleaves.com/blog/2013/09/nsurl-files-equality-comparison-and-equivalency/), I preferred to keep the symlink-checking separate.

There's also a fix for `-infoRec` returning the symlink target info instead of the symlink itself, which makes symlink actually "invisible" (and non-working, it would seem — at least here, current `master` doesn't allow me to right-arrow into a symlink because the code passes a symlink path to some directory-parsing code). I wonder how you guys feel about that, especially in light of #1897, #1485, and the broken test ;-).

The icing on the cake is that "resolvables" show their target path instead of theirs.

Based on the deprecation cleanup branch ~~for motivation~~ so the diff is cleaner.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
@@ -682,15 +682,15 @@ NSArray *recentDocumentsForBundle(NSString *bundleIdentifier) {
     if (!path) {
         return nil;
     }
-    
-    NSURL *fileURL = [NSURL fileURLWithPath:[path stringByResolvingSymlinksInPath]];
+
+    NSURL *fileURL = [NSURL fileURLWithPath:path];
     NSError *err = nil;
     
     // Note: NSURLLocalizedLabelKey gives a localized string of the Finder label (tag in 10.9+). E.g. Red / Rouge
     dict = [fileURL resourceValuesForKeys:@[NSURLTypeIdentifierKey, NSURLIsDirectoryKey, NSURLIsAliasFileKey, NSURLIsPackageKey, NSURLLocalizedLabelKey, NSURLLocalizedNameKey, NSURLVolumeURLKey, NSURLIsExecutableKey, NSURLIsWritableKey, NSURLLabelColorKey] error:&err];
     NSMutableDictionary *mutableDict = [dict mutableCopy];
 
-    if (err) {
+    if (!dict) {
         // The file is either a socket or fifo. Not much information is available on these files
         LSItemInfoRecord record;
         OSStatus status = LSCopyItemInfoForURL((__bridge CFURLRef)[NSURL fileURLWithPath:path], kLSRequestAllInfo, &record);
--- a/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
@@ -161,9 +161,14 @@ NSArray *recentDocumentsForBundle(NSString *bundleIdentifier) {
                 return @"iCloud";
             } else {
                 NSString *pathDetails = [path stringByAbbreviatingWithTildeInPath];
-                if (UTTypeConformsTo((__bridge CFStringRef)[object fileUTI], kUTTypeResolvable)) {
-                    NSString *extraInfo = [[object fileUTI] isEqualToString:(__bridge NSString *)kUTTypeSymLink] ? NSLocalizedString(@"symlink", @"Extra details shown for symlink files, in the UI") : NSLocalizedString(@"alias", @"Extra details shown for alias files, in the UI");
-                    pathDetails = [pathDetails stringByAppendingFormat:@" (%@)", extraInfo];
+                if (QSTypeConformsTo([object fileUTI], (NSString *)kUTTypeSymLink)) {
+                    pathDetails = [[path stringByResolvingSymlinksInPath] stringByAbbreviatingWithTildeInPath];
+                } else if ([object isAlias]) {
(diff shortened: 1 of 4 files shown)


skurfer:
Nice to see you, as always. So, let’s see. Some quick testing I did shows that this fixes

  * #2063
  * #2030
  * #1977 (I could never reproduce it, but I'm taking your word)
  * #1952
  * #1884
  * #1498 (if it was ever really broken)

In other words, every known issue with symbolic links. I had them all assigned to me, but hadn’t gotten to them yet. Nice work!

Regarding my comment above about the UTI comparison, I found an [old comment][1] that suggests this test might not even work, but I’m seeing the target path in the details, so maybe it was fixed at the OS level?

Personally, I’d like the target path in the details string prefixed with things like “Link to” and “Alias of”. Not saying you need to bother with that right now, though. I can add it later.

[1]: https://github.com/quicksilver/Quicksilver/issues/1884#issue-36291366

tiennou:
I wasn't too keen on adding the `-isSymlink` thing, but maybe for consistency that's better. I'll try to review how much manual symlink testing there is, and if it can consolidate some tests, I'll add it here.

#2063 ? I think you meant #2064. #2030 I'm not sure because of Path Finder being involved, but I think that's a red-herring. The others should be fixed, given the few tests I ran.

Speaking of tests, I'll write some. I don't think the ones we currently have cover all the cases so they behave themselves (aliased symlinks anyone ? :laughing:).

The comment you point out is the error I had, because of the "code passes a symlink path to some directory-parsing code". `-fileUTI` failing was the same reason, it returned the target's UTI.

I made my `details` changes while running Bezel, the goal was to make the "resolvable" nature apparent by displaying the target's path instead of the resolvable in that case, which IMHO looks nicer than (symlink) and less jarring that what you're proposing (still a path).

skurfer:
Actually, I meant #2036. I’ve updated my comment.

> the goal was to make the "resolvable" nature apparent by displaying the target's path instead of the resolvable in that case, which IMHO looks nicer than (symlink) and less jarring that what you're proposing (still a path).

It’s apparent to me, but I’m worried it will be confusing to some people. Not something we need to sort out immediately.

tiennou:
Oh, I'll check that one then. Haven't really looked at what happened to actions on resolvables with that change.

I was hoping that the "little alias arrow" that gets added was sufficient for conveying this was a symlink/alias. Granted, the GUI atm only says "this is a resolvable", not the actual type of it, but I don't think the difference is that meaningful. But yeah, this is something we can change.

tiennou:
Note that after looking at #1977, I will try to make a few network aliases (I had none of those when I wrote that), and then check what's their behavior in both `master` and this. Sad thing that you can't use your own MBP's file sharing to test...

tiennou:
#2036 is not fixed by that. It looks to me like a "feature request" to make process actions work on resolvables.

skurfer:
I looked through the list of actions for any that could be potentially problematic for symbolic links. All of these worked in a sane way:

* Reveal
* Always Open Type With…
* Tagging
* Move to Trash
* Delete
* Get Info
* Move To…
* Copy To…
* Rename
* Get File URL
* Get Path

These had minor problems.

* Get Location - show the HFS+ path to the target file
* Compress - doesn’t seem to do anything (when run on a link to a folder)
* Spotlight comment - puts comments on the original file
* Open in Terminal - doesn’t appear in the list of actions

Only one of those (Get Location) is provided in Core Support and needs to be dealt with here.

Things to note:

  * The ones that aren’t working probably never were, but we’re so close to fixing *everything*, we might as well talk about them now.
  * I only tested symbolic links. Not aliases. We should probably check those, too.

skurfer:
Tried the same actions with aliases. I wouldn’t say they all worked as expected, but they all worked consistent with Finder, which I’d call success.

I think there’s only one place (in the core application) that we’d use `-isSymlink`, so it’s not a priority. The only reason I’m not merging this now is I don’t have time to write the massive release notes. :smiley:

skurfer:
I was going to fix “Get Location” for symbolic links, then merge, but I’m not going to bother.

  1. It’s using `kCFURLHFSPathStyle` from `NSURL`, so unless we want to write our own code to make HFS paths, it’s going to return the path of the target.
  2. `kCFURLHFSPathStyle` is deprecated.
  3. Who cares?

On that last one, I would be OK with just removing the action completely. One less thing we depend on the ND classes for. :smiley:

tiennou:
I may or may not have based that PR against the deprecation-cleanup branch ;-). Sorry about that, I'll resubmit.

skurfer:
> I may or may not have based that PR against the deprecation-cleanup branch ;-).

You did, but that’s already been merged, so it doesn’t hurt anything. This has been merged too, so there’s no need to resubmit. :smiley:

tiennou:
Are you sure ? I just glanced at it after coming back from a lengthy weekend-of-music, and the graph made me thing those commits weren't in `master` at all (only your release notes did), so I opened #2157.

skurfer:
Oh, crap. You’re right. It went back to the branch it was based on, so it *did* hurt something. Now we have a mess. I’ll try to clean it up tonight and do another release.

PR #2144, opened 2015-10-28
Kill error handling

tiennou:
Because catching signals and exceptions only in debug mode felt weird. I had never before seen that "stability" alert before yesterday ;-).


skurfer:
I was really hoping this would get the Cube interface working again. Oh, well. :smile:

Issue #2111, opened 2015-10-02
Text-entry mode crash


eliasbe:
After restart I can do 1 web search then every time I tab to the 3rd pane QS crashes

Running El Capitan (started after I updated) and QS version 1.3.1

eliasbe:
Console says
2.10.2015 13:55:19,933 com.apple.xpc.launchd[1]: (com.blacktree.Quicksilver.59552[597]) Service exited due to signal: Segmentation fault: 11

molachai:
Getting the exact same error. I use a websearch for YubNub (setup detailed here: http://www.lifeclever.com/quicksilver-yubnub-rapid-web-searches-and-more/). Sometimes I don't even get the first search before QS crashes. Has been happening since I went to El Capitan betas. Now running full El Capitan 10.11. 

Error from console: 
10/3/15 12:43:04.852 com.apple.xpc.launchd[1]: (com.blacktree.Quicksilver.95392[3570]) Service exited due to signal: Segmentation fault: 11

skurfer:
I’m not seeing anything like this. Which interface are you using? (If you don’t know, go to Preferences → Appearance)

molachai:
I'm using Primer, but it happens on all of them.

eliasbe:
I'm using mini. This same error occurs when I try to rename files and do other opperations that make me tab to the 3rd pane. However some triggers that have the 3rd pane empty in text mode work.

molachai:
Woop. Spoke too soon. Did some testing. The YubNub search I have programmed to a trigger (Command-Shift-Y) crashes on the following interfaces:
Mini
Primer
White Bezel
Window

I've successfully completed multiple searches with no crashes on the following interfaces:

Bezel (built-in)
Bezel Classic
Flashlight
Menu
Mini-Bezel
Nostromo

Same error as the previous:
10/4/15 00:06:49.557 com.apple.xpc.launchd[1]: (com.blacktree.Quicksilver.95392[3866]) Service exited due to signal: Segmentation fault: 11

skurfer:
I can reproduce this in the listed problem interfaces. Can’t say that I know where to start fixing it, though.

eliasbe:
Switched interfaces and it works exept when creating triggers then it
crashes.

DavidShepherdson:
Sounds similar (in terms of the symptoms) to #2045? Does the workaround of moving the Quicksilver window to a non-retina display work for this issue too?

[...]

PR #2152, opened 2015-11-09
Fix for #2111

tiennou:
After much twiddling with debuggers and Flashlights, I think I got it.

The problem is that sometimes, the calculation of the containing scrollview doesn't end on an integral rect. [This](https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m#L746), returns this rect `
(origin = (x = 1.25, y = 1.25), size = (width = 57.5, height = 17.5))` which is then set as the scroll view frame, and this triggers a bug within the layout code where the text fields tries to adapt to that weird size but just ends up blowing up the stack because it's not pixel-aligned.

We might want check the various affected interfaces so this does not look too jarring visually.

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
@@ -746,7 +746,8 @@ NSMutableDictionary *bindingsDict = nil;
 		NSRect titleFrame = [self frame];
 		NSRect editorFrame = NSInsetRect(titleFrame, NSHeight(titleFrame) /16, NSHeight(titleFrame)/16);
 		editorFrame.origin = NSMakePoint(NSHeight(titleFrame) /16, NSHeight(titleFrame)/16);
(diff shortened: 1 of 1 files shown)


tiennou:
#2111 (because cannot click)

tiennou:
I'm also wondering about Retina, because it uses by definition non-integral rects. Thus I'm not sold on it being a bug in framework code or a bug on our end...

Also, why is there a scroll view here ;-).

skurfer:
The long international nightmare could be over!

I tried a few interfaces on Retina and non-Retina and don’t see any visual weirdness. More importantly, I don’t see any crashes. :fireworks:

> Thus I'm not sold on it being a bug in framework code or a bug on our end…

A recursive loop that ends in death is surely a bug, but if we can avoid triggering it…

> Also, why is there a scroll view here ;-).

I sort of wondered the same thing.

PR #2153, opened 2015-11-11
update deployment for all NIBs to 10.9

skurfer:
Should have done this when changing the deployment target.

The only one I didn’t update is `QSHelpersPrefPane.xib` because saving it (even untouched) crashes Xcode.


skurfer:
Scratch that. Xcode’s crash report included “An instance of NSWindowTemplate with object ID 5 did not archive its child (NSPopUpButtonCell) with an object of ID 144.”

There were two `NSPopUpButtonCell`s in that NIB that were unused. Removing them allowed me to save (and update) the file. So that’s all of them.

FYI, Travis appears to be using Xcode 6.x which is why it’s complaining. I didn’t see a way to change that.

pjrobertson:
Did you close this because of Travis not working?!

skurfer:
> Did you close this because of Travis not working?!

No, just because I realized when updating a plug-in that instead of choosing 10.9 explicitly, you can just make it match the `MACOSX_DEPLOYMENT_TARGET` automatically. So I did that instead, but forgot to open the new pull request. Stay tuned…

PR #2155, opened 2015-11-12
update NIBs to follow the project's deployment target

skurfer:
So, in the future, we should only need to change `MACOSX_DEPLOYMENT_TARGET`.


pjrobertson:
Travis is failing as you said. Can you update travis.yml?

See http://docs.travis-ci.com/user/languages/objective-c/#Supported-OS-X-iOS-SDK-versions

pjrobertson:
No problem actually - I realised it was something I could do without a Mac :-)

I'm not entirely happy with hard coding the XCode version into the travis.yml file, but I don't see any other solutions.

skurfer:
Good find. I can live with it.

pjrobertson:
I still can't test this as I have Xcode 6, and updating to 7 will take for ever from China. I assume it's not going to cause any problems though.

Do we have any open merge requests that update nibs, and so maybe we should look at merging those first (to avoid any merge conflicts)

skurfer:
> Do we have any open merge requests that update nibs, and so maybe we should look at merging those first (to avoid any merge conflicts)

Not a bad idea, since this one is very easy to repeat.

Currently, it looks like that would be

* #1395
* #1611, which was just rebased
* #2073, but I closed that since it seems to be a dead-end for now

skurfer:
I suppose we can just do this as NIBs are touched, or we can do another branch that doesn’t conflict with those listed. Either way, it won’t be this one.

PR #2160, opened 2015-11-17
Return nothing if there is nothing to search for

### LINKED ISSUE #2158: 1.3.3 (4015) crashes constantly
Since updating an hour or so ago, I can't keep Quicksilver running for more than half a minute or so. Entirely unusable. I have submitted crash reports.

I'm on 10.9.5

tiennou:
Fixes #2158.

### DIFF
--- a/Quicksilver/Code-QuickStepFoundation/NSMetadataQuery+Synchronous.m
+++ b/Quicksilver/Code-QuickStepFoundation/NSMetadataQuery+Synchronous.m
@@ -24,7 +24,7 @@
         search = [NSPredicate predicateFromMetadataQueryString:searchString];
     }
 	[self setPredicate:search];
-	if (![paths isEqual:nil]) {
+	if (paths) {
 		NSMutableArray *pathURLs = [NSMutableArray array];
 		for (NSString *path in paths) {
 			NSURL *pathURL = [NSURL fileURLWithPath:path];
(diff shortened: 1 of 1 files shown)


pjrobertson:
Nice, I see no issue with it. What happens if the search string is `@""`?

pjrobertson:
A quick question on this [line](https://github.com/quicksilver/Quicksilver/commit/dc0695a50f552586c5e25d512a016dd75ebc6acd#diff-16095b8a42e7d2cd504e89844a438a46R27), whilst we're in the file: why use `isEqual:nil`?

Also, do you think it's worth updating the whole file to spaces...? :fearful:

tiennou:
Haven't actually tried. I'd suppose it would not crash (the fixed crash happens because it does CFStringy stuff on `nil`), but it might not work correctly (returns immediately, maybe ?).

I think I'll ponder some more on that code. And write some tests ;-).

Spaces ? Or Tabs ? :laughing:

pjrobertson:
Tabs :D
(as much as I want spaces as well!)

skurfer:
> why use `isEqual:nil`?

Even I can’t answer that. Not sure what I was thinking. Maybe I copied and pasted it from somewhere and missed that test.

skurfer:
Added a commit to change that test @pjrobertson. If it looks good, merge. As for tabs, I don’t want to destroy all the history.

pjrobertson:
Yup fine. Sorry - 3 months later :(

PR #2162, opened 2015-11-21
Text preferences

skurfer:
Adds 4 new preferences for controlling the appearance of text in a `QSObjectCell`:

  * Underline matched text
  * Make matched text “glow” (a shadow that matches the details color, which is the text color faded by 20%)
  * Allow users to turn off “name ⟷ label” and just show the matched string (the original behavior). A gift for @hmelman and I’m sure others.
  * Allow users to disable text tightening. It causes letters to overlap with some fonts.

This is mostly motivated by some problems that were plaguing users, and my long-time hatred of the underlined letters. (I’ve been using “glow” for like a year locally.) I thought it made sense to do them all together. We could also add a color picker for matched text and some other options, but this is a good starting point.

Note that I updated the NIB’s deployment target.

The defaults for these prefs will mimic the existing behavior.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSPreferenceKeys.h
+++ b/Quicksilver/Code-QuickStepCore/QSPreferenceKeys.h
@@ -61,6 +61,7 @@ Following lines are no longer used in this project.
 #define kQSTextMatchedUnderline @"Text Matched Underline"
 #define kQSTextMatchedAlwaysShowName @"Text Matched Always Show Name"
 #define kQSTextAllowTightening @"Text Allow Tightening"
+#define kQSTextGlowDivisor @"Text Glow Divisor"
 
 #define kResultTableSplit @"Result Table Split Width"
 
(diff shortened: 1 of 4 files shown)


hmelman:
> On Nov 21, 2015, at 11:18 AM, Rob McBroom <notifications@github.com> wrote:
> 
> 	• Allow users to turn off “name ⟷ label” and just show the matched string (the original behavior). A gift for @hmelmanand I’m sure others.

Gifts don't go unnoticed. Thanks :)

Howard

pjrobertson:
Cool. Looks good. The code all looks fine apart from a few stylistic things. Gets my blessing if the dev previewers are also happy with it ;-)

skurfer:
> Maybe it could be made a hidden pref? Don't think it's that crucial to be a slider.

Done. I edited a line that was indented with spaces, hence the failure. Can’t hold that against me.

Sounds like you’re OK with all this, so I’ll merge tomorrow if you don’t object (or merge) first.

pjrobertson:
There's my confirmation :-)

I haven't tested it, but I assume that's what the dev preview is for

skurfer:
Reports from users that at least two of the prefs do what they were designed to do. (And I’ve been using it for weeks.) Thanks!

PR #2163, opened 2015-11-22
allow Quicksilver Selection to refer to other proxy objects

skurfer:
Not sure why this would have been prevented. The line above already protects against a circular reference.

I see that I’m the one that wrote it, but I was just rewriting a test that was already there.


skurfer:
@lgarron confirms that this is working as intended.

PR #2169, opened 2015-12-22
Fix check_indent false positives on comment blocks

pjrobertson:
Don't worry about indents inside comment blocks

I saw that the test on @skurfer's  `objectchosen` branch were failing because of bad spaces/tabs in comment blocks. This change allows spaces/tabs/anything in comment blocks, although I'm still not sure if we should do this. If we're enforcing one format over the other, then shouldn't we also do that inside comment blocks?

Discuss ;-)

### DIFF
--- a/Quicksilver/Tools/travis/check_indent.rb
+++ b/Quicksilver/Tools/travis/check_indent.rb
@@ -5,10 +5,16 @@ require 'set'
 
 changed_lines = `git diff $TRAVIS_COMMIT_RANGE -- *.[mh]`.split("\n").select { |l| l.start_with? "+" }
 
+comment_block = false
+
 changed_lines.each do |line|
-	if /^\+ /.match(line) then
+	if /^\+ /.match(line) and not comment_block then
 		puts "Invalid line #{line}"
 		exit 1
+	elsif /^\+\s*?\/\*\*/.match(line) then
+		comment_block = true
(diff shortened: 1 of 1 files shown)


skurfer:
I thought this was an easy one because:

1. Making exceptions would be a pain; especially for multi-line comments
2. The whole point of picking an indentation and sticking with it is to ensure that things stay lined up for readability. So… should *comments* be readable? :smiley:

But then I realized it probably tripped up on all the Doxygen stuff I added. Now I’m not so sure.

1. I haven’t looked for any official guidelines, but that seems to be the standard format for Doxygen comments (and maybe all multi-line comments in C).
2. Even if you tell Xcode to use tabs, typing /**↩ results in a line indented one space.

So maybe we should allow spaces there.

I’ve found that hitting tab inside a comment block even inserts one space, but if you hit tab twice, the space is removed and replaced with a tab. So it is possible and not *too* much of a pain to use tabs in comments. I can be convinced to just say tabs everywhere.

tiennou:
IMHO my own personal style guide says "Comments are indented with spaces", the rationale being that I have a tendency to *-prefix lines and OCD forces me keep them 1 space from the left gutter, like this :

```
/**
 * Some insightful comment on the meaning of life.
 */
```

I don't think tabs would pass the OCD test here :laughing:. My take on this would be ignore indentation in comments altogether.

Issue #2176, opened 2016-01-26
Address Sanitizer


tiennou:
Go to Scheme > Diagnostics > Enable Address Sanitizer, Run. Enjoy :heart_eyes: !

tiennou:
That's *2 complete seconds after startup*.

```
=================================================================
==7734==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x700000115664 at pc 0x00010013f9cd bp 0x7000001155b0 sp 0x700000114d68
READ of size 7 at 0x700000115664 thread T3
    #0 0x10013f9cc in wrap_strlen (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/7.0.0/lib/darwin/libclang_rt.asan_osx_dynamic.dylib+0x3d9cc)
    #1 0x7fff89ff1bae in +[NSString stringWithCString:encoding:] (/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation+0x3cbae)
    #2 0x100eb02eb in -[NSString(Hex) decodedHexString] NSString_BLTRExtensions.m:254
    #3 0x100eaf027 in -[NSString(Hex) decodedPasteboardType] NSString_BLTRExtensions.m:241
    #4 0x1010a22e9 in -[QSObject(Pasteboard) addContentsOfPasteboard:types:] QSObject_Pasteboard.m:99
    #5 0x1010a2ced in -[QSObject(Pasteboard) initWithPasteboard:types:] QSObject_Pasteboard.m:118
    #6 0x1010a1682 in -[QSObject(Pasteboard) initWithPasteboard:] QSObject_Pasteboard.m:80
    #7 0x1010a000d in +[QSObject(Pasteboard) objectWithPasteboard:] QSObject_Pasteboard.m:50
    #8 0x108e886e7 in 0x0000000108e886e7 (/Users/tiennou/Library/Application Support/Quicksilver/PlugIns/com.blacktree.Quicksilver.QSClipboardPlugIn.142.qsplugin/Contents/MacOS/Clipboard Plugin+0x56e7)
    #9 0x100144903 in __wrap_dispatch_async_block_invoke (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/7.0.0/lib/darwin/libclang_rt.asan_osx_dynamic.dylib+0x42903)
[...]

Issue #2020, opened 2015-01-24
Freeze on "Open With..." action


ar-rm:
### Steps:
- QS
- Command-G
- Tab
- Type "w" to get "Open With..." action. (Or press the down-arrow to reveal all of the potential actions. I can move through the list until I hit "Open With...".)

### Result:
- QS Freezes.
- I have to force quit it.
- After the force quit/restart of QS, the following is logged to system.log (probably not related):
```
Jan 24 12:44:24 com.apple.xpc.launchd[1] (com.blacktree.Quicksilver.131808[6515]): Service exited due to signal: Killed: 9
```
This is a crashlog from around the time of the crash, but it's possible that it's from another crash and not related: https://gist.github.com/ar-rm/0442c3e57fb873ca3905

### Versions:
QS: 1.2.2 (4011)
OS: 10.10.1

Relevant (maybe): I was previously using QS 1.0.0 (4000) and this wasn't an issue... But other things started to be, so I upgraded to 1.2.2 and this issue started. The exact same thing happened with another 10.10.1 install I was testing on an external HD. I upgraded QS to 1.2.2 and this started.

Which Plug-In does "Open With..." come from? I'd like to try disabling it. Or is it built in?

I found issue #1783, but don't know if it's actually related.

Thanks.

skurfer:
Do you have anything in your catalog that might be an alias to a file on an unreachable network drive?

ar-rm:
I doubt it, but that's a good thought.

I went through all of the catalogs, disabled each of them per major group (i.e.: Applications, Plugins, System, etc.) and ran a "Catalog Rescan" after each. Same result. The only slight difference is the Actions popup appears (with "Open" at the top and "Open With..." just below). But it's still frozen. I don't know why that changed.

I updated to 10.10.2 yesterday as well.

skurfer:
Does this still happen with Quicksilver 1.3.4?

ar-rm:
I don't know. Sometime shortly after this post I reverted to OS 10.9.5 and QS 1.0.0. I've been hessitant to update since :-)

You think 1.3.4 (or some other earlier version) resolved this?

skurfer:
If it was a network drive, yes. Otherwise, I’m not sure, but a lot has changed so it might be worth a look.

ar-rm:
I just upgraded to 1.3.4 (4016).

Same result.
- QS freezes
- The only way out is to force quit

Some other, hopefully useful, info:
[...]

PR #2177, opened 2016-01-26
Webloc crash

tiennou:
- A straightforward fix for #2176.

- #2020 :

  I'm actually cheating : `-pasteboardByFilteringContents:` doesn't actually do anything since I-cant-even-begin-to-guess-when so I'm just dispatching something that doesn't work to the main thread so it does nothing later™.

  This code was responsible for extracting contents of `.textClipping` files (which ~~are stored in resource forks~~ use dark voodoo magic). I know you could use it to right-arrow into text clippings because I played with that looong ago, but I can't seem to get it to work now. I suspect 64bit ? PPC ? I'm also really not sure what it's expecting to get back anyway... @hmelman does that ring a bell to you ?

- `LSCopyItemInfoForURL` : the commit says it all. I unfortunately got the crashlog only once. I think I got the explanation right though : the autorelease pool would get drained at the end of a scan "cycle", and since I think it does it's think in another thread, you'd get a crash.


pjrobertson:
@tiennou - indentation... :S
This is the first test my check_indent script has had.

Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
```diff
@@ -600,8 +600,11 @@ + (QSObject *)fileObjectWithPath:(NSString *)path {
     // initWithArray: deals with file objects that already exist
     QSObject *newObject = [[QSObject alloc] initWithArray:[NSArray arrayWithObject:path]];
     
-	if ([clippingTypes containsObject:[[NSFileManager defaultManager] typeOfFile:path]])
-		[newObject performSelectorOnMainThread:@selector(addContentsOfClipping:) withObject:path waitUntilDone:YES];
+    if ([clippingTypes containsObject:[[NSFileManager defaultManager] typeOfFile:path]]) {
```
pjrobertson: indentation?!

tiennou:
Hope I was quick enough that my failed rebase wasn't visible ;-). Looks like Travis saw me though...
I've made some changes to the check_indent script so it's not as painful. It could be better tough but at least you don't end up doing loads of little commits ;-).

I'll add a check for the CFRelease stuff though.

tiennou:
Yeah, you can also use it locally now, it'll just use `master` as a "default" diff base. It would be nice if it kept the diff header ;-). And then it would be ready to be promoted to commit-hook status ;-).

In case that wasn't clear I added the tab-switching stuff here (Xcode + TextMate).

tiennou:
(Also, w.r.t to your lost comment about thread safety, it's actually `NDResourceFork` that's not threadsafe. I think `pasteboardByFilteringContents:` is actually fine too...

Anyway, right now this code is dead, I've just kept in because 1) I wasn't sure if it was readily fixable, and 2) I'm not even sure what can be done about `NDResourceFork`...

hmelman:
> On Jan 25, 2016, at 8:57 PM, Etienne Samson <notifications@github.com> wrote:
> 
> 	• This code was responsible for extracting contents of .textClipping files (which are stored in resource forks use dark voodoo magic). I know you could use it to right-arrow into text clippings because I played with that looong ago, but I can't seem to get it to work now. I suspect 64bit ? PPC ? I'm also really not sure what it's expecting to get back anyway... @hmelman does that ring a bell to you ?

Not really sure. The Text Manipulation Plugin allows you to right-arrow into a text file and get back a results list with an item for each line. I haven't used it in all long time. Just trying it now, it seems to work right-arrowing into a .txt file.

Howard

tiennou:
Yeah, that's right. I think that resource-forky stuff tries to enable the same behavior with clippings. I might be conflating the two then...

pjrobertson:
@tiennou  - I just cherrypicked your 'smarter script for local deployment' into master.
My first PR since that travis test also just failed because of the same reason :S

Haha!

Quicksilver/Code-QuickStepCore/QSObject_Pasteboard.m
```diff
@@ -80,7 +80,7 @@ - (id)initWithPasteboard:(NSPasteboard *)pasteboard {
 	return [self initWithPasteboard:pasteboard types:nil];
 }
 
-- (void)addContentsOfClipping:(NSString *)path { // Not thread safe?
```
pjrobertson: why remove the comment if it's true?

PR #2178, opened 2016-01-26
Don't mount volumes when resolving aliases

tiennou:
I think this is going to be controversial. This makes QS never try to mount network volumes when resolving aliases, as right now we have no way to discern whether this is a "user-initiated" resolution, or some background scan.

Thinking about it, I should have split that... The second part makes the resolution fail completely instead of returning itself. Say you're browsing some folder containing one of those remote aliases, if you right-arrowed (and had the patience to wait for it to fail), you'd end up with a new result set with only that alias, instead of the expected bump.

### DIFF
--- a/Quicksilver/Code-QuickStepFoundation/NSFileManager_BLTRExtensions.m
+++ b/Quicksilver/Code-QuickStepFoundation/NSFileManager_BLTRExtensions.m
@@ -126,28 +126,27 @@
 
 - (NSString *)resolveAliasAtPath:(NSString *)aliasFullPath usingUI:(BOOL)usingUI {
 	NSURL *url = [NSURL fileURLWithPath:aliasFullPath];
-    
-    // First resolve any symlinks
-    NSURL *resolvedURL = [url URLByReallyResolvingSymlinksInPath];
-    
-    // File is a Finder alias file, resolve bookmark data first
-    NSError *err;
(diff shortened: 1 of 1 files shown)


pjrobertson:
we already have the `withUI` flag so to me what you've done seems fair enough. I'd like to know if right arrowing does or doesn't set `withUI = True` though. Hopefully it would.

tiennou:
Depends. `withUI:` is only used by the `openFile:` handler, so no right arrow. I guess we should just weight the damage that a lone remote alias can do vs. the amount of reports we've got because of stuck Quicksilvers. As a corollary, if I had 10 aliases to 10 different, accessible remote locations, I would get all 10 of them mounted *on each scan*...

skurfer:
> I think this is going to be controversial.

I don’t. :smiley:

skurfer:
So is this the *real* fix for #2020 or just revisiting #1977? Both?

tiennou:
It's a real "fix" for the deadlock in #2020, which is triggered by resolving an alias to a recognizable `clippingTypes` UTI. It's related to #1977 because also aliases, but this one was scanning-related right ? #2020 happens when using the interface (actually, when you move over an indirect-allowed action with an alias to a `clippingTypes`-compatible object).

tiennou:
Err, you lost me sir :laughing:. That comment above belongs to #2177 ;-). This one is just some more fixes over #1977 because I took a bunch of aliases to some remote workplace stuff, and I actually checked how QS handled those (the response is not nicely).

PR #2183, opened 2016-02-05
Fix #2182

### LINKED ISSUE #2182: Changing monitors makes window appear out of the screen
Changing monitors makes window appear out of the screen. Even if quit it, and restart, it still although less, has the top bar with the zoom, etc, buttons out of reach.

tiennou:
This is me looking at the interface code. If only bug-fixes were always this easy... :laughing:


skurfer:
I assume you’ve had a chance to try this at work? Feel free to merge if it works as expected.

PR #2190, opened 2016-03-15
call the local setObjectValue when clearing

skurfer:
I’m not sure of all the consequences, since the two implementations of `setObjectValue:` are pretty different. But since all we’re doing is wiping things out, the version in `self` seems to make more sense.


pjrobertson:
Cool. I think the key here is calling `[self clearSearch];` which in turn invalidates the timer.
This'll need some hands on testing though to check for any side effects.

skurfer:
> I think the key here is calling [self clearSearch]; which in turn invalidates the timer.

Exactly. I was going to just call that, but noticed that `[self setObjectValue:`] already did it.

> This'll need some hands on testing though to check for any side effects.

Yeah, I’m running with it already.

PR #2191, opened 2016-03-21
Fixes for Recents and Favorites

### LINKED ISSUE #2187: Favorite Servers and Recent Items catalog sources moved in El Capitan
The "Favorite Servers" catalog entry reads from `~/Library/Preferences/com.apple.sidebarlists.plist`

It appears these settings are now stored in `~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.FavoriteServers.sfl`. I pulled this file up in BBEdit and that's where they are.

The "Recent Items" catalog entries (recent Apps/Docs/Servers/Open and Save Folders) have also been moved into sfl files in this directory.

References here:
https://ryanmo.co/2015/10/31/list-server-favorites-in-os-x-1011-el-capitan/
https://gist.github.com/pudquick/4776b4b2075bf9b7e512
And unpacking El Capitan's new sfl files:
http://michaellynn.github.io/2015/10/24/apples-bookmarkdata-exposed/

Hope that helps contributors get a jump on the issue and parsing this new data structure. I'm not sure if this info can be easily read from a Defaults Reader or not, documentation seems poor.

QS 1.3.4 (4016)
OS X 10.11.4 Beta (15E61b)

skurfer:
Updates the following catalog presets to work in …whatever version of OS X broke them.

* Finder Sidebar Items
* Recent Applications
* Recent Documents
* Recent Servers
* Favorite Servers

This surely breaks those things for older OS X versions, but this crap changes so often, I wonder if they were even working in the first place. :smiley:

fixes #2187

### DIFF
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
@@ -20,7 +20,8 @@
 	}
 	NSString *path = [sflPath stringByStandardizingPath];
 	NSFileManager *manager = [NSFileManager defaultManager];
-	if (![manager fileExistsAtPath:path isDirectory:NULL]) {
+	BOOL isDir = NO;
+	if (![[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:&isDir] || isDir) {
 		return YES;
 	}
 	NSDate *modDate = [[manager attributesOfItemAtPath:path error:NULL] fileModificationDate];
@@ -36,19 +37,24 @@
 	NSMutableArray *sflItemArray = [NSMutableArray arrayWithCapacity:0];
 	NSString *sflPath = [settings objectForKey:kItemPath];
 	NSString *path = [sflPath stringByStandardizingPath];
-	if ([[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:nil]) {
-		NSDictionary *sflData = [NSKeyedUnarchiver unarchiveObjectWithFile:path];
-		for (SFLListItem *item in sflData[@"items"]) {
-			// item's class is SFLListItem
-			if ([item URL]) {
-				[sflItemArray addObject:item];
-			}
-		}
-		[sflItemArray sortUsingComparator:^NSComparisonResult(SFLListItem *item1, SFLListItem *item2) {
-			return item1.order > item2.order;
-		}];
-		
+	BOOL isDir = NO;
+	if (![[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:&isDir] || isDir) {
+		return nil;
 	}
+	NSDictionary *sflData = [NSKeyedUnarchiver unarchiveObjectWithFile:path];
+	NSString *kItems = @"items";
+	if (![[sflData allKeys] containsObject:kItems]) {
+		return nil;
+	}
+	for (SFLListItem *item in sflData[kItems]) {
+		// item's class is SFLListItem
+		if ([item URL]) {
+			[sflItemArray addObject:item];
+		}
+	}
+	[sflItemArray sortUsingComparator:^NSComparisonResult(SFLListItem *item1, SFLListItem *item2) {
+		return item1.order > item2.order;
+	}];
 	return [sflItemArray arrayByEnumeratingArrayUsingBlock:^id(SFLListItem *item) {
 		NSURL *url = [item URL];
 		if ([url isFileURL]) {
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
@@ -13,7 +13,21 @@
 
 - (BOOL)indexIsValidFromDate:(NSDate *)indexDate forEntry:(NSDictionary *)theEntry
 {
-	return NO;
+	NSDictionary *settings = [theEntry objectForKey:kItemSettings];
+	NSString *sflPath = [settings objectForKey:kItemPath];
+	if (!sflPath) {
+		return YES;
+	}
+	NSString *path = [sflPath stringByStandardizingPath];
+	NSFileManager *manager = [NSFileManager defaultManager];
+	if (![manager fileExistsAtPath:path isDirectory:NULL]) {
+		return YES;
+	}
+	NSDate *modDate = [[manager attributesOfItemAtPath:path error:NULL] fileModificationDate];
+	if ([modDate compare:indexDate] == NSOrderedDescending) {
+		return NO;
+	}
+	return YES;
 }
 
 - (NSArray *)objectsForEntry:(NSDictionary *)theEntry
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
@@ -1,0 +1,47 @@
+//
+//  QSSharedFileListSource.m
+//  Quicksilver
+//
+//  Created by Rob McBroom on 2016/03/16.
+//
+//
+
+#import "QSSharedFileListSource.h"
+#import "QSFoundation.h"
+
+@implementation QSSharedFileListSource
+
+- (BOOL)indexIsValidFromDate:(NSDate *)indexDate forEntry:(NSDictionary *)theEntry
+{
+	return NO;
+}
+
+- (NSArray *)objectsForEntry:(NSDictionary *)theEntry
+{
+	NSDictionary *settings = [theEntry objectForKey:kItemSettings];
+	NSMutableArray *sflItemArray = [NSMutableArray arrayWithCapacity:0];
+	NSString *sflPath = [settings objectForKey:kItemPath];
+	NSString *path = [sflPath stringByStandardizingPath];
+	if ([[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:nil]) {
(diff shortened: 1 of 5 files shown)


pjrobertson:
> This surely breaks those things for older OS X versions, but this crap changes so often, I wonder if they were even working in the first place.

I'm sure they used to. We used to do things like `Favourite Servers (10.6)` in the catalog but that was horrible. I'm fine with only supporting the latest version.

FYI the `combined` unit tests are failing. From the Travis logs:

```
/Users/travis/build/quicksilver/Quicksilver/Quicksilver/Tests/Tests-QSCore/TestQSObject.m:225: ((combined1) not equal to (combined2)) failed: ("<4019415a d17f0000>") is equal to ("<4019415a d17f0000>"):
222 	combined2.identifier = @"isEqualTest";
223 	// make sure they aren't literally the same object
224 	// otherwise, the next test would be pointless
225 	XCTAssertNotEqual(combined1, combined2);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
226 	XCTAssertEqualObjects(combined1, combined2);
227 	QSObject *data3 = [QSObject URLObjectWithURL:@"https://qsapp.com/" title:@"QS"];
```

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
```diff
@@ -0,0 +1,47 @@
...
+
+- (NSArray *)objectsForEntry:(NSDictionary *)theEntry
+{
+	NSDictionary *settings = [theEntry objectForKey:kItemSettings];
+	NSMutableArray *sflItemArray = [NSMutableArray arrayWithCapacity:0];
+	NSString *sflPath = [settings objectForKey:kItemPath];
+	NSString *path = [sflPath stringByStandardizingPath];
+	if ([[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:nil]) {
+		NSDictionary *sflData = [NSKeyedUnarchiver unarchiveObjectWithFile:path];
+		for (SFLListItem *item in sflData[@"items"]) {
```
pjrobertson: cool, nice to see `SFLListItem.h` coming in handy here :)

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
```diff
@@ -0,0 +1,47 @@
...
+}
+
+- (NSArray *)objectsForEntry:(NSDictionary *)theEntry
+{
+	NSDictionary *settings = [theEntry objectForKey:kItemSettings];
+	NSMutableArray *sflItemArray = [NSMutableArray arrayWithCapacity:0];
+	NSString *sflPath = [settings objectForKey:kItemPath];
+	NSString *path = [sflPath stringByStandardizingPath];
+	if ([[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:nil]) {
+		NSDictionary *sflData = [NSKeyedUnarchiver unarchiveObjectWithFile:path];
```
pjrobertson: A few concerns:

* can we always be sure that it'll unarchive properly? (Shared File Lists are undocumented right?)
* Will it always return an NSDict?
* Will the NSDict always contain a key `items`?

I think it might be worth adding some ifs() for those cases

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
```diff
@@ -13,7 +13,21 @@ @implementation QSSharedFileListSource
...
+	NSString *path = [sflPath stringByStandardizingPath];
+	NSFileManager *manager = [NSFileManager defaultManager];
+	if (![manager fileExistsAtPath:path isDirectory:NULL]) {
+		return YES;
+	}
+	NSDate *modDate = [[manager attributesOfItemAtPath:path error:NULL] fileModificationDate];
+	if ([modDate compare:indexDate] == NSOrderedDescending) {
+		return NO;
+	}
+	return YES;
```
pjrobertson: Adding on to my TODO list: pull this code out and put it in its own method. We must call this tens of times in Catalog entries! Would be much easier just to do something like

```
return [self indexIsValidFromDate:indexDate forPathEntry:theEntry];
```

then just have this boilerplate code once.

pjrobertson:
All looks good, I like the new object source. Happy to merge and pre-release once the travis issue has been sorted

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
tiennou: > - can we always be sure that it'll unarchive properly? (Shared File Lists are undocumented right?)

Agreed, and we sure don't like NSExceptions ;-).

Additionally, I'd prefer the `if` just above written as such : 
```
BOOL isDir = NO;
if (![[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:&isDir] || isDir)
    return nil; // Maybe ?
```
but I might be nit-picking ;-).

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
tiennou: One day I'll publish my rewritten object source ;-).

pjrobertson: @tiennou I feel like we're both getting too conservative in our old age.
Scared of all these NSExceptions and ifs :P

skurfer:
> FYI the `combined` unit tests are failing.

As the comment says, removing that test would make the rest pointless. I’m inclined to remove `testEquality` completely. It’s nothing but frustration.

The test only fails if those are literally the same object at the same memory address, and they can’t be, so… :hurtrealbad: :gun:

I did see the same failure locally at times. It seemed to fail the first time, but would pass after that, adding to the confusion. I think maybe Travis starts with a clean environment every time, so it’s _always_ the “first” run. Then again, it seems like Travis has also been ping-ponging between success and failure when there are no meaningful changes, so maybe not. (I’m pretty sure it passed on this branch when it was created, then failed when I filled out `indexIsValid`.)

Thoughts?

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
skurfer: > Adding on to my TODO list: pull this code out and put it in its own method. We must call this tens of times in Catalog entries!

Yes, please. But I’d rather have something that just takes a date and a path, rather than the entry dictionary. That would let us use it in other contexts, and doesn’t assume the path will always be at `settings → path`.

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSSharedFileListSource.m
skurfer: > can we always be sure that it'll unarchive properly? (Shared File Lists are undocumented right?)
> Will it always return an NSDict?
> Will the NSDict always contain a key items?

All of the real-world examples we have access to are like that. That might be as much certainty as we’re going to get.

I’ll add some checks based on both of your comments.

tiennou:
"What the hell is that test doing" is my first thought ;-). `obj.property = value` is syntactic sugar for `[obj setProperty:value]`, so that clearly doesn't bypass anything. Also, I don't get what the goal is by not having unique identifiers, they are *expected* to be unique.

But apart from that, yes, it looks like some intermittent test failure with combined objects. Also, am I the only one where testFileNaming fails (the second and fourth tests) ?

tiennou:
The merging intermittently fails on `if (objectCount == 1)` (2nd return). It looks like something goes wrong in the code above...

skurfer:
> "What the hell is that test doing" is my first thought ;-).

The short answer is “Making sure `-[QSObject isEqual:]` works like it’s supposed to.”

You’re right that we won’t normally see objects with the same identifier. The tests are faking it to reach the rest of the code. The first things that method checks are literal equality (by memory address) and identifier equality. So the test attempts to set up two distinct objects with the same identifier to bypass those checks so we can test the stuff deeper down.

HenningJ:
> You’re right that we won’t normally see objects with the same identifier. The tests are faking it to reach the rest of the code.

I haven't been following the rest of the discussion, but...Tests that are testing stuff that only happens in tests...seems wrong to me.

tiennou:
> and identifier equality.

Only if identifiers have been [set](https://github.com/quicksilver/Quicksilver/blob/19c8ca9825d61df8a9a03bba554322f9a03a854b/Quicksilver/Code-QuickStepCore/QSObject.m#L85). If they aren't set, the rest of the code is reached. If they are set, and they are the same, the rest of the code is *not* reached. Hence my "first thought" above.

Also, shouldn't that test be `if ([self identifier] && [other identifier] && [self.ID isEqualToString:otherID]) return YES;` ?

Anyway, as long as you don't ask the Librarian, there's nothing that prevents you from creating objects with the same identifier — things will start to get messy once the Librarian is involved.

Here's an [excerpt](https://gist.github.com/tiennou/0ce833631062d6650ea0) of 2 runs I did, the first succeeds, the other does not. Notable places are "test combined 1st" and "test combined 2nd", you can see that the second time around, `setOfObjects` considered our two *equal* objects to be equal (which is tested just above and expected to be true), while the first one considered our two *equal* objects to be different, hence skipping the count test. :fire:

skurfer:
> Tests that are testing stuff that only happens in tests...seems wrong to me.

It doesn’t just happen in the tests. It happens in practice, too (which is why there are no longer duplicates in the history). That’s just the way the tests trigger the other code.

skurfer:
Safety checks added.

PR #2193, opened 2016-03-31
Reduce Mutables

pjrobertson:
As promised, a pull to reduce the use of mutable objects. First for `splitObjects`, then for various arrays stored for the search object view. Only the `resultArray` really needs to be mutable as that's what is changed as you search

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -153,7 +153,10 @@ NSSize QSMaxIconSize;
 	NSMutableArray *array;
 	// Set used to keep track of the objects already added
 	NSMutableSet *setOfObjects = [[NSMutableSet alloc] init];
-	
+
+	// Make sure objects is immutable
+	objects = [objects copy];
+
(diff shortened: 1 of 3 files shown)


skurfer:
Merged this into my local build. I’ll try it out for a while.

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -153,7 +153,10 @@ + (id)objectByMergingObjects:(NSArray *)objects {
 	NSMutableArray *array;
 	// Set used to keep track of the objects already added
 	NSMutableSet *setOfObjects = [[NSMutableSet alloc] init];
-	
+
+	// Make sure objects is immutable
+	objects = [objects copy];
```
skurfer: I don’t think this matters anywhere except line 196. You could just change that to

    [object setObject:[objects copy] forCache:kQSObjectComponents];

and avoid making the copy in some cases. Not a big deal, though.

pjrobertson: I did it here in the rare case that `objects` might be mutated on the next line as it's being enumerated

PR #2204, opened 2016-05-06
Update history and mnemonics when dragging

skurfer:
Used the “informally” deprecated method we were already using. I suppose we should update those one day.


pjrobertson:
Nice, looks good - except for our still-failing tests! :'(

Issue #2214, opened 2016-06-14
I don't see the toggleApplication action


wmertens:
According to https://github.com/quicksilver/Quicksilver/blob/4203d087d2076ff3adfdb24cb3d036efeecc5b00/Quicksilver/PlugIns-Main/QSCorePlugIn/QSCorePlugIn-Info.plist#L345-L366 there should be an action that shows/hides an application, but I don't see any way to trigger it, it doesn't show up anywhere afaict.

skurfer:
Note the `enabled: false` on lines 356 and 357. That means it’s not enabled by default, but you can turn it on by going to Preferences → Actions.

wmertens:
Oh wow, that was well hidden :)

Could I request then that this action is enabled by default? It seems to work really well for me…

hmelman:
Someone want to update the manual to say as of 1.4.2 the Toggle Application is no long disabled by default :)

skurfer:
> Someone want to update the manual to say as of 1.4.2 the Toggle Application action is no long disabled by default

Done

PR #2215, opened 2016-06-15
Enable Toggle Application by default

wmertens:
It seems like an innocent action and a useful addition, especially for trigger actions.

See #2214


HenningJ:

skurfer:
Seems reasonable.

Issue #178, opened 2011-04-12
Add "Sort by Date" & "Sort by Size" menu items to results menu with keyboard shortcuts


sinewave:
Update: I have edited this issue with an updated title and description to reflect what i think would be a better implementation

From Comment #3:
"The best way to implement this i think would be to add "Sort by Date" and "Sort by Size" menu items to the results menu which already has sort by name, score. You would also need to assign each of the sort options a keyboard shortcut similar to Task Viewer, Clipboard, or any other menu item.

This also makes the workflow a lot easier in that you dont have to decide how you want to sort the results before you display them which i always found to be in clunky in other launchers. Instead you get your results and then you can choose to sort them any way you want by clicking on the appropriate menu item or hitting the keyboard shortcut. This would also allow you to change sort options on the fly instead of having to go back and then use a diff modifier key to enter the result set with a new sort option."

Original Comment:
Ability to sort folder contents by modified, size, name, etc. by pressing an additional key when right arrowing into a folder similar to how showing hidden files by Option-arrowing into a folder works. Extremely useful for things like downloaded files, etc..

joshgold22:
Since this is mainly useful for file folder scanning, I think we'd want to distinguish between date added, created and modified.  E.g., one version would make is very quick to open a file I just downloaded to my Downloads folder.  Another would make it easy to pull up the files I'm working on in a large programming project.

petonic:
Has there been any update to this?  I'd really enjoy sorting by modification or created date in the results.  It would help staying in QS mode instead of going to Finder.

skurfer:
I’ve always been against this because so many things don’t even have dates or sizes.

That doesn’t mean it won’t happen. I only speak for myself and this project is open to anyone (and there are other regular contributors). I just want everyone to be aware that I have no plans to ever look at this. If this feature really matters, you might need to do more than ask for it. :smiley:

petonic:
Fair point, Rob.  How about this?  If you do that, or better yet (perhaps), expand the Spotlight Plugin's "Smart Folder" functionality to yield results (sorted by ModTime?) then I'll donate $50 to you or to the QuickSilver cause.

To be fair, it's a win win for me, because I should've already donated the money.   <<<SCRATCH THAT>>>  I just donated $50 to donate@qsapp.com.  I love this utility.

In any case, to go back to my specific use case...

[...]

PR #2220, opened 2016-06-29
Results sorting

tiennou:
#178

This is some preliminary changes to make it easier to add result sorting options. Right now I've added (Modification) Date, but Size would follow the same logic.

I tried to keep it generic (so providers can err provide whatever date they want as a modification date). Size might be more convoluted in that regard.

Also, a rescan will fix the missing metadata on scanned objects, but the top-level Catalog entries required a manual index-deletion step. YMMV.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
@@ -684,6 +684,7 @@ NSArray *recentDocumentsForBundle(NSString *bundleIdentifier) {
 		[self setIdentifier:thisIdentifier];
 		[self setPrimaryType:QSFilePathType];
 		[self getNameFromFiles];
+		[self setObject:[[self infoRecord] objectForKey:NSURLContentModificationDateKey] forMeta:kQSObjectModificationDate];
 	}
 	return self;
 }
@@ -702,7 +703,7 @@ NSArray *recentDocumentsForBundle(NSString *bundleIdentifier) {
     NSError *err = nil;
     
     // Note: NSURLLocalizedLabelKey gives a localized string of the Finder label (tag in 10.9+). E.g. Red / Rouge
-    dict = [fileURL resourceValuesForKeys:@[NSURLTypeIdentifierKey, NSURLIsDirectoryKey, NSURLIsAliasFileKey, NSURLIsPackageKey, NSURLLocalizedLabelKey, NSURLLocalizedNameKey, NSURLVolumeURLKey, NSURLIsExecutableKey, NSURLIsWritableKey, NSURLLabelColorKey] error:&err];
+    dict = [fileURL resourceValuesForKeys:@[NSURLTypeIdentifierKey, NSURLIsDirectoryKey, NSURLIsAliasFileKey, NSURLIsPackageKey, NSURLLocalizedLabelKey, NSURLLocalizedNameKey, NSURLVolumeURLKey, NSURLIsExecutableKey, NSURLIsWritableKey, NSURLLabelColorKey, NSURLContentModificationDateKey] error:&err];
     NSMutableDictionary *mutableDict = [dict mutableCopy];
 
     if (!dict) {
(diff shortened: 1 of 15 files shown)


Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
```diff
@@ -684,6 +684,7 @@ - (id)initWithArray:(NSArray *)paths {
 		[self setIdentifier:thisIdentifier];
 		[self setPrimaryType:QSFilePathType];
 		[self getNameFromFiles];
+		[self setObject:[[self infoRecord] objectForKey:NSURLContentModificationDateKey] forMeta:kQSObjectModificationDate];
```
tiennou: I'm not fond of that*, if someone has a better place for that to be in, I'll take it.

** I *hate* this category BTW 😉 .

skurfer:
I’ll take a look at this when I can. I have sort-by-date working in a local branch, too. 😛

Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
skurfer: In my version, I created a `fileModificationDate` method, so it would act like all the other file properties. I also added a `datetime` property to QSObject, but of course it was stored in `objectForMeta` too so the indexes on disk would preserve it.

tiennou:
Sorry for the duplicated work 😟 . Choose the one you prefer in that case, even though I tried to cleanup some of the interface code while doing so — or, arguably, littering everything with FIXMEs and TODOs 😆 .

Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
```diff
@@ -1103,14 +1096,14 @@ - (BOOL)handleChangeSearchModeEvent:(NSEvent *)theEvent {
...
-            case SearchFilter:
-                aNewSearchMode = changeSearchModeRight ? SearchSnap : SearchFilterAll;
+            case QSSearchModeFilter:
+                aNewSearchMode = changeSearchModeRight ? QSSearchModeSnap : QSSearchModeAll;
                 break;
-            default:
-                aNewSearchMode = changeSearchModeRight ? SearchFilterAll : SearchFilter;
+            case QSSearchModeSnap:
+                aNewSearchMode = changeSearchModeRight ? QSSearchModeAll : QSSearchModeFilter;
                 break;
```
skurfer: No default for this `switch` statement now?

skurfer:
Finally looked through this. I love all the clean-up and modernization of some of the classes.

Knowing what I’ve done, I think I’ll cherry pick as much as I can from this, with two main differences:

1. I’ll probably use my changes to QSObject and its file handling category. (The way it’s implemented shouldn’t be “welded to file-systems”.)
2. I’ll use sort descriptors, unless there’s some reason not to. It let’s us get rid of so much junk. See https://github.com/skurfer/Quicksilver/commit/3c8f5b574e5b443337e0a9c998371fc22ef182ec Thoughts on that?

Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
tiennou: Because the compiler was happy ;-). And you'll get a warning about a "missing case statement" whenever the enum changes, which felt better than silence.

tiennou:
+:fire: for `NSSortDescriptors`. It seems like a perfect use for them (instead of a bunch of wacky overrides all over some forwarding classes). Just be sure it behaves correctly in case the `key` is missing.

tiennou:
I'll refrain from taking a look at your branch for now (unless you really want me to, there are a few things I'm uncomfortable with *already* :wink:), but I'll gladly do reviews when you're done. Whatever you prefer.

skurfer:
> I'll refrain from taking a look at your branch for now

Yeah, don’t bother for now.

I’ve got a copy of this branch locally to work from. Closing this.

Issue #2071, opened 2015-07-03
Problems with three-pane AppleScript actions


Sesquipedalian:
AppleScript actions that take an optional indirect object always end up receiving a randomly selected indirect object even if the user didn’t select anything in the third pane.

To reproduce, try the following:
```
using terms from application "Quicksilver"
 on process text thetext with args
 return "1. " & thetext & " 2. " & args
 end process text
 
 on get argument count
 return 3
 end get argument count
 
 on get direct types
 return {"NSStringPboardType"}
 end get direct types
 
 on get indirect types
 return {"NSStringPboardType"}
 end get indirect types
end using terms from
```

I have also found that the third pane will have random stuff in it for AppleScript actions that require a third pane selection, like so:
```
using terms from application "Quicksilver"
 on process text thetext with args
 return "1. " & thetext & " 2. " & args
 end process text
 
 on get argument count
 return 2
 end get argument count
 
 on get direct types
 return {"NSStringPboardType"}
 end get direct types
 
 on get indirect types
 return {"NSStringPboardType"}
 end get indirect types
end using terms from
```

It looks to me like there are two issues happening here:

1. AppleScript actions that have an optional third pane should always have the third pane empty by default, but they don’t.

2. For compiled three-pane actions, QS will pre-populate the third pane with the highest ranking catalog item that matches the action’s criteria for the third pane. This is possible because the compiled actions have to declare this criteria in their Info.plist (or somewhere like that), enabling QS to know it in advance. But for AppleScript-based actions, there is no way for QS to know the criteria in advance; it has to be discovered at run-time. So it just grabs something on the fly. (Perhaps the most recently indexed catalog item that matches the criteria, but who knows?) **This affects all AppleScript actions that use the third pane, regardless whether the third pane is required or optional.**

A good solution for the first problem would be to change the code to ensure that the third pane is always cleared the moment that an AppleScript-based action with an optional third pane is selected in the second pane. 

The best solution to the second problem would be to ensure that QS always grabbed the top-ranked item for whatever criteria the AppleScript action defined for its third pane. Alternatively, a quick and dirty fix would be to clear the third pane any time *any* three-pane AppleScript action is selected in the second pane. That would be a bit less useful, since the user would always need to type something into the third pane. But since we have to do that anyway with the current situation, it would still be a step forward. At least we’d be forcing the user to make a good selection instead of suggesting a terrible selection.

PR #2232, opened 2016-07-18
Applescript third pane

pjrobertson:
Fixes the stuff in #2071 

Thanks @Sesquipedalian for reporting this and and also Mike Petonic for alerting me to this on the dev forums. Quite a big usability problem.

There is now an addition of a '3rd pane not-optional' choice for 3-pane actions.
That is, if you set the 'get argument count' part to return `-3` then the 3rd pane is optional. If it returns `3` then the 3rd pane is *not* optional (and will display by default).

I used -3 and 3 as I thought it was kind of intuitive `-3` meaning optional, `3` meaning not optional.
Open to other suggestions though

### DIFF
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSAppleScriptActions.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSAppleScriptActions.m
@@ -293,7 +293,15 @@
     return [self typeArrayForScript:path forHandler:@"DAEDgiob"];
 }
 
+-(NSArray *)defaultIndirectObjectForScript:(NSString *)path {
+	/* TODO: This code currently does nothing. 
+	   Future: Allow the user to define an identifier for the default iObject. Could be a proxy object
+	*/
+	return [self typeArrayForScript:path forHandler:@"DAEDgiov"];
+}
+
 - (NSArray *)validIndirectObjectsForAction:(NSString *)action directObject:(QSObject *)dObject {
+	
 	if ([action isEqualToString:kAppleScriptOpenTextAction]) {
         return [NSArray arrayWithObject:[QSObject textProxyObjectWithDefaultValue:@""]];
     } else if ([action isEqualToString:kAppleScriptOpenFilesAction]) {
@@ -309,11 +317,14 @@
 -(NSArray *)validIndirectObjectsForAppleScript:(NSString *)script directObject:(QSObject *)dObject {
     NSString *scriptPath = [self scriptPathForID:script];
     
-    id indirectTypes = [self validIndrectTypesForScript:scriptPath];
+    NSArray * indirectTypes = [self validIndrectTypesForScript:scriptPath];
     if (indirectTypes) {
-        NSMutableArray *indirectObjects = [NSMutableArray array];
+		if ([indirectTypes count] == 1 && [indirectTypes[0] isEqualToString:QSTextType]) {
+			return [NSArray arrayWithObject:[QSObject textProxyObjectWithDefaultValue:@""]];
+		}
+        NSMutableArray *indirectObjects = [NSMutableArray arrayWithObject:[NSNull null]];
         for (NSString *type in indirectTypes) {
-            [indirectObjects addObjectsFromArray:[QSLib arrayForType:type]];
+            [indirectObjects addObjectsFromArray:[QSLib scoredArrayForType:type]];
         }
         return [indirectObjects copy];
     }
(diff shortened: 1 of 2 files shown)


pjrobertson:
@skurfer any idea why the `testEquality` test is failing?

skurfer:
> There is now an addition of a '3rd pane not-optional' choice for 3-pane actions.

I don’t know this code at all, but there’s a comment that says “1 = dObj only. 2 = dObj + iObj. 3 = indirect Optional”. So didn’t 2 already mean that the third pane was required?

Personally, I think I’d make 1 and 2 the only valid values, and have a separate thing you could put in the AppleScript that returns a `BOOL` to indicate whether or not the third pane is required. But maybe we should get opinions from people writing AppleScript actions. 😃

skurfer:
> any idea why the `testEquality` test is failing?

Probably the same reason it always has. `combined1` are `combined2` are literally the same object when you run the test the first time (and that really shouldn’t be possible). If you run the test again, it will work correctly, but of course Travis only does it once.

This might be moot soon. If `XCTAssertEqualObjects` and `XCTAssertNotEqualObjects` are checking `hash` before they check `isEqual:`, the test will need to be adjusted.

Sesquipedalian:
I suspect that this isn't really going to help. The problem in #2071 is not a matter of controlling whether the third pane is optional or required (as @skurfer already noted, setting `get argument count` to either `2` or `3` already does that anyway). The problem is that the third pane is being populated with random junk for AppleScript actions. To fix #2071, the third pane should cleared whenever the user selects an AppleScript action in the second pane.

pjrobertson:
> The problem is that the third pane is being populated with random junk for AppleScript actions.

That's what this PR mainly fixed, I added the indirect option code (which Rob's pointed out isn't needed) as an addition. I'll just remove that 2nd commit then we'll be good... (done. Force pushed)

Sesquipedalian:
Oh, okay. I guess I misunderstood your initial post. :)

skurfer:
>  I'll just remove that 2nd commit then we'll be good...

Yeah, just looking at the code @Sesquipedalian, I think the stuff that ends up in the third pane should be greatly improved.

pjrobertson:
(what I was trying to fix in that un-needed commit was a bug where the 3rd plane closes when you enter text mode, but I've found this is universal, not just with AppleScripts).

What this does for your first script (indirect optional, when you tab to the 3rd pane) is to cause the pane to open, go into text mode, close, then reopen again

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSAppleScriptActions.m
```diff
@@ -309,11 +317,14 @@ - (NSArray *)validIndirectObjectsForAction:(NSString *)action directObject:(QSOb
...
     NSString *scriptPath = [self scriptPathForID:script];
     
-    id indirectTypes = [self validIndrectTypesForScript:scriptPath];
+    NSArray * indirectTypes = [self validIndrectTypesForScript:scriptPath];
     if (indirectTypes) {
-        NSMutableArray *indirectObjects = [NSMutableArray array];
+		if ([indirectTypes count] == 1 && [indirectTypes[0] isEqualToString:QSTextType]) {
+			return [NSArray arrayWithObject:[QSObject textProxyObjectWithDefaultValue:@""]];
+		}
+        NSMutableArray *indirectObjects = [NSMutableArray arrayWithObject:[NSNull null]];
```
skurfer: The way this is, won’t the third pane always have null as the first result? Don’t we want it to be null if not type was specified, otherwise, be an array of things matching that type? Or does the null get thrown out automatically when there are other choices?

pjrobertson: So `[NSNull null]` basically means that there is *no selection* in the 3rd pane. The results array is still populated (if there's a iObject type specified)

Basically, I didn't think we should make any assumptions about what object should be selection (e.g. should it be the highest ranking object of that type or what?) so I thought: best to leave it blank and let the user decide. He can search for it or press the ↓ arrow to scroll through the results.

My idea to choose a 'better' default object is a TODO, as can be seen here:
https://github.com/quicksilver/Quicksilver/pull/2232/files#diff-56ce0ea6b3cc1cdd1b7838ac51e344a3R296

skurfer: OK. Just knowing it was intentional is good enough for me.

petonic:
Already said thanks on the dev forum, but THANKS again.  I'll check it out on the next release of QS.

pjrobertson:
You're welcome. Thanks to you for bringing the issue to our attention :)

PR #2236, opened 2016-07-25
Sorting by date

### LINKED ISSUE #178: Add "Sort by Date" & "Sort by Size" menu items to results menu with keyboard shortcuts

### LINKED ISSUE #1494: Quicklook and clear current object doesn't reset to global search
If I navigate, select an object and quicklook (at) it, after pressing ⌃U or twice ⌫ I can only further search in that hierarchy where the object is located instead of globally.

https://groups.google.com/forum/?fromgroups=#!topic/blacktree-quicksilver/CzognL03dMk

skurfer:
Taking pieces from both @tiennou and my branches.

Still to fix/discuss:
- If you change the sort order, then browse into something else, it goes back to sorting by name, but the menu still shows whatever you last selected as “checked”.
- The default sort order when results are populated is undefined. Users should be able to pick a default.
- There is no keyboard shortcut for changing the sort order
- Are we doing size?
- Are we doing a “folders first” option?
- Should we add “name” to every descriptors array so it will act as a fallback when the other criteria are missing? It seems to be doing this anyway, but maybe we should be explicit.

What _should_ happen if you set the sort order to “date (newest first)” then browse? Should it go back to the user-selected default? Should it continue sorting by date until the interface is dismissed? Should it continue sorting by date forever until the user picks something else (changing the default, in other words)?

As for the default, I’d like to add both an application-wide default, and a default per-type. Or leave the default as “name” and let the per-type thing handle it. (Most people will only care about files/folders.)

(The use case for a per-type default: The only reason I worked on this in the first place was so when I right-arrow into an artist, the albums can be sorted on release date by default, while everything else in QS gets sorted by name.)

fixes #178, fixes #1494

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSBasicObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSBasicObject.h
@@ -16,6 +16,7 @@
 - (NSString *)identifier;
 - (NSString *)label;
 - (NSString *)name;
+- (NSDate *)date;
 - (BOOL)enabled;
 - (void)setEnabled:(BOOL)flag;
 - (id)primaryObject;
--- a/Quicksilver/Code-QuickStepCore/QSBasicObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSBasicObject.m
@@ -22,6 +22,7 @@
 - (NSString *)identifier {return nil;}
 - (NSString *)label {return nil;}
 - (NSString *)name {return @"Object";}
+- (NSDate *)date {return [NSDate distantPast];}
 - (BOOL)enabled {return ![[QSLibrarian sharedInstance] itemIsOmitted:self];}
 - (void)setEnabled:(BOOL)flag {[[QSLibrarian sharedInstance] setItem:self isOmitted:!flag];}
 - (id)primaryObject {return nil;}
--- a/Quicksilver/Code-QuickStepCore/QSObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject.h
@@ -17,6 +17,7 @@ extern NSSize QSMaxIconSize;
 - (NSString *)detailsOfObject:(QSObject *)object;
 - (NSString *)identifierForObject:(QSObject *)object;
 - (NSString *)kindOfObject:(QSObject *)object;
+- (NSDate *)dateForObject:(QSObject *)object;
 - (void)setQuickIconForObject:(QSObject *)object;
 - (BOOL)loadIconForObject:(QSObject *)object;
 - (BOOL)drawIconForObject:(QSObject *)object inRect:(NSRect)rect flipped:(BOOL)flipped;
@@ -38,6 +39,7 @@ extern NSSize QSMaxIconSize;
 // meta dictionary keys
 #define kQSObjectPrimaryName      @"QSObjectName"
 #define kQSObjectAlternateName    @"QSObjectLabel"
+#define kQSObjectDate             @"QSObjectDate"
 #define kQSObjectPrimaryType      @"QSObjectType"
 #define kQSObjectSource           @"QSObjectSource"
 #define kQSObjectIconName         @"QSObjectIconName"
@@ -83,6 +85,7 @@ typedef struct _QSObjectFlags {
 	NSString *name;
 	NSString *label;
 	NSString *identifier;
+	NSDate *datetime;
 	NSImage *icon;
 	NSString *primaryType;
 	id primaryObject;
@@ -139,6 +142,8 @@ typedef struct _QSObjectFlags {
 // This private method is required for QSProxyObject.m
 - (id)_safeObjectForType:(id)aKey;
 
+- (NSDate *)date;
+- (void)setDate:(NSDate *)newDatetime;
 @end
 
 @interface QSObject (Icon)
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -69,6 +69,7 @@ NSSize QSMaxIconSize;
 		meta = [NSMutableDictionary dictionaryWithCapacity:0];
 		name = nil;
 		label = nil;
+		datetime = nil;
 		icon = nil;
 		identifier = nil;
 		primaryType = nil;
@@ -246,6 +247,7 @@ NSSize QSMaxIconSize;
     copy.name = [name copy];
     copy.label = [label copy];
     copy.identifier = [identifier copy];
+	copy->datetime = [datetime copy];
     copy.icon = [icon copy];
     copy.primaryType = [primaryType copy];
     copy.primaryObject = [primaryObject copy];
@@ -812,6 +814,28 @@ NSSize QSMaxIconSize;
 	lastAccess = newlastAccess;
 }
 
+- (NSDate *)date
+{
+	if (!datetime) {
+		datetime = [self objectForMeta:kQSObjectDate];
+	}
+	id handler = nil;
+	if (!datetime && (handler = [self handlerForSelector:@selector(dateForObject:)])) {
+		datetime = [handler dateForObject:self];
+	}
+	if (datetime) {
+		return datetime;
+	}
+	return [NSDate distantPast];
+}
+
+- (void)setDate:(NSDate *)newDatetime
+{
+	if (datetime != newDatetime) {
+		datetime = newDatetime;
+		[self setObject:datetime forMeta:kQSObjectDate];
+	}
+}
 @end
 
 @implementation QSObject (Archiving)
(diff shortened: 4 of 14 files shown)


tiennou:
How embarrassing, I left a fixup! in there... I think you can squash that with "Make QSSearchMode enumeration consistent".

tiennou:
The per-type argument makes me wonder if we shouldn't ask the handler to compare for us... I'm thinking that interactions between files (what I wanted) and release dates (what you wanted) might make for a weird ordering when searching... Like if we ask the handler to compare files and iTunes tracks, it can default to the rank, since it doesn't really make sense — I don't think iTunes knows the filesystem path, right, *right* 😟 ?

tiennou:
Also, here's a list of the things I don't quite like :

- I'm not sold on `-[QSObject date]` (hence my point about handlers). Also, properties :wink:.
- I don't like the ascending/descending thing. I think a way to cleanup that menu would be to assign the tag (taken from the sort order enum) in IB to each menu item, `IBOutlet` the containing menu (which allows to remove the child outlets), add a separator + sort descending (=> `IBOutlet`), and iterate at runtime through the menu items, toggling state as you go and checking the "sort descending" to build the `NSSortDescriptor`. But that's some work 😉.

hmelman:
I don't completely follow how this will work in practice but a few (perhaps naive) comments that might help you think about things:

- There are different date fields I might want to sort by. Songs by year or last time played. Files by last modified or last accessed. 
- Type might not be the right granularity. I might want the Downloads folder by last modified and a documents folder by last accessed. It would be cool if you could find a default value, per folder, from the finder (not sure if that's possible)
- For the what should the default value be question, perhaps (per type) remember last setting as another option and a good default choice?

pjrobertson:
Seeing this message from Rob:

> (The use case for a per-type default: The only reason I worked on this in the first place was so when I right-arrow into an artist, the albums can be sorted on release date by default, while everything else in QS gets sorted by name.)

Made me understand what’s entailed in this ‘sort by date’ idea - I was originally just thinking of files.
When I first saw this it made me think of perhaps setting an NSDictionary in the QSObject’s `meta` dict called something like `sortableKeys`. It would be a key/value list of names (that would show in the interface) mapped to keys (which are accessed from the `meta` dict. I like Etienne’s idea of using the handler though.

My idea in more detailed:

```objective-c

// creating the object

iTunesSong *song = [iTunes getSongSomehow];

QSObject *obj = [QSObject makeiTunesSongObject:song];
[obj setObject:@{kQSLastTimePlayed: @"Last Time Played", kQSSongReleaseDate: @"Release Date"} forMeta:kQSSortingInfo];
[obj setObject:song.lastTimePlayed forMeta:kQSLastTimePlayed];
[obj setObject:song.releaseDate forMeta:kQSSongReleaseDate];

// in the results window code:
NSMutableDictionary *sortableOptions = [NSMutableDictionary dict];
for (QSObject* obj in resultsList) {
[sortableOptions addObjectsFromDictionary:[obj objectForMeta:kQSSortingInfo];
}
[QSResultWindow setSortableOptions:sortableOptions];
// in the background, this saves the `sortableOptions` dict in an iVar, and updates the UI

// When changing the results window sortable option:
-(void)sortableOptionChanged:(NSString *)optionKey {
[self.resultArray sortUsingComparator:^NSComparisonResult(QSObject *o1, QSObject *o2) {
     if ([o1 objectForMeta:optionKey] == [o2 objectForMeta:optionKey]) {
         // sort using the default sorting option. Name? Rank?
         return o1.rank > o2.rank;
     }
     return [o1 objectForMeta:optionKey] > [o2 objectForMeta:optionKey];
}];
}
```

tiennou:
> Should it continue sorting by date until the interface is dismissed?

I think that's best. Don't want to trample too much on users' choice. Also, drop it if there's a new search, we only rank on name.

About sorting : now that I see @pjrobertson proposes, I really think this belongs in the handler, akin to `-compareObject:withObject:options:`. Though I'm worried about the performance implications of all this (ranking is one of the "hot spots" in there)... The thing is that we don't want to expose too many knobs, so maybe a few filesystem dates (because that's what most of us want), and a few other (like sorting iTunes tracks).

> @hmelman : Type might not be the right granularity. I might want the Downloads folder by last modified and a documents folder by last accessed. It would be cool if you could find a default value, per folder, from the finder (not sure if that's possible)

Like scoped sort orders — it's a good idea, but I don't think it's worth tackling it head-on. You might be able to store the objects' children sort order in its meta, so maybe that would not be too hard ? I think Finder stores it's sort settings in the `.DS_Store` file, I think someone reverse-engineered those, but I'm not keen on doing that 😉.

skurfer:
> The per-type argument makes me wonder if we shouldn't ask the handler to compare for us...

The object source will be in charge of assigning (or potentially computing) the date. That should cover the various cases.

> Like if we ask the handler to compare files and iTunes tracks, it can default to the rank, since it doesn't really make sense — I don't think iTunes knows the filesystem path, right, right 😟 ?

Tracks are files, but again, the object source gets to decide what the date will be (file mod date vs. release date). My use case involves assigning dates to “browser” objects. Don't really care about tracks, but I’m pretty sure you can get both behaviors. That is, if you browse via the iTunes library, that plug-in will assign release date to the tracks. If you browse the same tracks via the filesystem, the file handler will assign the mod date to those objects. (Can’t test since date support hasn’t been added to the iTunes plug-in and probably won’t be until this is merged.)

> I'm not sold on `-[QSObject date]` (hence my point about handlers). Also, properties :wink:.

There are two reasons for that.

  1. I was trying to make it consistent with the existing “style” of the class, right or wrong. The date can get included in the inevitable clean-up with everything else.
  2. By declaring a complicated getter and setter, aren’t we tossing out all the benefits of properties anyway?

> I don't like the ascending/descending thing. I think a way to cleanup that menu would be to assign the tag (taken from the sort order enum) in IB to each menu item, `IBOutlet` the containing menu (which allows to remove the child outlets), add a separator + sort descending (=> `IBOutlet`), and iterate at runtime through the menu items, toggling state as you go and checking the "sort descending" to build the `NSSortDescriptor`. But that's some work 😉.

It certainly doesn’t scale if we add more options, but just for dates, I do like that it lets us use descriptive names (like “Newest First”).

> There are different date fields I might want to sort by. Songs by year or last time played. Files by last modified or last accessed.

At some point, you just need to use iTunes or Finder. 😉 Something like that could maybe be implemented as a bunch of actions. It might be a little cumbersome, but so would a menu with 50 options.

> Type might not be the right granularity. I might want the Downloads folder by last modified and a documents folder by last accessed. It would be cool if you could find a default value, per folder, from the finder (not sure if that's possible)

Certainly possible. We just associate a sorting order with an identifier instead of a type. But setting and clearing all those prefs and trying to remember where you overrode the default sounds awful (from a user’s point of view).

> I like Etienne’s idea of using the handler though.

What would be ideal is if the handler could just offer an array of sort descriptors (or multiple arrays) for the result controller to use. The problem is, it might instruct the table to sort itself on something that isn't fundamental to all QSObjects, like “last time played”.

pjrobertson:
> What would be ideal is if the handler could just offer an array of sort descriptors (or multiple arrays) for the result controller to use. The problem is, it might instruct the table to sort itself on something that isn't fundamental to all QSObjects, like “last time played”.

My solution to that was to fall back to the ‘default’ sorting option (e.g. name) if the object doesn’t have the ‘property’ that is being searched for. This is how my (very simple) proposed solution worked, and I think it could work inside the handler - if we were to go that way.

tiennou:
This will conflict with #2249, because of my cleanup. Maybe we should split that part out ? I'm still on the fence on how to handle that sorting problem cleanly...

skurfer:
Haven’t had time to investigate, but something in here prevents searches in the Trigger preferences’ Command Builder from working.

skurfer:
> and I think it could work inside the handler - if we were to go that way.

Question on that idea… If the handler is responsible for sorting, how would you handle a results list that contained objects from multiple sources?

tiennou:
Well, good question. The converse is also true BTW, and you're more likely (IIRC) to have objects that don't have sources than objects that don't have handlers (eg. browsing somewhere outside the catalog). Wouldn't using the "primary" handler be sufficient though ?

n8henrie:
@skurfer -- it's been ~7 years since this PR's last activity -- do you want this to stay open?

PR #2237, opened 2016-07-25
Change Preferences → Preferences to Preferences → General

skurfer:
I also updated the deployment target for the Preferences NIB. It looks like the entire file was rewritten, but really, that’s all I did.


pjrobertson:
What a simple but good idea! ;-)

A few things:

* I assume with all these changes/PRs (mainly NIBs) you're keeping track of things so you don't get merge conflicts?
* Can you add the 'General' string to `Localizable.strings`
* Should the 'Preferences -> General -> Application -> **General**' title now be changed? Since two Generals in there might be confusing. Perhaps it should just be split up into two sections: "Log-in", and "Dock & Menu Bar"

skurfer:
> I assume with all these changes/PRs (mainly NIBs) you're keeping track of things so you don't get merge conflicts?

I’ll avoid them where I can (so far, I think it’s just date sorting and proxy icons, and that one’s easy), but keeping the commits easy to understand is the priority.

> Can you add the 'General' string to `Localizable.strings`

Is that necessary for the English ones? It’ll just use the key.

> Should the 'Preferences -> General -> Application -> **General**' title now be changed?

I thought about that for a minute, but couldn’t come up with what to call it. I guess we could split it, but there are only 4 things there.

pjrobertson:
> Is that necessary for the English ones? It’ll just use the key.
It’s the English that provides the keys (in Transifex) for all the other languages. If it’s not there, then translators on Transifex will never know that string exists ;-)

> I thought about that for a minute, but couldn’t come up with what to call it. I guess we could split it, but there are only 4 things there.

Alright, not a biggy for now

skurfer:
Localized, rebased (because I like green checks), and pushed.

skurfer:
I wonder if the strings files confound the indent checker. So much for my green check. 😢

pjrobertson:
Or it’s the .travis.yml settings.
Try adding this line to it then pushing to this PR:

osx_image: xcode7.2

If all goes well, I’m happy for you to merge this PR. I’m off to bed now

pjrobertson:
Green is the way forward. Except in Chinese culture it's red. Now I'm confused as to what I want to see!
Glad it's working :)

PR #2239, opened 2016-07-25
correctly reset the third pane

skurfer:
The method I changed is used for actions too, but it was already calling `clearObjectValue`, so the behavior there should be unchanged.


pjrobertson:
Excuse my forgetfulness, but which method did you change? I just did a quick git blame of the line you changed and it hasn't been changed for... 9 years :P

skurfer:
I was referring to the method I changed on this branch (`updateControl:withArray:`) and the changes made today.

pjrobertson:
merged in master and tests pass. Looks good

PR #2240, opened 2016-07-26
Equality test

skurfer:
This gets it passing. It’s more like real-world usage now too, so that’s good.


skurfer:
OK, now that’s not my fault. I’m using tabs and all I did was change a comment, anyway.

Having said that, I tried combining this with the `dupes` branch and it fails there, so I need to look some more.

PR #2241, opened 2016-07-26
don't check index validity during a scan

skurfer:
It turns out the validity check was already happening on the correct queue, but it could get stuck waiting for that queue, as shown in #2016.


pjrobertson:
seems to make sense to me

Issue #2242, opened 2016-07-27
Crash in string value for multiple objects


pjrobertson:
It seems @skurfer's latest changes cause a crash in `stringValue`. To reproduce:

* Create a trigger for a 'combined' object for a nonexistent file path. E.g. in text mode type `~/nowhere`, pres ESC, then comma, then enter text mode again.
* You'll be left with a `dObject` of 2 text objects (to nonexistent files). The action is irrelevant
* Now add a keyboard shortcut to the trigger
* Try and activate the trigger (whilst debugging Xcode). You'll see there's an infinite loop - caused by calling `[object splitObjects]` in the `stringValue` method.

This was broken by 9e84daa3a6dd

pjrobertson:
Basically calling `[obj splitObject]` calls `[obj children]` which calls `[obj loadChildren]` which calls `[stringHandler loadChildrenForObject:obj]` which calls `[obj stringValue]` which calls... `[obj splitObject]`

I'm sure there are other ways of reproducing this, but this is how I just came across it

pjrobertson:
Although, 8f902d7d4956 could also be the bug. Checking for a *real* object using the stringValue as the identifier seems very hacky, although I know what you were trying to fix there

skurfer:
> Checking for a *real* object using the stringValue as the identifier seems very hacky, although I know what you were trying to fix there

In general, text objects don’t have identifiers, but they did for a brief period as you probably remember. I think that code was there so any triggers (or other stored QSCommands) created during that period would still work. So most of the time, it doesn’t do anything.

But I too question using `stringValue` now. Since we know exactly what we’re after, it seems like `objectForType:` would work just as well.

Anyway, I’ll get this fixed.

Issue #1858, opened 2014-05-30
Current Selection Proxy Triggers broken on 400C upgrade


hmelman:
I can confirm that #1843 is fixed for me in 400C. But I found that my existing triggers using the Current Selection Proxy objects were broken. I recreated some and they work fine. In the image below I fixed the first three, the others are broken (though the Finder Selection one is working. The broken ones end up searching for the text "QSGlobalSelectionProxy". In the second image you can see other Proxy objects that are broken (they all seem to have that icon).
![screen shot 2014-05-30 at 7 20 52 pm](https://cloud.githubusercontent.com/assets/1999190/3137377/e0234358-e851-11e3-857e-82120c2b4dae.png)
![screen shot 2014-05-30 at 7 24 35 pm](https://cloud.githubusercontent.com/assets/1999190/3137379/e41817e0-e851-11e3-8376-32942905a9c8.png)

PR #2243, opened 2016-07-27
avoid hang in stringValue

skurfer:
I could never induce `splitObjects` to call `children`, even following the steps in #2242, but the potential for an infinite loop was undeniable. Good catch, @pjrobertson!

This should prevent such a loop. The second commit may or may not help as well, but at worst, it makes things more efficient.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject_StringHandling.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject_StringHandling.m
@@ -291,6 +291,9 @@
 	if ([self count] > 1) {
 		QSObject *obj = [self resolvedObject];
 		// get the string value for each object in the collection
+		/* NOTE: getting the splti objects directly from cache may return `nil` in some instances
+				 This avoids an infinite loop as exhibited in #2242, but it may cause unexpected behaviour in certain cases
+		 */
 		stringValue = [[[obj objectForCache:kQSObjectComponents] arrayByPerformingSelector:@selector(stringValue)] componentsJoinedByString:@"\n"];
 	}
     if (!stringValue) {
--- a/Quicksilver/Code-QuickStepCore/QSObject_StringHandling.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject_StringHandling.m
@@ -289,8 +289,9 @@
         stringValue = [[NSString alloc] initWithData:stringValue encoding:NSUTF8StringEncoding];
     }
 	if ([self count] > 1) {
+		QSObject *obj = [self resolvedObject];
 		// get the string value for each object in the collection
-		stringValue = [[[self splitObjects] arrayByPerformingSelector:@selector(stringValue)] componentsJoinedByString:@"\n"];
+		stringValue = [[[obj objectForCache:kQSObjectComponents] arrayByPerformingSelector:@selector(stringValue)] componentsJoinedByString:@"\n"];
 	}
     if (!stringValue) {
         // Backwards compatibility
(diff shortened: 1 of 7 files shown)


pjrobertson:
I like the 2nd commit, even though it is probably still a bit hacky. Since what you said about us storing identifiers for text objects is no longer the case, can it not just be removed in its entirety?

It seems a bit 'black magic' to me!

skurfer:
> Since what you said about us storing identifiers for text objects is no longer the case, can it not just be removed in its entirety?

I found the reason for that code. See #1858. Sounds like the identifiers were getting stored as text, but they referred to more complicated objects. Makes more sense than text objects with identifiers surviving in someone’s indexes all this time. I’m still not sure if it’s safe to remove, though. I can look deeper tomorrow with fresh eyes.

Quicksilver/Code-QuickStepCore/QSObject_StringHandling.m
```diff
@@ -289,8 +289,9 @@ - (NSString *)stringValue {
         stringValue = [[NSString alloc] initWithData:stringValue encoding:NSUTF8StringEncoding];
     }
 	if ([self count] > 1) {
+		QSObject *obj = [self resolvedObject];
 		// get the string value for each object in the collection
-		stringValue = [[[self splitObjects] arrayByPerformingSelector:@selector(stringValue)] componentsJoinedByString:@"\n"];
+		stringValue = [[[obj objectForCache:kQSObjectComponents] arrayByPerformingSelector:@selector(stringValue)] componentsJoinedByString:@"\n"];
```
pjrobertson: so will there ever be a case that the `kQSObjectComponents` won't have been set yet when you get here?

skurfer: I couldn’t get it into that state, but it must be possible, otherwise `children` would have never been called when we were using `splitObjects` as described in your report of the issue.

If it’s not set, it moves to the next thing. No big deal, since this entire section didn’t exist a year ago.

pjrobertson: True, what you added is an improvement, and if it's not there it'd be "how it always was"

I'm going to merge but then add a comment RE this isn't ideal - incase it catches us out one day

pjrobertson:
OK I pushed a small change to add a comment to your change. I'm happy with this as is (even if it is slightly hacky).
I also added some multiline string tests. Perhaps controversially, the new tests I added fail... because a multiline string of paths (multiple file objects) return `2 Folders in "XXX"`.

Now the question is: this is the *`stringValue`*, but in reality, it's the `displayName` right? I'd say the string value (useful for pasting etc.) should be the paths split up using `\n`

pjrobertson:
Oh, wait... the reason it's failing is because

1.  There are no handlers set for file objects (because QSReg hasn't been populated when the tests run)
2. Even if the handler *were* set, because we're not calling `[obj splitObjects]` anymore (to avoid the crash this PR was fixing) we're getting `nil` from the cache.

Now we have the eternal dilemma of tests failing because they don't represent the real world... because the real world is way too complicated!

pjrobertson:
OK, let's try and get this merged so we're not sidetracked and discussing something else for ever! I've commented out the problematic unit tests, to be revised at a later date

skurfer:
> OK, let's try and get this merged so we're not sidetracked and discussing something else for ever! I've commented out the problematic unit tests, to be revised at a later date

Yes, thank you! It’s supposed to be a stupid one-line change. 😃

PR #2244, opened 2016-07-27
change show results “Never” to “Manually”

skurfer:
One by one, the NIBs fall.

I should have rolled this into the Preferences → General branch, but forgot about it.


pjrobertson:
Looks good, and got it in the strings file as well! :D

PR #2246, opened 2016-07-28
object source and handler protocols

skurfer:
More cleanup of the fundamental classes.

### DIFF
--- a/Quicksilver/Code-App/QSCatalogPrefPane.m
+++ b/Quicksilver/Code-App/QSCatalogPrefPane.m
@@ -352,10 +352,17 @@ static id _sharedInstance;
 			currentItemHasSettings = NO;
 			if ([source respondsToSelector:@selector(settingsView)])
 				currentItemHasSettings = nil != (settingsView = [source settingsView]);
-			if ([source respondsToSelector:@selector(setCurrentEntry:)])
+
+            // Make the object source edit the currently selected entry
+            // The various names are here because of back-compat
+            if ([source respondsToSelector:@selector(setSelectedEntry:)]) {
+                [source setSelectedEntry:currentItem];
+            } else if ([source respondsToSelector:@selector(setSelection:)]) {
+                [source setSelection:currentItem];
+            } else if ([source respondsToSelector:@selector(setCurrentEntry:)]) {
 				[source setCurrentEntry:[currentItem info]];
-			if ([source respondsToSelector:@selector(setSelection:)])
-				[source setSelection:currentItem];
+            }
+            
 			if ([source respondsToSelector:@selector(populateFields)])
 				[source populateFields];
 			if (settingsView) {
--- a/Quicksilver/Code-QuickStepCore/QSCore.h
+++ b/Quicksilver/Code-QuickStepCore/QSCore.h
@@ -46,6 +46,7 @@
 #import "QSObject_StringHandling.h"
 #import "QSObject_URLHandling.h"
 #import "QSObjectFormatter.h"
+#import "QSObjectHandler.h"
 #import "QSObjectRanker.h"
 #import "QSObjectSource.h"
 #import "QSParser.h"
--- a/Quicksilver/Code-QuickStepCore/QSExecutor.m
+++ b/Quicksilver/Code-QuickStepCore/QSExecutor.m
@@ -304,7 +304,7 @@ QSExecutor *QSExec = nil;
     }
 	NSArray *actions = nil;
 	if ([[dObject handler] respondsToSelector:@selector(actionsForDirectObject:indirectObject:)])
-		actions = (NSMutableArray *)[[dObject handler] actionsForDirectObject:dObject indirectObject:iObject];
+		actions = [[dObject handler] actionsForDirectObject:dObject indirectObject:iObject];
     
 	BOOL bypassValidation =
 		(bypass && [dObject isProxyObject] && [(QSProxyObject *)dObject bypassValidation]);
--- a/Quicksilver/Code-QuickStepCore/QSObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject.h
@@ -1,37 +1,11 @@
 #import <Foundation/Foundation.h>
 
 #import <QSCore/QSBasicObject.h>
+#import <QSCore/QSObjectHandler.h>
 
 @class QSObject, QSBasicObject;
 
 extern NSSize QSMaxIconSize;
-
-@interface NSObject (QSObjectHandlerInformalProtocol)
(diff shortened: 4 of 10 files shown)


Quicksilver/Code-App/QSCatalogPrefPane.m
```diff
@@ -352,10 +352,17 @@ - (void)updateEntrySelection {
...
 				currentItemHasSettings = nil != (settingsView = [source settingsView]);
-			if ([source respondsToSelector:@selector(setCurrentEntry:)])
+
+            // Make the object source edit the currently selected entry
+            // The various names are here because of back-compat
+            if ([source respondsToSelector:@selector(setSelectedEntry:)]) {
+                [source setSelectedEntry:currentItem];
+            } else if ([source respondsToSelector:@selector(setSelection:)]) {
+                [source setSelection:currentItem];
+            } else if ([source respondsToSelector:@selector(setCurrentEntry:)]) {
```
skurfer: Weren’t you the one that pushed us to use tabs instead of spaces, @tiennou? 😏

skurfer:
Is the idea to leave `currentEntry` but try to transition toward passing actual `QSCatalogEntry` objects to the sources instead of dictionaries? If so, 👍. I was trying to start a similar transition in #1455.

I’m trying to clean up the warnings in `QSUserDefinedProxySource` just to see how the API works now. `self.selectedEntry.sourceSettings` works for a few cases, but there are other places where the settings dictionary needs to get written back. I think we need something in `QSCatalogEntry` similar to `setName:`, but for the settings dictionary.

Another potential problem is that we seem to be passing `currentEntry` as the object for `QSCatalogEntryChangedNotification` (in the app and in 5 plug-ins), though looking at the observers, I’m not sure the object really matters. I can dig into that, but I can’t push any changes since this is on your fork.

skurfer:
OK, there’s at least one place where the object passed with `QSCatalogEntryChangedNotification` is a `QSCatalogEntry` and not a dictionary, and none of the observers are even looking at the object, so we can just change these to not use `currentEntry` any more.

If you want, I can put this on a branch I can write to, then

  * add a way to update settings in the catalog entry’s dictionary
  * add a way to update the date in the catalog entry’s dictionary
  * update `QSUserDefinedProxySource` to use the new API

skurfer:
> transition toward passing actual QSCatalogEntry objects to the sources instead of dictionaries?

Found the answer to that on [one of the other branches](/tiennou/Quicksilver/tree/catalog/pass-entries-around) you mentioned. 😄

Quicksilver/Code-App/QSCatalogPrefPane.m
tiennou: Well, you know how old stashes are, do you :wink:. I'll fix it.

tiennou:
I rebased it (I promise, `check_indent.rb` *needed* to be fixed 😉) and pushed it to quicksilver so you can play with it.

tiennou:
> > transition toward passing actual QSCatalogEntry objects to the sources instead of dictionaries?

Yeah, that's the part where I stopped, because I wasn't sure how to make the librarian automatically know that it's a new-style source, and thus it expects an entry instead of its dict...

Since I've been doing some rebasing, it seems there's a problem with how some things expect a date, and others a time interval (but I'm running with my QSLib cleanup branch merged, and I haven't checked that it's not because of that). YMMV, but the crash is pretty quick.

skurfer:
> pushed it to quicksilver so you can play with it

And then deleted it?

tiennou:
Yep, I'm like that 😉 . They're still in my repo. I'm not sure what made me change my mind again...

skurfer:
OK. I’ll keep an eye out for the pull request.

tiennou:
I'll reply to your comments here so they don't get lost. My ultimate goal with those `catalog/*` branches was to tackle #1455 (async object source), by cleaning up as much as possible the current code before that, like :

- how scanning is handled. Right now, I'm pretty sure we're unconditionally rescanning everything on startup, which makes the index thing useless. It's not *that* bad, but it's weird. That part is in the branch I haven't pushed because it's WIP.
- the fact that object sources are passed as their info dictionary is a real problem : I can't really build a nice API if my users just get dictionaries. I tried to tackle that, but haven't been able to make it transparent enough (like if I change my shiny new `QSObjectSource` protocol, every plugin breaks). Another way would be to add`-[QSCatalogEntry objectForKey:]`, and remove that later™. It feels hacky, but it's the most transparent (and you'll get warnings when building plugins anyway).

> there are other places where the settings dictionary needs to get written back.

That's in `QSCatalogEntry_Private.h`. I've kept it in case there was a use for that. If the API can be extended to support what you want, it's better, but I'm not sure what your use case right now...

> add a way to update the date in the catalog entry’s dictionary

Which date ? index, scan, or modification ? You should be able to use `-updateCurrentEntryModificationDate` for the latter (modulo that entry typing issue), the others are under the librarian's responsibility, so it's not useful to you (unless you have a use case I don't see).

skurfer:
> Right now, I'm pretty sure we're unconditionally rescanning everything on startup, which makes the index thing useless.

You’re right about that. I found the same thing when working on #1455.

https://github.com/quicksilver/Quicksilver/pull/1455/commits/51505f2121a5433fb590466bcfb777bbf203030d

> If the API can be extended to support what you want, it's better, but I'm not sure what your use case right now…

I was planning to look through plug-ins to get specific examples and make sure all the cases are covered, but in general, any object source that provides a UI for creating a new type of catalog entry (web search, remote hosts, spotlight, etc.) needs a way to save the settings back to the entry when it’s created/changed via the catalog prefs.

> Which date ? index, scan, or modification ?

I was just looking at the `QSUserDefinedProxySource` code as an example, which changes the modification date, but if we’re cleaning things up, the modification date should probably be managed internally and automatically.

tiennou:
> You’re right about that. I found the same thing when working on #1455.

See the [catalog/librarian-cleanup](https://github.com/tiennou/Quicksilver/compare/catalog/object-source-protocol...tiennou:catalog/librarian-cleanup?expand=1) branch, last 3 commits, which tries to bring order to the chaos. You might want to use your favorite Git GUI to make sense of that, because there are 3 branch on top of one another (this one > clean-obj-source > librarian-cleanup).

> any object source [...] needs a way to save the settings back to the entry when it’s created/changed via the catalog prefs.

As you note, I'm not sure if that's general, but the "core" ones uses bindings through that `-info` dict.

> the modification date should probably be managed internally and automatically.

Yeah, that's why there's a method to "touch" an entry, which takes care of the duplicated date-setting+notification.

Issue #1767, opened 2014-02-04
Add delayed commands to the task viewer


skurfer:
Things created with “Run After Delay…” and “Run at Time…” should appear in the Task Viewer. That will let you see that they exist, when they will run, and there should be a way to remove/cancel them from there.

I’ve already implemented 90% of this, but I’m waiting on #1611 to be finished.

PR #2247, opened 2016-07-28
add delayed commands to the task viewer

skurfer:
The oldest stash in my repo finally gets set free!

This does what I described in #1767.

The code isn’t as DRY as I’d like. I made some effort to factor out the task creation, but it’s not that easy. The task needs to know about the timer (for its cancel block) and the timer needs to know about the task (for its user info).

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSCommand.m
+++ b/Quicksilver/Code-QuickStepCore/QSCommand.m
@@ -185,21 +185,61 @@ NSTimeInterval QSTimeIntervalForString(NSString *intervalString) {
 }
 
 - (QSObject *)executeCommand:(QSObject *)dObject afterDelay:(QSObject *)iObject {
-	NSTimer *timer = [[NSTimer alloc] initWithFireDate:[NSDate dateWithTimeIntervalSinceNow:QSTimeIntervalForString([iObject stringValue])] interval:0 target:self selector:@selector(runCommand:) userInfo:dObject repeats:NO];
+	NSString *taskID = [NSString stringWithFormat:@"delay-%ld-%ld", dObject.hash, iObject.hash];
+	NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
+	[dateFormatter setDateStyle:NSDateFormatterShortStyle];
+	[dateFormatter setTimeStyle:NSDateFormatterShortStyle];
+	QSTask *delayedTask = [QSTask taskWithIdentifier:taskID];
+	[delayedTask setName:[dObject displayName]];
+	[delayedTask setIcon:[dObject icon]];
+	[delayedTask setShowProgress:NO];
+	NSDictionary *userInfo = @{
+		@"command": dObject,
+		@"task": delayedTask,
+	};
+	NSDate *fireDate = [NSDate dateWithTimeIntervalSinceNow:QSTimeIntervalForString([iObject stringValue])];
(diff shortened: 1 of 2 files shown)


skurfer:
Force-pushed to change indents to tabs. (Like I said, this was from a really old stash.)

Quicksilver/Code-QuickStepCore/QSCommand.m
```diff
@@ -185,21 +185,61 @@ - (void)selectTriggerInPrefPane:(QSTrigger *)trigger {
 }
 
 - (QSObject *)executeCommand:(QSObject *)dObject afterDelay:(QSObject *)iObject {
-	NSTimer *timer = [[NSTimer alloc] initWithFireDate:[NSDate dateWithTimeIntervalSinceNow:QSTimeIntervalForString([iObject stringValue])] interval:0 target:self selector:@selector(runCommand:) userInfo:dObject repeats:NO];
+	NSString *taskID = [NSString stringWithFormat:@"delay-%ld-%ld", dObject.hash, iObject.hash];
```
pjrobertson: what happens if a task with this ID already exists?

e.g. I set up a timer to run "Go to sleep" ⇥ Show Large Type for 2 hours, and then decide to set another one 1 hour later just incase I forget...

Would just a random taskID be ok?

pjrobertson: Should we be making a QSCommand and using that? I don't know, but if we used a QSCommand it might be easier to get a nicer ID, and also it would potentially be easier to save the delayed task to disk, which means that they might last through a relaunch of QS

pjrobertson: Or maybe use the actual time the task will be fired (`firedDate` below)

Quicksilver/Code-QuickStepCore/QSCommand.m
```diff
@@ -185,21 +185,61 @@ - (void)selectTriggerInPrefPane:(QSTrigger *)trigger {
...
+		@"command": dObject,
+		@"task": delayedTask,
+	};
+	NSDate *fireDate = [NSDate dateWithTimeIntervalSinceNow:QSTimeIntervalForString([iObject stringValue])];
+	NSString *taskStatus = [NSString stringWithFormat:@"Will run at %@", [dateFormatter stringFromDate:fireDate]];
+	[delayedTask setStatus:taskStatus];
+	NSTimer *timer = [[NSTimer alloc] initWithFireDate:fireDate interval:0 target:self selector:@selector(runCommand:) userInfo:userInfo repeats:NO];
+	[delayedTask setCancelBlock:^(void){
+		[timer invalidate];
+	}];
```
pjrobertson: All this code can at least go in another method to keep things DRY right? It's also in  `(QSObject *)executeCommand:(QSObject *)dObject atTime:(QSObject *)iObject `

e.g. `makeTaskWithDirectCommand: atTime:`

Quicksilver/Code-QuickStepCore/QSCommand.m
skurfer: Yeah, I guess ID collisions can still be manufactured. I’ll just make it a UUID. Doesn’t matter what it is, as long as it’s unique.

And the direct object *is* a command. Note that when it finally runs in `runCommand:`, it just gets cast, but it’s the same object. I wonder if we could just change the type of object in the action’s method signature.

skurfer:
DRY and UUID-enhanced!

PR #2250, opened 2016-07-29
Catalog protocols

tiennou:
A rejig of #2246, with some tweaks to check_indent because it didn't like documentation comments and staged changes.

### DIFF
--- a/Quicksilver/Code-App/QSCatalogPrefPane.m
+++ b/Quicksilver/Code-App/QSCatalogPrefPane.m
@@ -352,10 +352,17 @@ static id _sharedInstance;
 			currentItemHasSettings = NO;
 			if ([source respondsToSelector:@selector(settingsView)])
 				currentItemHasSettings = nil != (settingsView = [source settingsView]);
-			if ([source respondsToSelector:@selector(setCurrentEntry:)])
-				[source setCurrentEntry:[currentItem info]];
-			if ([source respondsToSelector:@selector(setSelection:)])
+
+			// Make the object source edit the currently selected entry
+			// The various names are here because of back-compat
+			if ([source respondsToSelector:@selector(setSelectedEntry:)]) {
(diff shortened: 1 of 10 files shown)


skurfer:
I guess if we can live with the deprecation warnings, this doesn’t actually break anything, right? I’ll test it out for a bit.

tiennou:
It shouldn't. I just noticed I missed adding `@optional` at the top of QSObjectHandler, so I'll fix that.

pjrobertson:
I like this! Rob - if you're happy with what you've tested so far, I say go ahead with the merge

Aside: Wow, after typing "I like this!" I got a kind of spooky feeling... it reminded me of the commit message @tiennou made... 2 and a half years ago: 2719535f478f1c3a025e765086545868160c960f every time you force pushed that branch (and it was probably about 20 times!) I got an email who's content was just "@pjrobertson likes those"

Scary!

skurfer:
Between this and the task clean-up, I’m staring at a wall of deprecation warnings we’ve imposed on ourselves. 😄 I know it’s worth it, but I don’t want them to get too out of control.

It looks like you’ve taken on a lot of them in #2257. After merging this, I’ll see about getting `currentEntry` calls out of the core app (unless you have a branch for that in progress).

tiennou:
Yeah, right now I'm running `tiennou/develop` which is a heck of a monster-merge-thing, and it seems to work* mostly. I'm looking at 93 warnings, and most of them come from NIBs (26) and SUPlainInstallerInternals (14).

*There are some catalog problems because of unfinished librarian-cleanup being in, but it fixes itself after a rescan.

The weird thing is that somehow Xcode doesn't show those warnings here yet (they magically appeared when working with a plugin). I'll see what not using my Workspace does btw...

tiennou:
FYI there's a `deprecated/breaking` on top of #2257 that kills most of the process-related ones.

PR #2252, opened 2016-07-30
update deployment target for the Task NIBs

skurfer:
I meant to do this before merging the task clean-up branch. Anyway, I really just changed the drop-down and nothing else.


skurfer:
Actually, I just put this on the other PR related to tasks. Never mind.

PR #2255, opened 2016-08-01
Cleanup object sources

tiennou:
This streamlines the `QSObjectSource`/`QSCatalogEntry` API, the ultimate goal being asynchronous loading of objects where that makes sense. I feel like this is pretty far away still, because I'd like to tackle the librarian first.

I expect this not to break anything, the things that need plugin changes will be submitted separately in PR #2256, but I need the infrastructure in there to progress (eg. `-selectedEntry`).


Quicksilver/Code-QuickStepCore/QSCatalogEntry.h
```diff
@@ -259,6 +262,13 @@ extern NSString *const QSCatalogEntryInvalidatedNotification;
 
 - (NSArray *)contentsScanIfNeeded:(BOOL)canScan;
 
+/**
+ * Mark the entry as refreshed/modified.
+ *
+ * @param rescan If YES, the entry will be rescanned immediately.
+ */
+- (void)refresh:(BOOL)rescan;
```
pjrobertson: I'm not sure about this terminology, I think using what's already in `QSCatalogEntry` would be best:

`- (void)scanForced:(BOOL)force;`

Quicksilver/Code-QuickStepCore/QSObjCMessageSource.m
```diff
@@ -23,9 +23,9 @@ - (id)init {
 	return self;
 }
 
-- (BOOL)entryCanBeIndexed:(NSDictionary *)theEntry {return NO;}
+- (BOOL)entryCanBeIndexed:(QSCatalogEntry *)theEntry {return NO;}
```
pjrobertson: I'm assuming you checked all currently known plugins for what going from `NSDictionary` -> `QSCatalogEntry` might do there, incase it breaks anything?

Quicksilver/Code-QuickStepCore/QSObjectSource.h
```diff
@@ -1,49 +1,108 @@
...
 
+/**
+ * The settings view for the source.
+ */
+@property (retain) IBOutlet NSView *settingsView;
 
+// Please use -selectedEntry instead of those
+// The rational being that between -currentEntry, -selection and direct Ivar
+// everything can go wrong.
+@property (retain) NSMutableDictionary *currentEntry QS_DEPRECATED;
```
pjrobertson: if `everything can go wrong` would it not be better to just delete these now and be done with it? Anywhere that breaks we can fix fairly quickly (if you've missed anywhere, that is)
It'd also give us errors on build which are more obvious that warnings

Quicksilver/Code-QuickStepCore/QSObjectSource.m
```diff
@@ -11,51 +11,37 @@
...
+	return self.selection.info;
 }
 
-- (QSCatalogEntry *)selection { return selection;  }
-- (void)setSelection:(QSCatalogEntry *)newSelection {
-	if(newSelection != selection){
-		selection = newSelection;
-	}
-}
+- (void)setCurrentEntry:(NSMutableDictionary *)currentEntry {}
```
pjrobertson: if this is empty should it not just be a `readonly` property?

Quicksilver/Code-QuickStepCore/QSObjectSource.m
```diff
@@ -11,51 +11,37 @@
...
 
-- (NSView *)settingsView { return settingsView;  }
-- (void)setSettingsView:(NSView *)newSettingsView {
-	settingsView = newSettingsView;
+- (void)updateCurrentEntryModificationDate {
+	self.selectedEntry.sourceSettings[kItemModificationDate] = @([NSDate timeIntervalSinceReferenceDate]);
 }
 
+- (QSCatalogEntry *)selection { return self.selectedEntry; }
+- (void)setSelection:(QSCatalogEntry *)selection { self.selectedEntry = selection; }
```
pjrobertson: do we need any `@synchronized` round these? Just wondering if there could be a race condition with getting and setting, if you were in the middle of changing the selection?

Quicksilver/Code-QuickStepCore/QSCatalogEntry.h
pjrobertson: Maybe I understand more what this is trying to do. Maybe:

`- (void)updateAndRescan:(BOOL)rescan;`?

tiennou: Yeah, my inspiration was `touch`, but that's not very good either... I think I prefer the one I have to -`updateAndRescan:rescan-again`, and `scanForced:` makes it looks like you're in for the ride — since it's the same on entries), which is not true here. But I'm open to suggestions.

Quicksilver/Code-QuickStepCore/QSObjCMessageSource.m
tiennou: I actually added `-objectForKey:` to entries as a hacky workaround. The expectation being that we'll have warnings galore, but runtime won't notice (unless it does something more dictionary-ish than I expected). Granted, I didn't really check for too long, and I actually found a missing `respondsToSelector:` I'll have to add.

Quicksilver/Code-QuickStepCore/QSObjectSource.m
tiennou: It's just a placeholder, as I don't think I can find an entry based on its `info` dictionary without going through hoops... I wrote that some time ago, and IIRC it's not the most used of those "duplicates".

Quicksilver/Code-QuickStepCore/QSObjectSource.m
tiennou: Right. `-selectedEntry` should be atomic. If it's not, that's an oversight, but I think the default is atomic now ?

Quicksilver/Code-QuickStepCore/QSObjectSource.h
skurfer: “Anywhere that breaks” includes a decent number of plug-ins. And to your comment below about it being read-only, it can’t be. Most of said plug-ins are updating catalog entries (in response to UI changes) by grabbing `[[self currentEntry] objectForKey:kItemSettings]` and then setting it back again.

skurfer:
I thought I had looked at this already. Apparently not. Just from initial testing, there’s at least one thing we need to fix:

screen shot 2016-08-12 at 9 47 27 am

Oh, “Group”! I love “Group”. 😃

I haven’t seen any functionality problems, though.

Quicksilver/Code-QuickStepCore/QSCatalogEntry.m
```diff
@@ -435,14 +441,32 @@ - (NSString *)name {
...
+				_name = [bundle safeLocalizedStringForKey:sourceString value:nil table:@"QSObjectSource.name"];
+				if (_name) break;
+			}
+		}
+
+
+		// Check to see if the name isn't hardcoded in our info
+		if (!_name) {
+			_name = self.info[kItemName];
+		}
```
skurfer: Moving this up above the previous bock (that checks `QSObjectSource.name`) fixes the all the incorrect catalog entry names. I feel like an explicitly defined name should have more weight anyway, and not be the last resort fallback.

This doesn’t fix names that are coming back `nil`. Still investigating that one.

skurfer:
For the catalog entries completely missing names, it’s because `[NSBundle bundleForClass:[self.source class]]` is returning `QSCore.framework` instead of the Core Support plug-in’s bundle. That’s as far as I’ve gotten for now.

tiennou:
Rebased, fixed another misused method, and hacked back custom names. I don't like really the current approach, but I'll try to make it cleaner when tackling customizable presets.

Quicksilver/Code-QuickStepCore/QSCatalogEntry.m
```diff
@@ -753,11 +792,24 @@ - (QSCatalogEntry *)uniqueCopy {
...
 
 
 // Backward-compatibility
 - (BOOL)deletable QS_DEPRECATED { return self.canBeDeleted; }
 
 @end
+
+@implementation QSCatalogEntry (OldStyleSourceSupport)
+
+- (id)objectForKey:(NSString *)key {
```
skurfer: This isn’t in the header. 😄

tiennou:
Because it's not the "user-defined" name at all. "User-defined"-ness is a hack that replaces whatever name the catalog entry had with something the user typed, by force-overriding it.
In other words, `name` is : 1) what's set from the entry data in the plugin's registration, 2) a localized name, 3) something the user typed. After serialization to Triggers.plist, there are no differences left between any of those 3 cases anymore, so on reload you have no idea if `name` was a "user-defined" name at all.

tiennou:
🙏 🌟

PR #2256, opened 2016-08-01
Breaking object sources

tiennou:
This is the breaking part of #2255. Settings view will go missing because the 1st commit, and plugins _will_ fail to load because of 0e22a18 (at least not the core ones. I think).

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObjectSource.h
+++ b/Quicksilver/Code-QuickStepCore/QSObjectSource.h
@@ -96,8 +96,18 @@
 
 /**
  * The settings view for the source.
+ *
+ * Calling this will cause the source's NIB file to be loaded from
+ * the source's NSBundle, with the source as its owner.
+ *
+ * It is expected that this will cause the settingsView outlet to be set.
  */
 @property (retain) IBOutlet NSView *settingsView;
+
+/** The NIB name used by -settingsView.
+ * Defaults to the name of the source's class, but can be overridden.
+ */
+@property (readonly, retain) NSString *settingsNibName;
 
 // Please use -selectedEntry instead of those
 // The rational being that between -currentEntry, -selection and direct Ivar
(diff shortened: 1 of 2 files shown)


Quicksilver/Code-QuickStepCore/QSObjectSource.h
```diff
@@ -112,7 +107,5 @@
 // Please use -selectedEntry instead of those
 // The rational being that between -currentEntry, -selection and direct Ivar
 // everything can go wrong.
```
pjrobertson: Wait, I mentioned these in your last PR I just looked at, saying "why not just remove them?". The answer seems to be dyld right? But here you've decided to remove them?!

Perhaps the comment needs removing as well?

pjrobertson:
OK I see what you're doing here, splitting up the breaking stuff and non-breaking stuff. Just to be sure: this PR would need to be merged *before* #2255 is merged right?

tiennou:
The reverse actually. This one's on top of #2255, after we went through object sources.

Quicksilver/Code-QuickStepCore/QSObjectSource.h
tiennou: Oops ! dyld is effectively quite good at catching disappearing ivars...

skurfer:
So shouldn't the NIBs be updated too? For example, the `QSFileSystemObjectSource.xib` is still using `selection.info.settings`. That should change to `selectedEntry.sourceSettings`, right?

josephholsten:
with #2255 merged, does this need to exist any longer?

skurfer:
I believe that was part 1, the backward-compatible part. This is part 2, which could break stuff.

n8henrie:
It's been several years, I'm going to close this as stale, happy to reopen if work on this resumes! Thank you @tiennou for your efforts here!

PR #2257, opened 2016-08-01
Deprecation cleanup

tiennou:
Again 😉.

Apart from a4f069, which is a slight change so we don't need to be explicit about indeterminate progress on `QSTask`, tackled in 85e3159, the rest is pure warning fixes.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject_URLHandling.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject_URLHandling.m
@@ -188,10 +188,11 @@
 
 
 - (BOOL)loadChildrenForObject:(QSObject *)object {
-	
-    
-	[QSTasks updateTask:@"DownloadPage" status:@"Downloading Page" progress:0];
-    
+	QSTask *task = [QSTask taskWithIdentifier:@"DownloadPage"];
+	task.status = NSLocalizedString(@"Downloading Page", @"");
+	task.showProgress = NO;
+	[task start];
+
     if (![object objectForMeta:QSURLTypeParsersTableKey]) {
         // it's possible `objectHasChildren:` was never called to populate this
         [self objectHasChildren:object];
@@ -200,8 +201,8 @@
     id <QSParser> parser = [QSReg instanceForKey:[object objectForMeta:QSURLTypeParsersTableKey] inTable:@"QSURLTypeParsers"];
     
 	NSArray *children = [parser objectsFromURL:[NSURL URLWithString:[object objectForType:QSURLType]] withSettings:nil];
-    
-	[QSTasks removeTask:@"DownloadPage"];
+
+	[task stop];
     
 	if (children) {
 		[object setChildren:children];
--- a/Quicksilver/Code-QuickStepCore/QSPlugInManager.m
+++ b/Quicksilver/Code-QuickStepCore/QSPlugInManager.m
@@ -23,6 +23,11 @@
 #define pPlugInInfo QSApplicationSupportSubPath(@"PlugIns.plist", NO)
 #define MAX_CONCURRENT_DOWNLOADS 2
 
+@interface QSPlugInManager ()
+@property (retain) QSTask *downloadTask;
+@property (retain) QSTask *installTask;
+@end
+
 @implementation QSPlugInManager
 + (id)sharedInstance {
 	static id _sharedInstance;
@@ -200,7 +205,11 @@
 																	 delegate:self];
 
 		if (theConnection) {
-			[QSTasks updateTask:@"Retrieving Plugins..." status:@"Updating Plugin Info" progress:0.0];
+			self.downloadTask = [QSTask taskWithIdentifier:@"PluginUpdateInfo"];
+			self.downloadTask.status = NSLocalizedString(@"Updating Plugin Info", @"");
+			self.downloadTask.cancelBlock = ^{
+				[theConnection cancel];
+			};
 		} else {
 			NSLog(@"Problem downloading plugin data. Perhaps an invalid URL");
             receivedData = nil;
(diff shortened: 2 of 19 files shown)


Quicksilver/Code-QuickStepCore/QSObject_URLHandling.m
```diff
@@ -188,10 +188,11 @@ - (BOOL)objectHasChildren:(QSObject *)object {
 
 
 - (BOOL)loadChildrenForObject:(QSObject *)object {
-	
-    
-	[QSTasks updateTask:@"DownloadPage" status:@"Downloading Page" progress:0];
-    
+	QSTask *task = [QSTask taskWithIdentifier:@"DownloadPage"];
+	task.status = NSLocalizedString(@"Downloading Page", @"");
```
pjrobertson: I like this, but could it be more descriptive?

Like `@"Download Page %@", [object stringValue]` or something?
Also, the comments are useful for localisers so I think they should be kept in.

Quicksilver/Code-QuickStepCore/QSPlugInManager.m
```diff
@@ -200,7 +205,11 @@ - (void)downloadWebPlugInInfoFromDate:(NSDate *)date forUpdateVersion:(NSString
 																	 delegate:self];
 
 		if (theConnection) {
-			[QSTasks updateTask:@"Retrieving Plugins..." status:@"Updating Plugin Info" progress:0.0];
+			self.downloadTask = [QSTask taskWithIdentifier:@"PluginUpdateInfo"];
+			self.downloadTask.status = NSLocalizedString(@"Updating Plugin Info", @"");
+			self.downloadTask.cancelBlock = ^{
+				[theConnection cancel];
```
pjrobertson: my use of blocks and retain cycles is still a bit rusty, but does this want to be a `weak` reference? Or is the fact that you're explicitly setting the task to `nil` in all cases enough to stop `theConnection` from getting stuck in a retain cycle?

Quicksilver/Code-QuickStepFoundation/NSWorkspace_BLTRExtensions.m
```diff
@@ -16,46 +16,36 @@
...
-			aeDesc = [script executeAndReturnError:nil];
-			return [aeDesc stringValue];
-		} else {
-			NSBeep();
-		}
+	if (!path) return nil;
+
+	NSArray *finders = [NSRunningApplication runningApplicationsWithBundleIdentifier:@"com.apple.Finder"];
+	if ([finders count] == 0) {
+		NSBeep();
```
pjrobertson: The dreaded NSBeep()! ;-) 
Anyway this could magically turn into an NSLog instead? I know the chance of this happening is basically `nil` but...

Quicksilver/Code-QuickStepFoundation/NSWorkspace_BLTRExtensions.m
```diff
@@ -16,46 +16,36 @@
...
-
-		CFURLRef fileURL = CFURLCreateWithFileSystemPath(NULL, (CFStringRef) path, kCFURLPOSIXPathStyle, NO);
-		hfsPath = (NSString *)CFBridgingRelease(CFURLCopyFileSystemPath(fileURL, kCFURLHFSPathStyle));
-		CFRelease(fileURL);
-
-		scriptText = [NSString stringWithFormat:@"tell application \"Finder\" to set comment of item \"%@\" to \"%@\"", hfsPath, comment];
-		script = [[NSAppleScript alloc] initWithSource:scriptText];
-		result = ([[script executeAndReturnError:nil] stringValue] != nil);
-	} else {
+	NSArray *finders = [NSRunningApplication runningApplicationsWithBundleIdentifier:@"com.apple.Finder"];
```
pjrobertson: looking at this now it might be nice to create an `isFinderRunning` method here, and use that. I was confused for a minute about what you were doing with this `finders` array before I realised it was just a check to see if Finder is running

Quicksilver/Code-QuickStepCore/QSPlugInManager.m
tiennou: Well, it depends... You could mark `theConnection` as `weak`, which would break the retain cycle automatically, but you can also do it manually by making sure that url gets set to `nil` sometimes. Arguably, I find this cleaner — you make `theConnection`'s lifetime dependent on `downloadTask`.

Now that I'm looking at it, what's causing `theConnection` to stay alive anyway ? At the end of that function, there's no reference to it anywhere, right ?

Quicksilver/Code-QuickStepFoundation/NSWorkspace_BLTRExtensions.m
tiennou: Your call. My personal take is that this `NSBeep()` is effectively at the wrong level, but fixing that required tweaking the signature, which I didn't want to do.

Also, while I'm tackling this, how does that function relates to our non-Finder users (eg. PathFinder) ? I just realized this might have to go though the FSBrowserMediator thingy...

Quicksilver/Code-QuickStepFoundation/NSWorkspace_BLTRExtensions.m
tiennou: I agree it might be cleaner. But I'll hold off until the point about the "standard" workspace manager is tackled above.

Quicksilver/Code-QuickStepFoundation/NSWorkspace_BLTRExtensions.m
skurfer: It’s just asking for information, so as long as Finder is running, it shouldn’t matter if you actually use it. I use Path Finder, but keep Finder running. I think there’s an option to make it quit, but there are edge cases that make it not worth it. No one has complained about it ever, so I wouldn’t worry about it.

Quicksilver/Code-QuickStepCore/QSObject_URLHandling.m
skurfer: There’s a name and a status, sort of like “details” for the normal object views. The string above should probably be the name, leaving the status (which can fit more text) to display the actual URL.

Also, wouldn’t it be possible in rare circumstances to get more than one of these going simultaneously? Should the URL be used as part of the task’s identifier to prevent collision?

Quicksilver/Code-QuickStepFoundation/NSWorkspace_BLTRExtensions.m
skurfer: > Anyway this could magically turn into an NSLog instead?

Having looked closer, I agree. No need to beep here. The user would have no idea what caused it or how to address it.

PR #2268, opened 2016-09-02
save mnemonics again

skurfer:
Revert "Clear last search when closing interface"

This reverts commit aeaa261352e3caa42838e3a1f4caa2a9ef90953c.

It seems that this was wiping out the search string faster than we could save it, so QS stopped “learning”. 😢 

It *may* have been related to this comment https://github.com/quicksilver/Quicksilver/pull/2249#issuecomment-236476810

But I can’t reproduce that problem after reverting the commit. Can someone confirm?


tiennou:
Yep, that fixes it.

PR #2269, opened 2016-09-02
Only hide the interface after everything has been set up for execution

tiennou:
Fixes #2268 while keeping the "clear on close" behavior of aeaa261352e3caa42838e3a1f4caa2a9ef90953c


skurfer:
Seems to work. Wish I could figure out why `saveMnemonic` gets hit twice when you type ↩. It _claims_ to be coming from `insertNewline:` both times, but if I move the breakpoint there, it only gets hit once.

Another issue for another day.

PR #2274, opened 2016-09-17
New App icon. Looks great in macOS Sierra.

vadimrob:
I've changed the defaultAction.icns so it looks great in macOS Sierra UI.


pjrobertson:
Does this have high a high resolution set built in as well? I'm assuming from the filesize (70kb) it doesn't

vadimrob:
I've made a new one, it's almost the same size but it's sharp and looks good on retina displays

skurfer:
I don’t like the current icon, but if we’re replacing it, we should quit using one or more gears. That never made sense to me. It should be some kind of arrow, or a person running, or something having to do with “action”. Gears have come to be associated with settings.

pjrobertson:
> Gears have come to be associated with settings.

Good point, I like the idea of an arrow or a running man!

vadimrob:
OK, we'll approach this by a different angle. Let's update the default icon. Then we will go from there. I'm proposing a design:
quicksilver 2x

skurfer:
Now that looks more promising. I’ll have to look at it in context, though.

josephholsten:
@skurfer any thoughts about merging this? does it require more work to get the icon referenced correctly?

pjrobertson:
I think this looks like a good update, but the icns file format/sizes seems incorrect.

See this [Apple doc on the image sizes](https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/) and then [this blog post](https://blog.macsales.com/28492-create-your-own-custom-icons-in-10-7-5-or-later) on how to create the ICNS file

Screenshot 2019-04-24 at 08 12 29

fturcheti:
I like the idea of inverting the icon colors — make purple what is current gray and vice versa. This could result in a more vibrant icon. But the implementation here is just too bland, it lacks contrast and depth. So instead of a potentially richer icon, it goes the opposite way. 

To illustrate what I'm talking about, I did a sketch using ideas from the proposed icon but targeting something bolder, lively:

Artboard 2@2x

About the suggested action icon, although sharper and retina-friendly, it looks too complex. And I agree with @skurfer, an arrow would be much more appropriate icon for "run" than a gear.

pjrobertson:
@vadimrob  and @fturcheti  - it's been a while, but both your icon suggestions are great. Let's give Quicksilver a facelift!

If either of you still has your files, please can you provide the icon in the following format then we'll merge this in!

Screenshot 2019-04-24 at 08 12 29

Let's bring QS into the 202s :D

vadimrob:
Hello! I'll check and see what I've got but I think if you really want to update it now we should modify to a square type of icon design that is better integrated to the current macOS design language.

vadimrob:
@pjrobertson
qs
Screenshot 2022-02-03 at 12 17 54

n8henrie:
Unclear if this will be any help, but I whipped it up yesterday for something entirely unrelated, and it seemed relevant.

I don't really know what I'm *supposed* to do with the `@2x` thing, but this seems to make an `.app` happy.

```
#!/bin/bash
# icons.sh
# Quick and dirty script to make an icon package based on a single higher
# resolution image.
# Usage: `./icons.sh MyLargeIcon.jpg`

set -Eeuf -o pipefail
set -x

main() {
  img=$1

  sips -s format png --out tmp.png "${img}"
  mkdir icon.iconset

  sz=16
  while [[ "${sz}" -le 512 ]]; do
    for multiplier in "" "@2x"; do
      sips -z "${sz}" "${sz}" tmp.png --out "icon.iconset/icon_${sz}x${sz}${multiplier}.png"
    done
    ((sz *= 2))
  done
  iconutil -c icns icon.iconset
  rm -r icon.iconset tmp.png
}
main "$@"
```

pjrobertson:
@vadimrob - looks great! I will comment on the other pull request.

skurfer:
We already have one of the new style square icons on `master`, but this one’s fine if you like it better.

PR #2293, opened 2016-11-30
Various lost in limbo fixes

tiennou:


skurfer:
The conflict panel is still getting called on the main thread in `-[FSActions renameFile:withName:]`. 😃 

Why not force it to the main thread in `QSFileConflictPanel` itself so the one calling it doesn’t have to remember?

skurfer:
Added some commits so we can get a release out @tiennou. Let me know what you think.

As for the panel getting detached from the interface, it seems to happen only if I build on 10.10 and run on 10.12. Running QS on the same machine I build from always seems to work.

On that subject… Is there any reason to keep building on 10.10? We’re going to disable ATS. Were there any other reasons? (Also, since we require 10.10 now, I’m going to update the deployment target and fix the OS version code in another pull request.)

skurfer:
One more commit, but this isn’t ready. Even if you choose to replace the destination, it fails because it already exists.

skurfer:
The problem with replacing files seems to be from very old code.

https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSActionProvider_EmbeddedProviders.m#L669

Doesn’t that mean the files will *never* be replaced, even when you explicitly approve it? What am I missing?

Anyway, no need to hold this up for that issue.

PR #2295, opened 2016-12-08
change deployment to 10.10

skurfer:
Another attempt…

Same as last time, but I was able to squash the comment updates into the other commit.

NIBs were updated and saved using Xcoe 7.2.1 this time.


pjrobertson:
all looks good, builds fine. I wrote one small comment, once that's fixed I say go ahead. (I haven't tested/built this, but I trust Travis)

skurfer:
I don’t see a comment anywhere @pjrobertson.

One thing I forgot to mention… I didn’t update `QSHelpersPrefPane.xib` because I couldn’t. Opening it and saving it (without changes) crashes Xcode (7 and 8). In Xcode 8, just selecting it will eventually lead to a crash. Any ideas?

Quicksilver/Code-QuickStepFoundation/NSApplication_BLTRExtensions.m
```diff
@@ -142,41 +142,28 @@ - (void)updateLaunchStatusInfo {
...
-    Gestalt (gestaltSystemVersion, &version);
-    return version;
++ (NSOperatingSystemVersion)macOSXSystemVersion {
+	return [[NSProcessInfo processInfo] operatingSystemVersion];
 }
-#pragma clang diagnostic pop
-#endif
 
 + (NSString *)macOSXReleaseVersion {
     return [NSString stringWithFormat:@"%i.%i", (int)[self macOSXMajorVersion], (int)[self macOSXMinorVersion]];
```
pjrobertson: hmm... my comment disappeared. Although I clicked 'start a review' instead of 'add single comment' (some new GH feature?)

I was just saying that it's probably better to not cast to `int`, although it probably won't make much of a difference. I'd use `%ld` then cast to a long. Although this has other suggestions: http://stackoverflow.com/questions/4405006/nslog-printf-specifier-for-nsinteger

Tiny/nit-picking comment, sorry! :P

pjrobertson: hmm... my comment disappeared. Although I clicked 'start a review' instead of 'add single comment' (some new GH feature?)

I was just saying that it's probably better to not cast to `int`, although it probably won't make much of a difference. I'd use `%ld` then cast to a long. Although this has other suggestions: http://stackoverflow.com/questions/4405006/nslog-printf-specifier-for-nsinteger

Tiny/nit-picking comment, sorry! :P

Quicksilver/Code-QuickStepFoundation/NSApplication_BLTRExtensions.m
```diff
@@ -142,41 +142,28 @@ - (void)updateLaunchStatusInfo {
...
-    SInt32 version;
-    Gestalt (gestaltSystemVersion, &version);
-    return version;
++ (NSOperatingSystemVersion)macOSXSystemVersion {
+	return [[NSProcessInfo processInfo] operatingSystemVersion];
 }
-#pragma clang diagnostic pop
-#endif
 
 + (NSString *)macOSXReleaseVersion {
```
pjrobertson: hmm... my comment disappeared. Although I clicked 'start a review' instead of 'add single comment' (some new GH feature?)

I was just saying that it's probably better to not cast to `int`, although it probably won't make much of a difference. I'd use `%ld` then cast to a long. Although this has other suggestions: http://stackoverflow.com/questions/4405006/nslog-printf-specifier-for-nsinteger

Tiny/nit-picking comment, sorry! :P

skurfer:
The suggestions from SO are interesting, but since we’re 64-bit only, I’ve gone with `long` as you suggested.

pjrobertson:
Once it's been compiled and built can you actually open the QSHelpersPrefPane panel? So it's just Xcode's problem or what?
So I just managed to open it by navigating to the file in Finder then double clicking.  Do you want to merge this then I'll take a stab at updating that Nib?

I'm on 10.11 / Xcode 8.1

skurfer:
That’s “Handlers”, right? I can open it in the prefs. I can also open it in Xcode most of the time, but saving it is a disaster. It crashes Xcode 8, which tries to open it every time you relaunch and immediately crashes again. You have to delete your workspace files to make it forget you had that file selected.

Maybe we need to install a much older Xcode and try to update it incrementally?

pjrobertson:
OK, go ahead and merge. I have Xcode 6 on a Snow Leopard VM I think, I can try and go through the motions there.

PR #2303, opened 2017-01-03
avoid calling QSUTIForAnyTypeString() twice

### LINKED ISSUE #2299: Quicksilver 1.5.1(build 401C) can't startup on macOS Sierra
Thanks for all your hard work for the latest release of QS. However, after upgrading to version 1.5.1, it can't startup. 

screen shot 2016-12-28 at 8 00 13 pm

I have to turn back to version 1.4.2 :( 
Any further information needed?

skurfer:
This seems to fix #2299. At the very least, it’s slightly more efficient.


pjrobertson:
Seems reasonable

skurfer:
I’m glad you think so since it was released. 😄

Issue #2301, opened 2017-01-01
Mnemonics not being remembered


craigfrancis:
I'm doing a clean install of macOS Sierra (10.12.2), and QS 1.5.1 (401C); but this also seems to effect my old computer on El Capitan (OSX 10.11.6) that has just updated QS to 1.5.1 (401C).

Normally when using QS, if you start typing, you can press the down arrow and select something else that matches - doing so increases it's score, so when it's score is high enough, it can be the default for what you just typed.

At the moment the score does not change - for example, when I type "cal", I currently have the Calendar app in second place with a score of 92, and this does not not change, no matter how many times I select it.

I'm also unable to right hand click and use the "Make Default" menu item (it's there, but it's disabled).

On my old computer, where I've been using QuickSilver for years, the scores are no longer changing either - but that's kind of ok, because the defaults are pretty much correct.

![screen-shot-2017-01-01-at-13 49 20](https://cloud.githubusercontent.com/assets/1296695/21581506/be7d52b0-d02b-11e6-85f2-fe7d3ef05eaa.jpg)

craigfrancis:
Is this to do with Mnemonics.plist?

After quitting QuickSilver, deleting this file, starting again, typing "cal" and selecting "Calendar", I'm getting:

	<?xml version="1.0" encoding="UTF-8"?>
	<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
	<plist version="1.0">
	<dict>
		<key>abbreviation</key>
		<dict>
			<key>QSActionMnemonic:NSFilenamesPboardType</key>
			<array>
				<string>AppLaunchAction</string>
				<string>AppReopenAction</string>
			</array>
			<key>cal</key>
			<array>
				<string>/Library/Scripts/Script Editor Scripts/Image Manipulation/Scale.scpt</string>
			</array>
		</dict>
		<key>implied</key>
		<dict>
			<key>/Applications/Internet Explorer.app</key>
			<dict>
				<key>ie</key>
				<integer>2</integer>
			</dict>
			<key>/Applications/Mail.app</key>
			<dict>
				<key>m</key>
				<integer>2</integer>
				<key>mail</key>
				<integer>2</integer>
			</dict>
			<key>/Applications/Safari.app</key>
			<dict>
				<key>s</key>
				<integer>2</integer>
				<key>safari</key>
				<integer>2</integer>
[...]

PR #2304, opened 2017-01-03
Another go at ensuring search string is cleared when closing interface

### LINKED ISSUE #2300: Image Manipulation PlugIn BUG
Dear Support, please note that after new update of QuickSilver I face a specific bug. Whenevery I make a screenshot via Image Manipulation PlugIn, the Quicksilver window remains on the screenshot (while in previous versions this Quicksilver window was hidden during screenshot snapping). See example attached
screen region 2016-12-31 at 18 57 11

### LINKED ISSUE #2302: 'Type Text' action not working in 1.5.1
The 'Type Text' action appears not to be working anymore in version 1.5.1 (build 401C). OS X 10.11.6 on 2012 rMBP.

Rather than typing the text, the system error bell chimes and Quicksilver dismisses as normal. The 'Paste' and 'Paste as Plain Text' actions work as expected. Quicksilver still has permission under assistive applications. (I assume Type Text works by accessing the virtual keyboard?)

Nothing is reported in the console that seems obviously relevant. However, both invoking and dismissing Quicksilver throws multiple WindowServer errors e.g.:
```
1/1/17 2:01:54.439 PM WindowServer[197]: _CGXRemoveWindowFromWindowMovementGroup: window 0x9e is not attached to window 0x2e7
```
Hex code at the end varies. I'm using the Bezel Interface.

Let me know if I can adjust anything to help you test!

pjrobertson:
The problem is that clearing the search string too early means mnemonics are not saved, so if we move the mnemonic saving code earlier this fixes it.
The reason we want to clear the search string on interface close is a UI thing. See this comment: https://github.com/quicksilver/Quicksilver/pull/2249#issuecomment-236476810
Fixes #2302
Fixes #2300
Related to #2268 #2269 

------

OK guys take a look at this. Basically, I took a look at what @tiennou had done in 2b2cc2dfe4b7a3d4270581527d9ddd130bd8ce6f but realised that that `if (!cont) {}` code is needed but understood why Etienne was doing this - because closing the interface meant the search string was cleared, which meant mnemonics could not be saved (since that was done later on).

Instead, what I did was to make saving the mnemonics happen earlier. Now I'm not sure of the consequences of saving mnemonics *before* an action has been run. Does anybody know?

I've just done some testing, and in fact saving the mnemonics *too late* caused other bugs. 
For example take the an action that accepts an iObject, e.g. `[some folder] ⇥ [make new...] ⇥ [my folder]`

This action returns an object (the newly created folder) and displays it in the interface. The new object is created and returned when the action is executed (see `-[QSInterfaceController executeCommandThreaded] -> -[QSCommand execute] ->` which leads to [here](https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/Code-QuickStepCore/QSCommand.m#L474) where the interface is cleared), which is *before* the mnemonics were originally saved. This way the mnemonic would never be saved for the original `[some folder]` object, since it's already been cleared for the newly created folder.

Make sense?

In short - we need to be sure that saving the mnemonic before executing the command doesn't cause more harm than good

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/QSInterfaceController.m
+++ b/Quicksilver/Code-QuickStepInterface/QSInterfaceController.m
@@ -618,9 +618,7 @@
 		NSBeep();
 		return;
 	}
-    
-    // add the object being executed to the history
-    [dSelector updateHistory];
+
     QSAction *action = [aSelector objectValue];
 	NSInteger argumentCount = [action argumentCount];
 	if (argumentCount == 2) {
@@ -634,9 +632,24 @@
 		}
 		[QSExec noteIndirect:[iSelector objectValue] forAction:action];
 	}
+	
+	// add the object being executed to the history
+	[dSelector updateHistory];
+	// make sure to save mnemonics before interface is closed. Closing the interface clears the search string so they must be saved before this
+	[QSHist addCommand:[self currentCommand]];
+	[dSelector saveMnemonic];
+	[aSelector saveMnemonic];
+	if (argumentCount == 2) {
+		[iSelector saveMnemonic];
+	}
+	
 	if (encapsulate) {
 		[self encapsulateCommand];
 		return;
+	}
+	if (!cont) {
+		// this ensures the interface is hidden before an action is run e.g. the 'capture screen region' and 'type text' actions needs this
+		[self hideMainWindowFromExecution:self]; // *** this should only hide if no result comes in like 2 seconds
 	}
 	if ([[NSUserDefaults standardUserDefaults] boolForKey:kExecuteInThread] && [action canThread]) {
         QSGCDAsync(^{
@@ -647,12 +660,6 @@
         QSGCDMainSync(^{
             [self executeCommandThreaded];
         });
-    }
-	[QSHist addCommand:[self currentCommand]];
-	[dSelector saveMnemonic];
- 	[aSelector saveMnemonic];
-	if (argumentCount == 2) {
-        [iSelector saveMnemonic];
     }
 	if (cont) {
         [[self window] makeFirstResponder:aSelector];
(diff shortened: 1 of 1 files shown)


pjrobertson:
P.S. do I win an award for the most long-winded commit message?! :D

Quicksilver/Code-QuickStepInterface/QSInterfaceController.m
```diff
@@ -634,10 +632,25 @@ - (void)executeCommand:(id)sender cont:(BOOL)cont encapsulate:(BOOL)encapsulate
...
+	}
+	
 	if (encapsulate) {
 		[self encapsulateCommand];
 		return;
 	}
+	if (!cont) {
+		// this ensures the interface is hidden before an action is run e.g. the 'capture screen region' and 'type text' actions needs this
+		[self hideMainWindowFromExecution:self]; // *** this should only hide if no result comes in like 2 seconds
+	}
```
tiennou: If you restore that one, maybe the `if (cont)`-else branch below should go ?

skurfer: Agreed. The last else can go.

pjrobertson: oh wait yeah, you mean line 660

skurfer:
I’ll go ahead and make the change to the if/else on this branch, and hopefully also add a fix for #2301.

pjrobertson:
sorry, you probably heard the pollution levels in Beijing were through the roof this week (I'm running a startup educating people about air pollution here in Beijing) so I didn't have a second to breath this week. Will make the change now

pjrobertson:
Done

skurfer:
Looks good. Thanks!

PR #2308, opened 2017-01-10
String dance

### LINKED ISSUE #2301: Mnemonics not being remembered

skurfer:
Right now, “Reset search after” is crippling the training/learning process. (And it’s enabled by default, so this probably affects a lot of people.) Possible solutions:

  1. Keep the matched string until the interface is dismissed or the selection is explicitly cleared
  2. Rip out the “Reset search after” feature entirely
  3. Add a yet another string (a 4th?) to keep track of the abbreviation until the interface is dismissed

I’ve attempted 1. Thoughts?

The first commit fixes a bug I ran across while working on this. (You couldn’t clear the selection when using the comma trick unless the current object was in the collection.) Seemed like the best solution.

Fixes #2301

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/QSCollectingSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSCollectingSearchObjectView.m
@@ -180,6 +180,10 @@
 }
 - (void)deleteBackward:(id)sender {
 	if ([collection count] && ![partialString length]) {
+		if (![collection containsObject:[super objectValue]]) {
+			// search string cleared, but main object was never added to the collection
+			[collection addObject:[super objectValue]];
+		}
 		[self uncollect:sender];
 	} else {
 		[super deleteBackward:sender];
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
@@ -906,7 +906,7 @@ NSMutableDictionary *bindingsDict = nil;
 		if ([resetTimer isValid]) {
 			[resetTimer setFireDate:[NSDate dateWithTimeIntervalSinceNow:resetDelay]];
 		} else {
-			resetTimer = [NSTimer scheduledTimerWithTimeInterval:resetDelay target:self selector:@selector(clearSearch) userInfo:nil repeats:NO];
+			resetTimer = [NSTimer scheduledTimerWithTimeInterval:resetDelay target:self selector:@selector(resetString) userInfo:nil repeats:NO];
 		}
 	}
 }


tiennou:
Sorry, the code changes doesn't make that clear to me, how is it crippling it ?

skurfer:
It wipes the abbreviation from existence before you have a chance to do anything that would trigger `saveMnemonic`.

tiennou:
Sorry, I still don't get it. Isn't wiping the abbreviation the point of resetting the search ?

skurfer:
Hard to say for sure, but I think there’s a reason for three or more strings to track what the user typed. I think the point of resetting the search is to allow you to start a new search, and that still works (by clearing `searchString`).

What it *doesn’t* do (with my changes) is clear the `matchedString`. So if the reset timer clears things, but the user hasn’t done anything yet, they still have an opportunity to set defaults in the results, and `saveMnemonic` still has enough information to do its job if they run an action or something.

tiennou:
Ok, I think I understand. The reset timer can fire while the user is doing nothing, which will clear the search string without deselecting the found object. Then, if you tried to manipulate mnemonics for that object, you'd have no way to know what was typed in the first place, hence the crippling.

I'm not sure how to go about it (it just makes the learning process be execution-only, albeit if you pause for too long you would get that behavior anyway). I'd favor 2, actually, because User Input is King™, and the delete key now should make it explicit.

Also, I'm wondering : what did we change to cause that behavior to happen ? Reset Search has been around for quite some time, and I can't remember anything recent that could have cause it to go so badly...

skurfer:
Force-pushed a much simpler fix. Thanks.

This reintroduces the potentially confusing behavior where the search can be reset via timer or by tabbing to another pane, but there’s no visual indication. I’ll take that over losing almost all mnemonics.

And for what it’s worth @tiennou, I have no objection to ripping out the reset timer but I want to get a fix out ASAP. I’ll ask about it in Slack.

PR #2312, opened 2017-01-25
speed up QSObject comparisons

skurfer:
An attempt to alleviate some of the delays we’ve seen related to UTI look-ups.

I also tried to more generally speed up comparisons between QSObjects since they seem to be far more common than I thought.

(One example: When an `NSSet` is copied, the objects aren’t just blindly thrown into a new set. As they get added, the new set checks *again* to make sure there are no duplicates, which triggers calls to `isEqual:` for everything in the set. This happens twice in `purgeImagesAndChildrenOlderThan` alone.)

I look forward to hearing how terrible some of these changes are and why. 😝 I was just going for raw speed, OK?

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -764,18 +764,22 @@ NSSize QSMaxIconSize;
 }
 
 - (NSString *)primaryType {
+	if (primaryType) {
+		return primaryType;
+	}
     if (!primaryType)
         primaryType = [meta objectForKey:kQSObjectPrimaryType];
 	if (!primaryType)
 		primaryType = [self guessPrimaryType];
-	if (!primaryType)
-		[self setPrimaryType:QSUTIForAnyTypeString(primaryType)];
+	if (primaryType)
+		[self setPrimaryType:primaryType];
 	return primaryType;
 }
 - (void)setPrimaryType:(NSString *)newPrimaryType {
     if (primaryType != newPrimaryType) {
         primaryType = newPrimaryType;
     }
+	newPrimaryType = QSUTIForAnyTypeString(newPrimaryType);
     [self setObject:newPrimaryType forMeta:kQSObjectPrimaryType];
 }
 
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -44,17 +44,16 @@ NSSize QSMaxIconSize;
         tempChildSet = [childLoadedSet copy];
     }
 
-    [tempIconSet enumerateObjectsWithOptions:NSEnumerationConcurrent usingBlock:^(QSObject *obj, BOOL *stop) {
-        if (obj->lastAccess && obj->lastAccess < (tempLastAccess - interval)) {
-            [obj unloadIcon];
-        }
-    }];
-
-    [tempChildSet enumerateObjectsWithOptions:NSEnumerationConcurrent usingBlock:^(QSObject *obj, BOOL *stop) {
-        if (obj->lastAccess && obj->lastAccess < (tempLastAccess - interval)) {
-            [obj unloadChildren];
-        }
-    }];
+	for (QSObject *obj in tempIconSet) {
+		if (obj->lastAccess && obj->lastAccess < (tempLastAccess - interval)) {
+			[obj unloadIcon];
+		}
+	}
+	for (QSObject *obj in tempChildSet) {
+		if (obj->lastAccess && obj->lastAccess < (tempLastAccess - interval)) {
+			[obj unloadChildren];
+		}
+	}
 }
 
 + (void)interfaceChanged {
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -769,7 +769,9 @@ NSSize QSMaxIconSize;
         primaryType = [meta objectForKey:kQSObjectPrimaryType];
 	if (!primaryType)
 		primaryType = [self guessPrimaryType];
-    return QSUTIForAnyTypeString(primaryType);
+	if (!primaryType)
+		[self setPrimaryType:QSUTIForAnyTypeString(primaryType)];
+	return primaryType;
 }
 - (void)setPrimaryType:(NSString *)newPrimaryType {
     if (primaryType != newPrimaryType) {
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -104,7 +104,7 @@ NSSize QSMaxIconSize;
 		}
 	} else {
 		for(NSString *key in data) {
-			if (![[data objectForKey:key] isEqual:[anObject objectForType:key]]) return NO;
+			if (![[data objectForKey:key] isEqual:[anObject->data objectForKey:key]]) return NO;
 		}
 	}
 	return YES;
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -80,11 +80,10 @@ NSSize QSMaxIconSize;
 
 - (NSUInteger)hash
 {
-	NSString *ident = [self identifier];
-	if (!ident) {
-		ident = [self stringValue];
+	if (!identifier) {
+		return (NSUInteger)self;
 	}
-	return [ident hash];
+	return [identifier hash];
 }
 
 - (BOOL)isEqual:(id)anObject {
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -92,19 +92,20 @@ NSSize QSMaxIconSize;
     anObject = [anObject object];
   }
 	if (self == anObject) return YES;
-	if (([self identifier] || [anObject identifier]) && ![[self identifier] isEqualToString:[anObject identifier]]) return NO;
+	QSObject *otherObject = (QSObject *)anObject;
+	if ((identifier || otherObject->identifier) && ![identifier isEqualToString:otherObject->identifier]) return NO;
 	if ([self count] > 1) {
-		if ([self count] != [(QSObject *)anObject count]) {
+		if ([self count] != [otherObject count]) {
 			return NO;
 		}
 		NSSet *myObjects = [NSSet setWithArray:[self splitObjects]];
-		NSSet *otherObjects = [NSSet setWithArray:[anObject splitObjects]];
+		NSSet *otherObjects = [NSSet setWithArray:[otherObject splitObjects]];
 		if (![myObjects isEqualToSet:otherObjects]) {
 			return NO;
 		}
 	} else {
 		for(NSString *key in data) {
-			if (![[data objectForKey:key] isEqual:[anObject objectForType:key]]) return NO;
+			if (![[data objectForKey:key] isEqual:[otherObject objectForType:key]]) return NO;
 		}
 	}
 	return YES;
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -85,26 +85,26 @@ NSSize QSMaxIconSize;
 	return [identifier hash];
 }
 
-- (BOOL)isEqual:(id)anObject {
+- (BOOL)isEqual:(QSObject *)anObject {
 	if (!anObject) return NO;
-  if (self != anObject && [anObject isKindOfClass:[QSRankedObject class]]) {
-    anObject = [anObject object];
-  }
+	if (self != anObject && [anObject isKindOfClass:[QSRankedObject class]]) {
+		anObject = [(QSRankedObject *)anObject object];
+	}
 	if (self == anObject) return YES;
-	QSObject *otherObject = (QSObject *)anObject;
-	if ((identifier || otherObject->identifier) && ![identifier isEqualToString:otherObject->identifier]) return NO;
+	NSString *otherIdentifier = anObject->identifier;
+	if ((identifier || otherIdentifier) && ![identifier isEqualToString:otherIdentifier]) return NO;
 	if ([self count] > 1) {
-		if ([self count] != [otherObject count]) {
+		if ([self count] != [anObject count]) {
 			return NO;
 		}
 		NSSet *myObjects = [NSSet setWithArray:[self splitObjects]];
-		NSSet *otherObjects = [NSSet setWithArray:[otherObject splitObjects]];
+		NSSet *otherObjects = [NSSet setWithArray:[anObject splitObjects]];
 		if (![myObjects isEqualToSet:otherObjects]) {
 			return NO;
 		}
 	} else {
 		for(NSString *key in data) {
-			if (![[data objectForKey:key] isEqual:[otherObject objectForType:key]]) return NO;
+			if (![[data objectForKey:key] isEqual:[anObject objectForType:key]]) return NO;
 		}
 	}
 	return YES;
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -87,6 +87,7 @@ NSSize QSMaxIconSize;
 }
 
 - (BOOL)isEqual:(id)anObject {
+	if (!anObject) return NO;
   if (self != anObject && [anObject isKindOfClass:[QSRankedObject class]]) {
     anObject = [anObject object];
   }
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -79,10 +79,7 @@ NSSize QSMaxIconSize;
 
 - (NSUInteger)hash
 {
-	if (!identifier) {
-		return (NSUInteger)self;
-	}
-	return [identifier hash];
+	return identifier.hash ^ data.hash;
 }
 
 - (BOOL)isEqual:(QSObject *)anObject {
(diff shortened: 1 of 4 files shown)


Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -80,31 +80,32 @@ - (id)init {
 
 - (NSUInteger)hash
 {
-	NSString *ident = [self identifier];
-	if (!ident) {
-		ident = [self stringValue];
+	if (!identifier) {
+		return (NSUInteger)self;
```
tiennou: `CFHash()` ? This is purely theorical though.

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -80,31 +80,32 @@ - (id)init {
...
 
 - (BOOL)isEqual:(id)anObject {
+	if (!anObject) return NO;
   if (self != anObject && [anObject isKindOfClass:[QSRankedObject class]]) {
     anObject = [anObject object];
   }
 	if (self == anObject) return YES;
-	if (([self identifier] || [anObject identifier]) && ![[self identifier] isEqualToString:[anObject identifier]]) return NO;
+	QSObject *otherObject = (QSObject *)anObject;
+	if ((identifier || otherObject->identifier) && ![identifier isEqualToString:otherObject->identifier]) return NO;
```
tiennou: I'd really much prefer caching `identifier` vs. using `->` here (even though "raw-speed"). I'm actually surprised that it works, isn't direct ivar access ARC-verboten ?

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -80,31 +80,32 @@ - (id)init {
...
 }
 
 - (BOOL)isEqual:(id)anObject {
+	if (!anObject) return NO;
   if (self != anObject && [anObject isKindOfClass:[QSRankedObject class]]) {
     anObject = [anObject object];
   }
 	if (self == anObject) return YES;
-	if (([self identifier] || [anObject identifier]) && ![[self identifier] isEqualToString:[anObject identifier]]) return NO;
+	QSObject *otherObject = (QSObject *)anObject;
```
tiennou: This is nice, change the parameter !

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: It’s not guaranteed unique, though I don’t know how important that is. The only things that really get to this line are pure text objects. But wouldn’t we have to convert the type for both the input and output if using that function?

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: I can’t tell if you’re being serious or I’ve done something bad. 😄

Since `identifier` is part of the QSObject protocol, I wanted to just somehow say `anObject` conforms to the protocol, but couldn’t figure out a way to do that after the fact.

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: There’s not even a warning. Maybe it’s only forbidden if you use properties?

What do you mean by “caching `identifier`”? It’s effectively already cached in the iVar. I guess the question is “What’s the best way to access it?”.

Quicksilver/Code-QuickStepCore/QSObject.m
tiennou: Yeah, I was thinking about that uniqueness thingy we might or might not have.

The theory was that using `self` directly would make objects allocated closely together would have mostly the same pointer value, while `CFHash` would at least garble that a little.

*But*

`CFHash` might just call through to `-hash` for Obj-C objects anyways 😆 .

Quicksilver/Code-QuickStepCore/QSObject.m
tiennou: I was being serious, just go all the way and change the methods' parameters. Makes it easier on the brain.

Also, you'll be forced to reindent lines 91:92, so there's that too 😜 .

Oh, I now understand. It's the cast you're after... If the compiler is nice enough, maybe change the parameter type ? `id <QSObject`, or even `QSObject *` ?

Quicksilver/Code-QuickStepCore/QSObject.m
tiennou: caching = stash it in a local variable, like `otherIdentifier`. I'm always wondering if the compiler is clever enough to do it by itself or not 😉 .

I'm really curious on those direct access thing. I was pretty sure it was gone because of Obj-C 2 and the way objects are composed via linker symbols instead of structs now... But if it ain't broke...

skurfer: I could do this to create a local variable, and then we wouldn’t even need `otherObject`:

~~~obj-c
NSString *otherIdentifier = ((QSObject *)anObject)->identifier;
~~~

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: I’d like to, but what if we get passed a `QSRankedObject`? I guess we aren’t sending it any messages it won’t understand, so it should be fine, right?

tiennou: Would changing the parameter to `QSObject*` and cast the `-[QSRankedObject object]` line work ?

skurfer:
> Would changing the parameter to `QSObject*` and cast the `-[QSRankedObject object]` line work ?

Seems to work. I’ve pushed the change here, and am trying it out locally.

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -44,17 +44,16 @@ + (void)purgeImagesAndChildrenOlderThan:(NSTimeInterval)interval {
         tempChildSet = [childLoadedSet copy];
     }
 
-    [tempIconSet enumerateObjectsWithOptions:NSEnumerationConcurrent usingBlock:^(QSObject *obj, BOOL *stop) {
```
pjrobertson: You're probably right on this, although it's quite possible that the code inside the block is 'a lot of work', which means calling it concurrently may well be faster.
Is `unloadIcon` an intensive method?

pjrobertson:
oooh I'd love to chime in here, I do love myself a good bit of 'optimisation' ;-)
I'm in Mongolia right now so am a little bit handicapped in terms of helping out, but I've looked over the code :)

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -85,26 +85,26 @@ - (NSUInteger)hash
...
+- (BOOL)isEqual:(QSObject *)anObject {
 	if (!anObject) return NO;
-  if (self != anObject && [anObject isKindOfClass:[QSRankedObject class]]) {
-    anObject = [anObject object];
-  }
+	if (self != anObject && [anObject isKindOfClass:[QSRankedObject class]]) {
+		anObject = [(QSRankedObject *)anObject object];
+	}
 	if (self == anObject) return YES;
-	QSObject *otherObject = (QSObject *)anObject;
```
pjrobertson: Were most of these changes just done to  get rid of warnings? I don't think any of this will have improved performance. My guess is the compiler is way smarter than us anyway :P


what happens if I run: `[myQSObject isEqual:@"some string"]`, will this die?

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -769,7 +769,9 @@ - (NSString *)primaryType {
         primaryType = [meta objectForKey:kQSObjectPrimaryType];
 	if (!primaryType)
 		primaryType = [self guessPrimaryType];
-    return QSUTIForAnyTypeString(primaryType);
+	if (!primaryType)
```
pjrobertson: seems what you've changed this to is slightly different. Surely it should be `if(primaryType)` no?

Calling `QSUTIForAnyTypeString()` on `nil` (as you're doing here) is useless (or is it?)

Once you've changed this to what I presume it should be (`if (primaryType) {}`), then I can think of a race condition that may well throw some bugs when this is released.

The only other place `[meta objectForKey:kQSObjectPrimaryType]` is called (directly) is in `-[QSObject extractMetadata]`. That's for loading a saved object (e.g. trigger etc. - I think). Now if an object was saved to disk before this change, and then reloaded in `extractMetadata`, the old type of `primaryType` (instead of the UTI will be used).

Basically, I think both here and in `extractMetadata` we need to call `QSIsUTI()`.
Perhaps the actual if statement here should be:

`if (primaryType && !QSIsUTI(primaryType)) { ... }`

Quicksilver/Code-QuickStepCore/QSObject.m
pjrobertson: > caching = stash it in a local variable, like otherIdentifier. I'm always wondering if the compiler is clever enough to do it by itself or not 😉 .

Pretty sure the compiler is clever enough! ;-)
This code is cleaner/easier to read, so is probably the right way to go.

Quicksilver/Code-QuickStepCore/QSObject.m
pjrobertson: So using the pointer of the object would mean to identical objects (in our minds) would not be considered equal. e.g.:


```
QSObject *one = [QSObject objectWithString:@"a"];
QSObject *two = [QSObject objectWithString:@"a"];

BOOL isEqual = [[one hash] isEqual:[two hash]];
// isEqual == False
```

There two objects are the same, but the pointer isn\t, so we can't really do what you're proposing... I think?

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: There were no warnings, but there would be errors that prevent building if we don’t tell it the other object is a QSObject before we start sending messages to it.

I suspect passing a string here would be bad, but that would have always been true. We could check the class again early on, but I’m hesitant to add more code that will get called every single time. Especially since the case we’re guarding against would have made itself known very loudly by now if it was a problem.

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: No. It checks a couple of booleans and sets some things to `nil`/`NO`.

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: The thinking was “if there’s no type set so far, go get one” but I suppose you’re right that we’d be asking it to get they type for `nil`, which is pointless. I’ll give it some more thought.

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: I think what you’re saying is what will happen. I was going to add that to the unit test for `isEqual:` butI can’t get tests to run at the moment. In any case, I think `stringValue` is too expensive/complicated to call from `hash`. Other ideas?

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: So the goal here was to avoid calling `QSUTIForAnyTypeString()` for every call to `primaryType`. Now that I look at it, should we be calling that function here at all?

I’m thinking we should just call it in `setPrimaryType:` and `extractMetaData`, but not in `primaryType` since that should mostly be about getting an established value and not about computing that value.

skurfer:
See the last commit. It should “fix” `extractMetadata` for free since that calls `setPrimaryType:`.

skurfer:
OK @pjrobertson, I got the tests running again and added some for string objects. They actually work as expected, which is… not what I expected. 😛 

All the existing tests for `isEqual:` are passing, too. Should have checked that earlier.

skurfer:
Another questionable commit with direct iVar access. 😛 

Tests are still passing, for what that’s worth.

skurfer:
I’ve been able to reproduce some of these “hangs” reported by users, which aren’t really hangs, just QS doing a ton of comparisons. I found that `hash` is a big part of the problem. Commenting it out reduces launch time from 35s to 10s. Any ideas on how we can improve that? I thought `hash` was supposed to make things faster. 😒

skurfer:
Now we’ve got something. Thanks to a huge borrowed `.bash_history` file, I was able to dig in with the Time Profiler in Instruments and make some optimizations. (A bit of an understatement.) Here are some launch times:

With a large catalog cached…

This branch: 4.5s
`master`: 5+ minutes (I didn’t wait for it to finish)

One major change is that before, if identifiers were equal, it would keep checking just in case one had more data then the other or something. Now, if identifiers are equal, the objects are assumed to be equal.

I also removed some UTI code that I don’t think we need any more. Catalogs and triggers should have been regenerated on everyone’s systems by now, right?

Something else I want to bring up… With caches cleared, launch time is less than 2s. I have to ask why we’re caching these things to disk if loading them takes more than twice as long as just recreating them from the source. (Some of this is surely related to scanning on a background thread vs. loading from cache on the main thread, but still, scanning is fast.) My first thought was to rip out the caching altogether, but there could be some network bound object sources that would benefit from it. So, I vote we keep the caching stuff in place, but start telling the vast majority of object sources not to use it (by returning `NO` from `entryCanBeIndexed:` or fiddling with the default behavior in `canBeIndexed`).

Quicksilver/Code-QuickStepCore/QSObject_PropertyList.m
```diff
@@ -96,7 +96,7 @@ - (id)initWithDictionary:(NSDictionary *)dictionary {
 
     // Check immediately if we already have loaded that object
     // ***warning  * should this update the name for files?
-    id dup = [QSLib objectWithIdentifier:self.identifier];
+    id dup = [QSLib objectWithIdentifier:identifier];
```
pjrobertson: I don't like your wording in the commit message:

> the intended value *should* be there

For the sake of bypassing an `if()` statement you're doing this?
Although looking at `-[QSObject identifier]` there's an `@synchronised` in there. It seems that's what'll be slowing things down. I suggest that only the code inside the `if (!identifier) {` block be put in the `@synchronised` block, then this be changed back.

I'll push a commit (which you can reject if you like)

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -79,10 +79,7 @@ - (id)init {
 
 - (NSUInteger)hash
 {
-	if (!identifier) {
-		return (NSUInteger)self;
-	}
-	return [identifier hash];
+	return identifier.hash ^ data.hash;
```
pjrobertson: what's the chance of a collision here? Probably very rare, and probably a risk worth taking for the speed improvement this brings. *but* I would still add a comment in here about what you've done

Maybe you saw this? http://nshipster.com/equality/

What are the chances an object has no identifier or data?

Quicksilver/Code-QuickStepCore/QSObject_PropertyList.m
```diff
@@ -99,19 +99,6 @@ - (id)initWithDictionary:(NSDictionary *)dictionary {
     id dup = [QSLib objectWithIdentifier:identifier];
     if (dup) return dup;
 
-    // Backwards compatibility: make sure all dict keys are UTIs (where applicable)
```
pjrobertson: I'm fine with this

pjrobertson:
Wow, so that's a pretty drastic improvement! What task were you performing that was taking 4.5s vs 5+ minutes? Just loading the catalog or what?

You changes seem pretty good all round though, I'm looking forward to seeing some speed improvements!

> Something else I want to bring up… With caches cleared, launch time is less than 2s. I have to ask why we’re caching these things to disk if loading them takes more than twice as long as just recreating them from the source.

I can think of network stuff, and perhaps some plugins (like web search module)?
QS has been crashing lately for me and we still have that code to clear the cache on next launch after a crash... I'm often left waiting about 5s for the catalog to rescan which is really infuriating when I've just launched QS and want to do something *quick*. 

Caches are loaded in the background, not on launch right, so that 2s comment is not relevant?

Quicksilver/Code-QuickStepCore/QSObject_PropertyList.m
pjrobertson: OK no, my suggestion doesn't work, because there still could be a race condition whereby an object could be asked for its identifier from two places at the same time - doing what I suggest means the identifier would be set twice. Hmmm

Quicksilver/Code-QuickStepCore/QSObject.m
skurfer: Yes, I saw that and it was very helpful. In [the hashing section][1], he specifically addresses the concern about collisions:

> One of the most common misconceptions about implementing a custom hash function comes from affirming the consequent, thinking that hash values must be distinct.

So I don’t think we need to worry, as long as we’re following the rules, which we are. To be equal, objects have to have the same identifier and data, so they will also have an equal hash.

> What are the chances an object has no identifier or data?

Zero… I think? I’m pretty sure the only thing without an identifier is a plain string, and that will at least have an object for the key `QSTextType` in `data`.

[1]: http://nshipster.com/equality/#hashing

[1]: http://nshipster.com/equality/#hashing

Quicksilver/Code-QuickStepCore/QSObject_PropertyList.m
skurfer: Heh. You’re the reason I try to avoid saying “should”, but it still slips through.

But it’s true, right? If you’re pulling an object out of a file, the identifier should be there.

skurfer:
> Wow, so that's a pretty drastic improvement! What task were you performing that was taking 4.5s vs 5+ minutes? Just loading the catalog or what?

I was just launching the app (with caches populated) and waiting for the spinning “QS” to go away. (Maybe you have that disabled, but it’s a good indicator of when you can start using the UI.)

I think the biggest impact came from all the UTI lookups (via `primaryType` mostly). Of course we need them, but I’ve rearranged things to minimize those calls.

Something else that made a huge difference was assuming equality if we saw equal identifiers. The code below that test is expensive and was getting hit a lot more frequently than I expected.

(Note: This is why the identifier for a string can’t just be the string. If you typed/pasted an identifier into QS as a string, that string would clobber the real object.)

> Caches are loaded in the background, not on launch right, so that 2s comment is not relevant?

I didn’t trace it through the call stack, but given the way it hangs the UI, I don’t think it’s running in the background.

PR #2317, opened 2017-02-09
Add release dates to changelog

danielcompton:
It's often helpful to see the release dates in the changelog.

### DIFF
--- a/Quicksilver/SharedSupport/CHANGELOG
+++ b/Quicksilver/SharedSupport/CHANGELOG
@@ -1,4 +1,4 @@
-## Changes - 1.5.2 (401D) ##
+## Changes - 1.5.2 (401D) 2017-01-18 ##
 
 ### Fixed ###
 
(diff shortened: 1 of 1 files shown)


skurfer:
I guess I don’t have a problem with doing this retroactively. FYI, the reason it isn’t typically done is that the release notes are built into the app and we don’t necessarily know when the actual date will be. We might put a new version as a developer preview for a month. If it looks good, the same build will just get flagged as a “pre-release” in the update system. If it still seems OK, we’ll flag it as a final release sometime later. So the “release” date is often several weeks after the build date.

In the long run, those weeks probably don’t matter, so maybe we should start putting the date of the initial build. Yeah, you’ve talked me into it.

FYI, there are dates listed on https://qsapp.com/archives/ if you want to add any more. Otherwise, I’ll merge this.

danielcompton:
Cool, I've added the released dates in for all of the changelog.

skurfer:
Thanks!

PR #2320, opened 2017-02-18
clean up some old stuff

skurfer:
I noticed a warning about the droplet NIB. It got overlooked when I updated the deployment target. That’s all that’s changed.

And Internet Explorer… Probably safe to drop that, right?


pjrobertson:
Any ideas why Travis failed?

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

skurfer:
> Any ideas why Travis failed?

Because it fails on every other run no matter what? I did see long stalls (30+ minutes) when I was using a 10.10 VM to build, but it did eventually finish. Maybe it’s running into the same thing. It happens during testing, I think.

PR #2322, opened 2017-02-23
better proxy caching

skurfer:
I think I’ve finally figured out how to get rid of the timer.

There are generally two places we want to resolve a proxy object: to show it in the interface, and to use it in a command. Once resolved, we want to hold onto that value until one of two things happens: The command is finished using the object, or the interface was dismissed.

Common scenarios (and their effect on the proxy):

Select and do nothing.

1. Select a proxy in the interface (resolve)
2. Dismiss the interface without doing anything (release)

Run a trigger.

1. Run a command that uses a proxy object (resolve)
2. The trigger’s command finishes (release)

Those are fine on their own, but an even more common scenario combines them in a weird order.

1. Select a proxy in the interface (resolve)
2. Hit Return to run the command, which dismisses the interface (release)
3. Run the command (resolve)
4. The command finishes (release)

The problem with that was, between 1 and 3, the object could change so you could run the action on something you didn’t want.

The previous solution was to just hold onto the resolved object for 2 more seconds, but that broke workflows that used the same proxy in rapid succession.

Now, when the interface is dismissed, instead of blindly clearing the proxy cache, it will check to see *why* the interface was dismissed. If it’s because a command was run, it leaves the cache in place knowing that it will get cleared anyway when the command finishes. Otherwise, the cache is cleared immediately.

The only thing I’m not 100% comfortable with is the change to when the notifications are posted. It used to be [more limited](https://github.com/quicksilver/Quicksilver/pull/2322/commits/db95bcf46035210f798d2c302972bc1ccc50922f#diff-67e3a94a52a5e4a92915f1f10797045bL199) it seems, but I don’t know if that made a difference in practice.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSDefines.h
+++ b/Quicksilver/Code-QuickStepCore/QSDefines.h
@@ -162,4 +162,8 @@
  *  E.g. if 10.7 is specified, 10.6.8, 10.6.9... would all be supported
  */
 #define kPluginRequirementsOSUnsupportedVersion @"osUnsupported"
-
+/**
+ *  Type: string
+ *  Key to the reason the interface was hidden/deactivated
+ */
+#define kQSInterfaceDeactivatedReason @"deactivate reason"
--- a/Quicksilver/Code-QuickStepCore/QSNotifications.h
+++ b/Quicksilver/Code-QuickStepCore/QSNotifications.h
@@ -2,6 +2,8 @@
 #define QSReleaseAllCachesNotification @"QSReleaseAllCachesNotification"
 #define QSReleaseAllNotification @"QSReleaseAllNotification"
 #define QSInterfaceChangedNotification @"QSInterfaceChangedNotification"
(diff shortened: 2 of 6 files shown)


pjrobertson:
Hmmm yeah so you've basically moved the notification to take place earlier on - I have no idea of the effects this will have on how things work. I'm guessing you're running with this build now, places I think to look out for are probably:

abbreviations, icon clearing/caches, proxies (duh!), triggers

skurfer:
> Hmmm yeah so you've basically moved the notification to take place earlier on

Actually, I think it’s later (since it used to be in the `willHideMainWindow` method) but not meaningfully so. I’m more concerned that it might fire at times when it didn’t previously since it’s no longer inside that `if`.

But the only place watching for that notification before this was the trigger center which re-scoped when focus changed. I’ve tested some of my scoped triggers and all appears to be well.

And I just touched a few files in `~/Downloads`, then selected and trashed them via QS as fast as I could. Always seemed to get the right target.

PR #2325, opened 2017-03-03
don’t add identifiers to string objects

skurfer:
Of course `git blame` said “look in the mirror” on this one.


pjrobertson:
haha, smart `git blame`! ;-)

Why does giving a string object replace its identifier? Because its identifier is the string value right? I remember in the past I tried to give string objects the identifier `QSString:[the string value]` but that still caused problems with triggers. I guess that's what was happening again?

Damn this thing is messy!

skurfer:
I found the old discussions.

https://github.com/quicksilver/Quicksilver/pull/1505#issuecomment-19392083
https://github.com/quicksilver/Quicksilver/pull/1569

Sounds like we were going to try to fix it “for real” one day so everything could safely have an identifier. I guess that day hasn’t come yet. 😃 

I re-introduced the problem (for reasons I can’t explain) when moving the object dictionary to `QSLibrarian`.

PR #2342, opened 2017-04-15
QSSense fixup

tiennou:
This is also a really old branch (2012 was a good year !).

So, we have 2 different versions of the ranking algorithm in-tree, both dating back from the Dawn of Time. The main ranking loop would use QSSense (through `QSStringRanker`). I haven't seen any use of the `NSString` category, but since it's hidden behind forwarding, it might be used sometimes.

~~They *had* a differing behavior (but I never found the time to thoroughly check).~~

The differing behavior was introduced in 87d328a actually. The comparison options are also a give-away (QSSense is case and diacritic insensitive and respect localization, the other is just case-insensitive.

So THE DIFFERENCE is what it is.

I've added a test case for it, because I wanted to 1) see it work 2) see what was different.

### DIFF
--- a/Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
+++ b/Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
@@ -50,13 +50,13 @@ const float ACC = 0.00001;
 	XCTAssertEqualWithAccuracy(score, 0.90909, ACC);
 
 	score = [str scoreForAbbreviation:@"ts"];
-	XCTAssertEqualWithAccuracy(score, 0.83636, ACC);
+	XCTAssertEqualWithAccuracy(score, 0.92727, ACC);
 
 	score = [str scoreForAbbreviation:@"te"];
 	XCTAssertEqualWithAccuracy(score, 0.91818, ACC);
 
 	score = [str scoreForAbbreviation:@"tet"];
-	XCTAssertEqualWithAccuracy(score, 0.84545, ACC);
+	XCTAssertEqualWithAccuracy(score, 0.93636, ACC);
 
 	score = [str scoreForAbbreviation:@"str"];
 	XCTAssertEqualWithAccuracy(score, 0.91818, ACC);
@@ -65,7 +65,7 @@ const float ACC = 0.00001;
 	XCTAssertEqualWithAccuracy(score, 0.93181, ACC);
 
 	score = [str scoreForAbbreviation:@"ng"];
-	XCTAssertEqualWithAccuracy(score, 0.18181, ACC);
+	XCTAssertEqualWithAccuracy(score, 0.63636, ACC);
 }
 
 - (void)testScoreLongString {
@@ -108,13 +108,13 @@ const float ACC = 0.00001;
 	XCTAssertEqualWithAccuracy(score, 0.90222, ACC);
 
 	score = [str scoreForAbbreviation:@"ts"];
-	XCTAssertEqualWithAccuracy(score, 0.86444, ACC);
+	XCTAssertEqualWithAccuracy(score, 0.88666, ACC);
 
 	score = [str scoreForAbbreviation:@"te"];
 	XCTAssertEqualWithAccuracy(score, 0.80777, ACC);
 
 	score = [str scoreForAbbreviation:@"tet"];
-	XCTAssertEqualWithAccuracy(score, 0.79000, ACC);
+	XCTAssertEqualWithAccuracy(score, 0.81222, ACC);
 
 	score = [str scoreForAbbreviation:@"str"];
 	XCTAssertEqualWithAccuracy(score, 0.78555, ACC);
@@ -129,7 +129,7 @@ const float ACC = 0.00001;
 	XCTAssertEqualWithAccuracy(score, 0.75888, ACC);
 
 	score = [str scoreForAbbreviation:@"ng"];
-	XCTAssertEqualWithAccuracy(score, 0.52444, ACC);
+	XCTAssertEqualWithAccuracy(score, 0.74666, ACC);
 }
 
 - (void)testLongString {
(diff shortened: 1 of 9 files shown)


skurfer:
Cool!

Strange that Git doesn’t show `QSSense.m` getting moved or removed, but the original is gone, so I guess it “knows” somehow.

tiennou:
I think it's just the way GitHub show (or more precisely, doesn't) renames though.

Also, you might want to take a look at the test case. It has some "interesting" characteristics I didn't expect.

Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
```diff
@@ -0,0 +1,153 @@
...
+	NSIndexSet *results = nil;
+
+	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	XCTAssertEqualWithAccuracy(score, 0.74074, ACC);
+	results = [NSIndexSet indexSetFromArray:@[@(0), @(5), @(15), @(16), @(22), @(23)]];
+	XCTAssertEqualObjects(indexes, results);
+
+	strRange.length += STEP;
+	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	XCTAssertEqualWithAccuracy(score, 0.76129, ACC);
```
tiennou: Like this. It actually matches higher just because I added one word (which I don't think should matter) anyway.

Just noticing my test is not what I expected too, I forgot there's a 't' 2nd place in "string" 😆.

pjrobertson: Can you add some comments to this section, I got my head round it, but it took me a while (e.g. line 89 you could say what each index in the `NSIndexSet` represents)

So here you're saying matching `test` with `This excellent string tells ` gives a *higher* score than matching without the space at the end (`This excellent string tells`)? Seems strange to me!

Also can you explain what `ACC` is?

Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
```diff
@@ -0,0 +1,153 @@
...
+	NSMutableIndexSet *indexes = [[NSMutableIndexSet alloc] init];
+	NSIndexSet *results = nil;
+
+	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	XCTAssertEqualWithAccuracy(score, 0.74074, ACC);
+	results = [NSIndexSet indexSetFromArray:@[@(0), @(5), @(15), @(16), @(22), @(23)]];
+	XCTAssertEqualObjects(indexes, results);
+
+	strRange.length += STEP;
+	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
```
pjrobertson: Also shouldn't you use `abbr` here instead of re-defining `CFSTR("test")`?

tiennou: Good catch. Will fix.

Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
tiennou: Will do. I was actually surprised that the index set is for each subsequent matching character of the abbreviation in the search string, not just "the best match", as I expected (ie, it doesn't correspond to what is shown underlined in the UI). 

Yes, that's the weird thing, scoring increases proportionally to the string's length. I think it's that way because of the change you made (which is why I built the test case), because the other test case doesn't behave like that.

`ACC` is just a constant I made up because non-determinism related to floating-point calculation.

tiennou: The test case [here](https://github.com/quicksilver/Quicksilver/pull/2342/commits/5cba78fae1ba0df5b8a145ce1cfd15cceea9e893#diff-86c92ca75387e03f87312001fe115fb9R204)

skurfer:
Grrrr. For some reason, my scheme is missing everything but “Quicksilver Tests”. I guess I need to manually put the rest back.

Should we also add some tests for strings that _don’t_ match?

skurfer:
Another thought on the tests…

Do we really need to check for a specific score? A small change to the algorithm down the road doesn’t mean the score is “wrong”. Should we adjust the tests to just check that the scores for certain strings are higher or lower than others?

Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
fwextensions: @tiennou Perhaps it's bad form to comment on a year-old PR thread (and perhaps this has already been figured out), but I also ran into the index set issue in porting the scoring algorithm to JavaScript (and I've been poking into old PRs here to try to understand why the scores in `TestQSSense.m` don't quite line up with the results of my code).

Consider matching `"test"` against the string `"t e s t tess!"`.  The full abbreviation doesn't appear as a substring in the first loop, so it's reduced to `"tes"` and the function recurses.  That substring is found near the end of the string, and those indices (`[8, 9, 10]`, I think) get set in the mask.   Then it looks for the remaining `"t"` in the rest of the string, but only finds `"s!"`, so it has to start over and shrink the abbreviation again.  Eventually, it finds the individual letters separated by spaces and returns a score.  But the mask still contains those first hits that it gave up on.

My solution was to clear the mask just before the `return 0` at the end of the function.  I believe that's correct, as the 0 score means the algorithm will start over with a shorter abbreviation, and that shorter bit might hit in a different place, rendering the previous hits obsolete.  If the UI isn't underlining all those letters, then something else must be cleaning up the mask.  (I tried clamping it to the length of the abbreviation, but I don't think that gives the correct results in all cases.)

PR #2344, opened 2017-04-17
QSObject cleanup

tiennou:
This is the result of my weekend's code-crunching-fest. Now I'm depressed.

I enabled TSan. I noticed a bunch of races, and tried to fix them all. There's more in `QSIconLoader`, which I tried to look at but ended up resetting everything. The icon loading system is hopeless.

This has broken icon loading, and guaranteed deadlocks, instead of weird crashes (did you know you could nil objects *inside* a set ? I didn't. At one point I had an `iconLoadedSet` with mostly nils inside, which exception-ed on `-copy`).

Also, exception aren't catchable anymore. See those `@try/@catch` handlers ? They're useless, and the runloop will skip whatever it's doing and start anew, dropping whatever you were trying to accomplish.

So "Contacts" raises because it expects an array as its object identifier (PR for that), and "Displays" because it.. fails to do stuff with strings.

At least we have guaranteed deadlocks ☹️. I'll try to run that, as I'm trying to pin-point the weird autorelease crash we have.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject.h
@@ -38,17 +38,6 @@ extern NSSize QSMaxIconSize;
 #define kQSContents               @"QSObjectContents"
 #define kQSObjectComponents       @"QSObjectComponents"
 
-typedef struct _QSObjectFlags {
-	unsigned int		multiTyped:1;
-	unsigned int		iconLoaded:1;
-	unsigned int		childrenLoaded:1;
-	unsigned int		contentsLoaded:1;
-	unsigned int		noIdentifier:1;
-	unsigned int		isProxy:1;
-	unsigned int		retainsIcon:1;
-	//  NSCellType		  type:2;
-} QSObjectFlags;
-
 @interface QSObject : QSBasicObject <NSCopying, NSCoding> {
 	NSMutableDictionary *data QS_DEPRECATED; /* Temporary */
 }
--- a/Quicksilver/Code-QuickStepCore/QSObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject.h
@@ -85,8 +85,6 @@ extern NSSize QSMaxIconSize;
 - (BOOL)loadIcon;
 - (BOOL)unloadIcon;
 
-- (void)updateIcon:(NSImage *)newIcon QS_DEPRECATED_MSG("Use -setIcon:");
-
 /** Type-handling */
 
 - (NSArray *)types;
--- a/Quicksilver/Code-QuickStepCore/QSObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject.h
@@ -7,8 +7,6 @@
 
 extern NSSize QSMaxIconSize;
 
-#define itemForKey(k) [data objectForKey:k]
-
 // meta dictionary keys
 #define kQSObjectPrimaryName      @"QSObjectName"
 #define kQSObjectAlternateName    @"QSObjectLabel"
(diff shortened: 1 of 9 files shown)


tiennou:
Also, you might find that useful : https://gist.github.com/tiennou/6519afd04954abd2f12dfdffc932f3ef

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -88,276 +149,199 @@ - (BOOL)isEqual:(QSObject *)anObject {
 		anObject = [(QSRankedObject *)anObject object];
 	}
 	if (self == anObject) return YES;
-	NSString *otherIdentifier = anObject->identifier;
-	if ((identifier || otherIdentifier) && [identifier isEqualToString:otherIdentifier]) {
-		return YES;
+
+	NSString *otherIdentifier = anObject.identifier;
```
skurfer: The `identifier` getter was too expensive with even more expensive side-effects, depending on the object. That’s why it was using direct access (and still should I think).

Quicksilver/Code-QuickStepCore/QSProxyObject.m
```diff
@@ -81,28 +88,30 @@ - (QSObject*)proxyObject {
...
+		NSString *reason = notif.userInfo[kQSInterfaceDeactivatedReason];
 		if ([reason isEqualToString:@"execution"]) {
 			// if the interface is hiding from running a command, keep the cached value
 			// it will get cleared after the command executes
 			return;
 		}
 	}
-	[cache removeObjectForKey:QSProxyTargetCache];
+	// Send the notification before niling our target
+	/* FIXME: I'm not actually quite sure what we're trying to do here... */
```
skurfer: Reasons proxies watch for these notifications:

* If the icon for the target object changes, the icon for the proxy needs to change as well (since that’s what the interface will be showing)
* If the interface is dismissed, we don’t need to hold onto the target.
* If the command is executed, we don’t need to hold onto the target. Yes, that also dismisses the interface, but we still need to watch both. See #2322 for details.

Any more questions?

tiennou: I think my actual question is : "Does it make sense to unregister the specific object we have in the cache ?". 

Since the current code has just `nil`-ed the cached object, it actually unregister that proxy with any `IconModified` notification (eg. `nil`), while the one I'm proposing is specific to the object currently cached. I'd be tempted to change that to `object:nil`, I just wanted to be sure I wasn't missing something 😉 .

Quicksilver/Code-QuickStepCore/QSObject.m
tiennou: Yeah, but direct access could be wrong if we don't lock...

tiennou:
I've been thinking about it some more... The problem we face is that `QSObject` is too mutable for it's own good, and hence we need locks everywhere, which kills performance (you can try arrowing results, it's pretty bad).

I think we need a way to mark objects as "frozen", so object sources can build them, but as soon as they enter the catalog (IOW being returned from `-scannedObjects`), `-freeze` them, which would drop the locking and forbid any weird mutation (changing anything that's considered "cache", ie anything in `-meta`). This would not work that well with multi-typed objects (eg. applications are a good example) though, maybe if `-data` is kept out of the freezed things ?

Quicksilver/Code-QuickStepCore/QSProxyObject.m
skurfer: Can we just drop all three `removeObserver:` calls at the end of this method, then?

Quicksilver/Code-QuickStepCore/QSObject.h
```diff
@@ -40,155 +38,108 @@ extern NSSize QSMaxIconSize;
...
+/** Icons */
+
+@property (retain) NSImage *icon;
+@property BOOL retainsIcon;
+@property (readonly) BOOL iconLoaded;
+
+- (BOOL)loadIcon;
+- (BOOL)unloadIcon;
+
+- (void)updateIcon:(NSImage *)newIcon QS_DEPRECATED_MSG("Use -setIcon:");
```
skurfer: It should be safe to just drop this method now.

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -368,195 +362,192 @@ - (NSString *)details {
...
+	if (_primaryType != newPrimaryType) {
+		_primaryType = newPrimaryType;
+	}
 
-- (id)objectForCache:(id)aKey {
-    return [cache objectForKey:aKey];
+	/* FIXME: This can cause primaryType to have a different value between ivar
+	 * and _meta. At first, it's doesn't seems so bad, but a deserialized object
+	 * after a restart would get the UTI instead of the original primaryType
+	 */
```
skurfer: This should keep them in sync…

~~~
if (_primaryType == newPrimaryType) {
	return;
}
newPrimaryType = QSUTIForAnyTypeString(newPrimaryType);
_primaryType = newPrimaryType;
[self setObject:newPrimaryType forMeta:kQSObjectPrimaryType];
~~~

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -368,195 +362,192 @@ - (NSString *)details {
...
+	return self.cache[kQSObjectAltChildren];
 }
 
-- (BOOL)isProxyObject
-{
-    return NO;
+- (void)setAltChildren:(NSArray *)newAltChildren {
+	if (newAltChildren) {
+		/* FIXME: Looks like a bug */
+		if (self.cache[kQSObjectChildren] != newAltChildren) {
```
skurfer: I agree that this looks wrong. Let’s change it to the correct key.

Quicksilver/Code-QuickStepCore/QSProxyObject.m
```diff
@@ -111,66 +120,57 @@ - (BOOL)enabled {
...
 }
 
-- (BOOL)loadIcon
-{
+- (BOOL)loadIcon {
     NSString *namedIcon = [self objectForMeta:kQSObjectIconName];
     if (!namedIcon || [namedIcon isEqualToString:@"ProxyIcon"]) {
         // use the resolved object's icon instead
-        QSObject *resolved = [self resolvedObject];
-        [self setIcon:[resolved icon]];
```
skurfer: What happens if we never set the loaded icon?

tiennou: Unless I'm missing something, nothing different ? We're already broken in that case, since we don't have `-icon`/`-setIcon:` overrides here, we'll forward those to our proxyied object implementation instead, which will (or will not, as you imply) set a icon.

Frankly, I have no idea what's happening w.r.t proxy icons. We're registering a IconModified notification on an unhandled -objectIconModified: selector, and this methods reads as : "if we have a icon-name meta-key and it's @"ProxyIcon", grab the quick-icon and set that as our icon, else ask our superclass to (which will mostly do the same thing but could ask the handler)", which to me means we might never get the "final" icon anyway.

skurfer: > since we don't have `-icon`/`-setIcon:` overrides here, we'll forward those to our proxyied object implementation instead

It calls the implementations from `QSObject`, but it calls them on the proxy object (`self`), not the thing the proxy points to (`resolvedObject`).

I don’t know what to say about `loadIcon` except that I’m pretty sure the existing implementation is correct. I’d summarize it this way: “If we want the resolved object’s icon, take the current value now, then start loading the real icon. If not, just ignore this implementation and do whatever you would have done anyway in `QSObject`.”

I tried your changes, and proxy objects just have a question mark icon and the real icon never appears.

A good way to test both icon loading and cache invalidation for proxies is to select “Random Track” then hide and show the interface over and over. You should see a new object with the correct icon every time.

Another good way to test icons (since Random Track might load them too quickly) is to download a large PDF, then select the Latest Download proxy. If things are working correctly. you should see

skurfer:
The new changes look good. I think proxy icons are still broken, but I haven’t tested that.

tiennou:
Yep, I have a WIP commit on moving the proxy icon loading in `QSProxyObjectHandler` instead of overriding `-loadIcon` and friends. I'm just not sure on what the current behavior is w.r.t to showing the "cube" proxy icon, showing the proxied object quick-icon, or showing the proxied object "full" icon...

Does "show the cube icon as a quick icon and then the full icon when loaded" sounds like acceptable to you ? We could also display the cube as a smaller cube on the bottom-left of that full icon like ISTR we do for some things ?

skurfer:
> I'm just not sure on what the current behavior is w.r.t to showing the "cube" proxy icon, showing the proxied object quick-icon, or showing the proxied object "full" icon…

From looking around a little, I think proxy object icons are just like any other object. The cube icon only appears for objects that explicitly define “Object” as their icon. It’s not a default for proxy objects or anything. For example, the Finder Selection and Path Finder Selection proxies both use the icon for the respective application.

> Does "show the cube icon as a quick icon and then the full icon when loaded" sounds like acceptable to you ? We could also display the cube as a smaller cube on the bottom-left of that full icon like ISTR we do for some things ?

Before looking, I would have said “maybe”, but now that I see the generic object cube icon has nothing to do with proxies (except those that have chosen to use it), I think we should respect the icon defined in the property list.

As for what to do when the name of the icon is “ProxyIcon”, I like the current behavior.

1. As soon as possible, steal the icon from the resolved object.
2. Watch the resolved object for changes to the icon. When they happen, update the proxy’s icon to match and notify the UI.

skurfer:
What do you think about changing `cache` to an `NSCache` instead of a dictionary as @pjrobertson suggested? I’m running it that way locally. No problems so far, and it should theoretically fix some of the icon loading crashes.

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -567,456 +554,498 @@ - (BOOL)childrenValid {
...
-			NSString *parentID = [self identifier];
-			for (QSObject *child in newAltChildren) {
-				[child setParentID:parentID];
-			}
-        }
-    } else {
-        [[self cache] removeObjectForKey:kQSObjectAltChildren];
-    }
+- (void)setIconLoaded:(BOOL)flag {
+	@synchronized (self) {
```
skurfer: I’m pretty sure this needs to lock on `[QSObject class]`.

1. The other changes to this set use that to lock.
2. The set doesn’t really belong to any on QSObject
3. The `hash` method also locks on `self`, which causes problems since it gets called a lot to test equality when adding to the set.

tiennou: This one locks on `self` because of `atomic` semantics. You're talking about the `@synchronized` just below though, so I'm taking a note.

Point 3 is the hard one. I have no idea how to fix that, because we're accessing potentially "shared" data so we need the lock, but locking kills hashing because we use the identifier. The few solutions I had were :

1) make hash independent for identifier again, but this will allow duplicates.
2) make identifier settable once-and-only-once, IOW assert if the id is already set when entering `setIdentifier:`.
3) do the object-freeze dance I think I proposed. IOW, object sources build objects. As soon as the librarian sees them (IIRC returned from `-scannedObject`) we mark them as frozen, which prevents some of the setters from working (like `-identifier/label/whateverMakesSense`.

n8henrie:
It's been 5 years and there are a handful of conflicts to resolve. Is this still worth pursuing @tiennou @skurfer?

n8henrie:
Closing as stale / unfinished, thank you @tiennou for all the effort put into this, happy to reopen if it is picked back up!

PR #2345, opened 2017-04-18
Don't trigger VDKQueue on the main thread

tiennou:
I pushed an updated version of VDKQueue to our fork (there was some movement in the Github Network).

Note that I don't think VDKQueue is responsible for our deadlocks 😉.


skurfer:
Nice. Running this in a test build…

pjrobertson:
Seems reasonable. To summarise the main VDKQueue changes:

* Code cleanup
* Switch to ARC
* Add the ability to specify the GCD queue for notifications (which Etienne's made use of here)

PR #2346, opened 2017-04-19
Deprecation cleanup

tiennou:
I closed #2257 in error it seems. Here's a rejig of it. I still need to fix the comments made there, but I would *really* like it to go in.


skurfer:
Code looks good. I’ve merged this into my local build so I can test.

tiennou:
I've added a few more easy ones.

- The catalog fixes are in catalog/clean-obj-source #2255.
- I don't think the `QSCatalogEntry` changes allows us to get rid of the NSNib warnings yet (need to keep the *topLevelObjects* array around)
- Most of the `LS` ones are "deprecated in 10.10, new in 10.10", so thanks Apple for the mess...
- I think the NSWorkspace ones could go as soon as we get the process management (which is in deprecation/breaking).
- Some of the alert ones are there too, but maybe we want to change some of those to notifications, so it's likely I'll post it separately.
- I seem to remember there was "riots" when trying to fix the `WebFrame*Delegate` ones, but I can't remember. Was that 10.9+ ?

PR #2347, opened 2017-04-19
Fix CA transactions warnings

tiennou:
This should fix the CA transaction warnings we have (at least the ones I saw).

I went pretty heavy-handed on the proxy thing, but I didn't want to make all `QSTask`'s accessors to become main-thready-asyncy...

I'll check that I didn't forget some of the keypaths though (I haven't checked everything, and I don't think we have an easy indeterminate progress one anywhere).

It seems to also fix the flicker we had, so 👍.

### DIFF
--- a/Quicksilver/Code-App/QSTaskViewController.m
+++ b/Quicksilver/Code-App/QSTaskViewController.m
@@ -47,7 +47,7 @@ const NSString *QSTaskProxyObservationContext = @"QSTaskProxyObservationContext"
 	}
 }
 
-- (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary<NSKeyValueChangeKey,id> *)change context:(void *)context {
+- (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary<NSString *,id> *)change context:(void *)context {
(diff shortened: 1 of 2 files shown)


tiennou:
Check completed. I manually switched `scanTask` and it didn't fail.

skurfer:
I assume we can close #2329?

tiennou:
Sorry, that one slipped my radar. I'm not sure actually, #2329 does (IMHO) weird things[1] to tasks, and the 2 other changes are unrelated, so they might be needed. My run was mostly "background", so I might not have got some of the warnings that we can get when eg. actions are executed.

[1] I don't see how changing to a sync block can fix anything. At worst, you might improve the deadlock rate because now whatever caused the task view to update (like a notification) will be waiting on the main thread.

skurfer:
I just did like it said and set `CA_DEBUG_TRANSACTION`, but I don’t remember exactly what I found or why I did things that way. It was a loooong time ago.

skurfer:
Looking back, we do need those changes, or something like them, no? Those things shouldn’t be allowed to happen on a background thread.

tiennou:
True, but if it's called on the main thread, it's the fault of whatever called them *not* on the main thread (that's how I consider everything in `QSInterface`). So IMHO those are not correct fixes per se.

PR #2348, opened 2017-04-20
Add FMDB as an external dependency

tiennou:
Because I got *really* tired of `objc[14949]: Class FMDatabase is implemented in both $1 and $2. One of the two will be used. Which one is undefined.` * @[Safari, Firefox, Chrome] * @[FMDatabase, FMResultSet, FMStatement] on each launch.

Note, I'm pretty sure it works, but just in case, you *should* try running QS, right-click "Show in Finder" the Products/fmdb.framework and delete that from the built product **as soon as QSFoundation is built**, before it launches (else Xcode will rebuild it). You'll either get a successful launch, or a `dyld` failure.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "Quicksilver/Code-External/LaunchAtLoginController"]
 	path = Quicksilver/Code-External/LaunchAtLoginController
 	url = https://github.com/quicksilver/LaunchAtLoginController.git
+[submodule "Quicksilver/Code-External/FMDB"]
+	path = Quicksilver/Code-External/FMDB
+	url = git://github.com/ccgus/fmdb.git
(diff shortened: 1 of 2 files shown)


skurfer:
So glad to see this. I was planning to do it one day.

Should we add JSONKit, too? I thought there was more than one plug-in using that.

tiennou:
I'm actually deleting it 😆. I don't think we need it, now that there's `NSJSONSerialization`.

tiennou:
Note there's a catch, the resulting framework is not *really* complete : it's missing a `Frameworks` -> `Versions/Current/Frameworks` symlink. I'm not sure if it matters or not. It doesn't seem to, at least, and I don't think it makes a difference in practice, because the rpath is always relative to the link target (ie. `Versions/A/QSFoundation`).

skurfer:
I don’t know if that symlink matters or not either, but I was able to build the Chrome plug-in, so probably not.

skurfer:
I finally tracked the code signing failures to this change. I’m reading the docs, but if you have any ideas @tiennou, let me know.

skurfer:
I’ve made progress on this. I noticed that before we sign the app, the FMDB framework already contains a `_CodeSignature`, even if I uncheck “Code Sign On Copy”. I went into the Xcode Project for the framework and changed its signing cert from “Ad Hoc” to “Developer ID Application” and it stopped signing the thing. What weird is that now I can’t change it back. It just says “Don’t Sign”, which is fine I guess.

So maybe this is something I should report upstream.

PR #2353, opened 2017-04-29
remove xctool

skurfer:
I’ve had it commented out in the build script for months. Time to make it official.

### DIFF
--- a/Quicksilver/Tools/qsrelease
+++ b/Quicksilver/Tools/qsrelease
@@ -1,14 +1,5 @@
 #!/bin/zsh
 
-## test
-if ! which xctool &> /dev/null
-then
-    echo "Tests require xctool (available via Homebrew)"
-    exit 78
-fi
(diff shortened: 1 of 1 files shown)


pjrobertson:
I can't remember why we used xctool... I this we chose to use it for Travis because it worked better for headless systems. Yeah I think that was it. See #2021. Is it OK that we're using two different testing tools for CI and for builds?

Why did you have it commented out?

skurfer:
> I can't remember why we used xctool...

I thought the reason was that the Quicksilver tests involving the GUI would never pass under `xcodebuild` (even locally).

> Is it OK that we're using two different testing tools for CI and for builds?

Actually, I wasn’t thinking about Travis, just the local release builds. I should probably update that file to use a newer Xcode and a different test command. Agree?

> Why did you have it commented out?

Because the default on my system is Xcode 8, so `xctool` would always die. I’d switch to Xcode 7 and re-enable it for actual releases.

PR #2362, opened 2017-05-31
Event changes for hotkey editor

skurfer:
I removed the problematic entries, and switched the others to use the newer names. It means we can’t build with Xcode 7 any more, but that ship already sailed with `NSWindowStyleMask`. I’ll add another commit to update Travis.

Take a quick look at this if you can @tiennou.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: objective-c
-osx_image: xcode7.3
+osx_image: xcode8.3
 xcode_project: Quicksilver/Quicksilver.xcodeproj # path to your xcodeproj folder
 xcode_scheme: "Quicksilver Distribution"
 script: |
-    ruby Quicksilver/Tools/travis/check_indent.rb &&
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: objective-c
-osx_image: xcode7.3
+osx_image: xcode8.3
 xcode_project: Quicksilver/Quicksilver.xcodeproj # path to your xcodeproj folder
 xcode_scheme: "Quicksilver Distribution"
 script: |
-    ruby Quicksilver/Tools/travis/check_indent.rb &&
-    xctool -project "$TRAVIS_XCODE_PROJECT" -scheme "$TRAVIS_XCODE_SCHEME" build test
-
+    ruby Quicksilver/Tools/travis/check_indent.rb
```
tiennou: Haven't you killed the `xctool` invocation here ?

tiennou:
LGTM !

.travis.yml
skurfer: Yes. It fails under Xcode 8 and they aren’t planning to fix it.

PR #2366, opened 2017-06-08
Touch Bar Support

skurfer:
A working demo for anyone that wants to try it out. To customize the touch bar, you’ll need to make the Dock icon visible, then choose “Customize Touch Bar” from the menu.

Available buttons:

  * History back/forward
  * Grab current selection
  * Clear the interface
  * Quick Look
  * Reveal Proxy
  * Show Task Viewer

And for dealing with collections:

  * Back/forward in collection
  * Explode collection
  * Remove Last Item

I don’t know if the touch bar will be the primary way to do any of these things. I mainly see it as a way to expose the fact that they’re possible. I’m guessing a lot of people don’t know about the functionality listed above, even though it’s all been here for years.

The main thing preventing this from being ready, in my opinion, is the icons. We need some better ones for things like grab selection.

I’d like to hear ideas for additional buttons.


Quicksilver/Code-App/QSController.m
```diff
@@ -870,7 +870,9 @@ - (void)startQuicksilver:(id)sender {
 	[QSLibrarian sharedInstance];
 	[QSExecutor sharedInstance];
 	[QSTaskController sharedInstance];
-	[NSApp setAutomaticCustomizeTouchBarMenuItemEnabled:YES];
+	if ([NSApplication isSierra]) {
```
pjrobertson: I checked if there was a way to do `if(mac_has_touchbar)` but apparently [Apple doesn't want you to do that](https://stackoverflow.com/a/40916781), so this seems fine to me

pjrobertson:
Cool, I'm still dubious as to the use of the touchbar, but as a way of exposing hidden features this is great. Unfortunately I have no way of testing it. Can you send a screenshot/picture of the icons?

My guess is they are 'adequate' for now. I've looked through the code and it all seems good to me.
I've made a couple of small comments directly on the code

Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
```diff
@@ -2066,4 +2073,115 @@ - (NSRect)previewPanel:(QLPreviewPanel *)panel sourceFrameOnScreenForPreviewItem
...
+		NSTouchBarItemIdentifierFlexibleSpace,
+	];
+//	touchBar.escapeKeyReplacementItemIdentifier = DismissQSItemIdentifier;
+	return touchBar;
+}
+
+- (nullable NSTouchBarItem *)touchBar:(NSTouchBar *)touchBar makeItemForIdentifier:(NSTouchBarItemIdentifier)identifier
+{
+	BOOL somethingIsSelected = ([self objectValue] != nil);
+	if ([identifier isEqualToString:QSHistoryItemIdentifier]) {
```
pjrobertson: In the Swift sample code (`ToolbarSample/WindowController.swift`) it uses a `switch identifier` instead of doing ifs. I think this is just a fancy part of Swift that means it can do switches on not just ints, but strings as well (?).

Quicksilver/Code-App/QSController.m
skurfer: I had it in there unconditionally at first because the documentation implies that the SDK will work it out for you. That was not the case when I tried (unsuccessfully) to run it under 10.10.

skurfer:
> Unfortunately I have no way of testing it. Can you send a screenshot/picture of the icons?

You can use [Touché](https://red-sweater.com/touche/) to check it out. There’s also a way to simulate in Xcode, but I never bothered to figure that out.

> The main thing preventing this from being ready, in my opinion, is the icons. We need some better ones for things like grab selection.

I’m going to talk to a designer I know about this to see what it would cost. (She’s family, so probably cheap.) Any objections?

pjrobertson:
> I’m going to talk to a designer I know about this to see what it would cost. (She’s family, so probably cheap.) Any objections?

If you think it’s something that will improve the usefulness of the touchbar feature, sounds good to me.

PR #2369, opened 2017-06-14
use string value as identifier for string objects

skurfer:
I nearly did this before, but decided not to based on the fear that if you were to type (or paste) an existing object’s identifier in text mode, the string object representing the identifier would replace the real object in the catalog. That doesn’t seem to be the case.

You do get some weirdness with history (you can’t store both an object and a string with its identifier), but that’s an edge case I don’t think we need to worry about.


skurfer:
Already found one huge problem with this: It completely breaks ⌘A and ⌘⎋ for multiple files.

skurfer:
This no longer causes the previously mentioned problems. I suspect it has to do with recent changes to the way collections are displayed. No hurry to merge. I’d like to run it for a while to be sure.

PR #2387, opened 2017-08-01
refactor to avoid confusing Xcode

skurfer:
The symbols menu for this file was getting cut off.

I did this back in May, but apparently never pushed it.

### DIFF
--- a/Quicksilver/Code-App/QSUpdateController.m
+++ b/Quicksilver/Code-App/QSUpdateController.m
@@ -37,10 +37,11 @@
 	NSUserDefaults* defaults = [NSUserDefaults standardUserDefaults];
 
 #ifdef DEBUG
-	if (![defaults boolForKey:@"QSPreventAutomaticUpdate"]) {
+	NSString *checkKey = @"QSPreventAutomaticUpdate";
 #else
-    if ([defaults boolForKey:kCheckForUpdates]) {
+	NSString *checkKey = kCheckForUpdates;
 #endif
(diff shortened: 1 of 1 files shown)


Quicksilver/Code-App/QSUpdateController.m
```diff
@@ -37,10 +37,11 @@ - (void)setUpdateTimer {
 	NSUserDefaults* defaults = [NSUserDefaults standardUserDefaults];
 
 #ifdef DEBUG
-	if (![defaults boolForKey:@"QSPreventAutomaticUpdate"]) {
```
tiennou: There's a `!` 😉 .

Alternatives would be to switch to an early return and reindent, or stash the test in a `BOOL` somewhere (but I'm somewhat concerned that there's differing behavior between `DEBUG`/`!DEBUG`.

``` obj-c
#ifdef DEBUG
    if ([defaults boolForKey:@"QSPreventAutomaticUpdate"])
        return;
#endif
    if (![defaults boolForKey:kCheckForUpdates])
        return;
```

skurfer: > There's a `!` 😉 .

Well, that seems somewhat important. 🤦🏻‍♂️

I’ll close this and do something else as you suggested (unless you want to just add it to the other update controller fixes).

tiennou: I expect you will get around to that first. I need to split the other branch between the alert stuff (which we want) and the background updating (which doesn't work because `NSURLDownload`).

PR #2395, opened 2017-09-19
Transliterate non-latin strings into latin if there's no match

pjrobertson:
I'm back! Although for my own selfish needs :'(

Most non-latin languages have formal romanizations from the native script to latin (e.g. 中国 -> zhongguo). This change allows you to search using latin characters for non-latin names/labels. It should apply to all latin languages that have formal romanizations.

One issue that I had with this is that when you romanize a non-latin search string (e.g. 中国) its length is changed (often increased):

```
中国 => length 2
zhongguo => length 8
```

This would cause problems when displaying the characters that matched in the string (normally done by underlining them in the SOV. To solve this problem I just opted to not show which letters matched:

screen shot 2017-09-19 at 10 05 32

Note that this change will have an affect on all users, and my suspicion is most of our users are limited to latin-character only languages. As such I'd be interested to:

a) know if this makes searching slower for people
b) if so, should it just be included as a hidden pref?

### DIFF
--- a/Quicksilver/Code-QuickStepFoundation/QSSense.h
+++ b/Quicksilver/Code-QuickStepFoundation/QSSense.h
@@ -10,4 +10,5 @@
 
 
 CGFloat QSScoreForAbbreviation(CFStringRef string, CFStringRef abbr, id hitMask);
+CGFloat QSScoreForAbbreviationOrTransliteration(CFStringRef str, CFStringRef abbr, id mask);
 CGFloat QSScoreForAbbreviationWithRanges(CFStringRef str, CFStringRef abbr, id mask, CFRange strRange, CFRange abbrRange);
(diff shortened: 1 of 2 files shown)


pjrobertson:
To read up on what I've done here, NSHipster is always here to help: http://nshipster.com/cfstringtransform/

skurfer:
> I'm back! Although for my own selfish needs :’(

But isn’t that the source of all pull requests?

I’ll give this a look.

pjrobertson:
Great, thanks! I was just jumping on here to ask if I could merge :)

skurfer:
I’ve been using this for a while and haven’t noticed any performance problems. I’ll be merging this and some of the others for a new release soon. (Hoping to hear something on the exceptions from the updater first.)

pjrobertson:
OK great,

Is Etienne looking over the updater exceptions? Can you point me to the specific PR if it needs another look over?

skurfer:
We’re discussing it in #2402. There isn’t really a clean solution.

skurfer:
Would you mind adding this one to the release notes @pjrobertson? Not sure I would summarize it quite right.

pjrobertson:
Done

PR #1596, opened 2013-09-10
Bash QSUpdateController out of the main thread.

tiennou:
Since I'm currently accessing the 'ternet from my world-changing iPhone, I just noticed that the update system does its thing on the main thread. When there's a DNS hiccup at the same time, you end up with a spinning-beachball QS for _too long_.

This fixes it by generous use of dispatch queues. This also makes its API more self-contained.

Not completely tested, by which I mean I clicked on the "Check Now" button, but that's all I did.

As a bonus, every user-facing string is now wrapped in a gentle `NSLocalizedString`.

PR #2402, opened 2017-10-16
Update controller bug fixes

skurfer:
This is a follow up to #1596.

I fixed the bugs I ran into while testing it out. I wanted to let @tiennou look over these changes before going to `master`.

On that last commit, yes we’ve got a strong reference to the update controller in the block, but

1. It’s a singleton, so keeping a reference to it shouldn’t be a problem.
2. I think we *want* it to stay around for the life of the block, so it can do what the user asks if needed.


skurfer:
Bah! I forgot about the [autolayout exceptions][1] being thrown. We should probably do something about that before release.

[1]: https://github.com/quicksilver/Quicksilver/pull/1596#issuecomment-319370818

skurfer:
The exceptions have something to do with the `NSAlert` that gets passed to `QSAlertManager`. They’re being created on the update controller’s background thread, but then the alert manager’s main thread block sends messages to it, then it continues to exist in the background thread’s scope. (Maybe the error happens when it’s freed?)

We don’t need it to exist after the alert manager gets it, but since it isn’t declared locally with `__block`, we can’t do anything to it there.

Do you think we need to refactor the alert manager so it creates the `NSAlert` in the same block where we use it instead of allowing them to be created in other scopes? Other ideas?

tiennou:
I kinda like the API I've come up with on `QSAlertManager` (though I'm open to suggestions), namely the `completionHandler` part.

One hacky-workaround I can think of is copying the alert in `+[QSAlertManager beginAlert:…]` before doing the `beginSheet…` calls, so the *actual* alert we display is owned by the main-thread. Sadly `NSAlert` isn't `-copy`-able, so it would need to be done manually…

Quicksilver/Code-App/QSUpdateController.m
```diff
@@ -216,18 +215,13 @@ - (void)checkForUpdates:(BOOL)userInitiated {
 
 				[[QSAlertManager defaultManager] beginAlert:alert onWindow:nullEvent completionHandler:^(QSAlertResponse response) {
 					if (response == QSAlertResponseOK)
-						shouldInstallApp = YES;
-					else if (response == QSAlertResponseCancel)
-						shouldInstallApp = NO;
+						[self installAppUpdate];
```
tiennou: Heh, you managed to completely sidestep the queuing problem I [had](https://github.com/quicksilver/Quicksilver/pull/1596#issuecomment-174290059) with `QSURLDownload` by doing that.

Note that it will move back most of the installation process back under the main thread, which will cause it to lock again, but between beachballs and crashes and rewriting `QSURLDownload`, I'm not sure what's best…

tiennou:
I've pushed some parts I was working on [here](https://github.com/quicksilver/Quicksilver/tree/t/updater-fixes). Of relevance are the first few commits, and the last, which has the workaround discussed above, feel free to drop the WIP one for WIPedness (but that's the start of the changes needed to `QSURLDownload`).

I've tested without the WIP commit, and the download started, and cancelling it worked — that's the extent I've done. HTH !

skurfer:
Thanks, I’ll look at those. Feel free to push things straight to this branch if they’re related.

tiennou:
As a side note, my comment on QSURLDownload above is wrong : what was beachballing were the synchronous calls the updater does when checking its status, and the network is flaky. The actual download will be done on some of NSURLDownload's queues.

skurfer:
Copying the alert seems to work. I had actually tried that using `copy`, but that obviously isn’t an option. I didn’t think of doing it the hard way.

I’m a little surprised that it works. You’re still sending messages to the original alert on the main thread to get its properties. Maybe reading properties doesn’t count as “accessing the autolayout engine”?

I’ll look at the other commits later.

tiennou:
I think that's the gist of it, yes. After the `-begin…` calls, the NSAlert is bumped around by the layout engine, which is what makes it fails. What completely baffled me was that the layout code actually runs in a queue when the exception happens, not on the main thread. So it's complaining about a non-main-thread-owned NSAlert while running in a non-main thread 😜 .

pjrobertson:
Sorry, I haven't been keeping up with all the happenings here. But one
quick question: do we have to use an NSAlert? Could a notification be used
instead? I have a feeling Rob's battled with them before and using them for
making choices isn't great, but I thought I' suggest it just in case.

skurfer:
> Could a notification be used instead? I have a feeling Rob's battled with them before and using them for making choices isn't great, but I thought I' suggest it just in case.

They *can* be suitable for offering choices, if the user has things configured for Alerts instead of Banners, but that’s unlikely since it would make **every** notification from QS require manual dismissal.

pjrobertson:
Aha, I thought there was some kind of horrible usability problem with notifications. Damn.

Could we at least replace the NSAlerts that don't need to be NSAlerts with notifications? E.g. we could replace the "You're up-to-date!" one with a notification. If there are any other ones that don't require user input then they should also be notifications.

I just checked and that'd get rid of 2 alerts: the 'You're up-to-date!' one and the 'Internet Connection Error' one.

Thoughts?

skurfer:
I’ve added all of the commits here except the WIP. Planing to release unless someone objects soonish. 😃

Issue #2396, opened 2017-09-26
Recent Documents are broken under High Sierra


skurfer:
Once again, we’ll have to figure out how to get this working.

The Console just shows `<private>`. I’m *sure* this is just Apple’s way of saying “use the public API instead”, right?

skurfer:
Of course the Xcode console has a bit more to say:

```
2017-09-26 16:19:18.972723-0400 Quicksilver[35769:1443206] [General] *** -[NSKeyedUnarchiver decodeObjectForKey:]: cannot decode object of class (SFLListItem) for key (NS.objects); the class may be defined in source code or a library that is not linked
2017-09-26 16:19:18.974290-0400 Quicksilver[35769:1443206] [General] (
	0   CoreFoundation                      0x00007fff328f20fb __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00007fff591dec76 objc_exception_throw + 48
	2   Foundation                          0x00007fff34a1b23f -[NSCoder(Exceptions) __failWithExceptionName:errorCode:format:] + 0
	3   Foundation                          0x00007fff34a1b3bc -[NSCoder(Exceptions) __failWithExceptionName:errorCode:format:] + 381
	4   Foundation                          0x00007fff34906474 _decodeObjectBinary + 2856
	5   Foundation                          0x00007fff34907510 -[NSKeyedUnarchiver _decodeArrayOfObjectsForKey:] + 1779
	6   Foundation                          0x00007fff3490b367 -[NSArray(NSArray) initWithCoder:] + 202
	7   Foundation                          0x00007fff349062c1 _decodeObjectBinary + 2421
	8   Foundation                          0x00007fff34907510 -[NSKeyedUnarchiver _decodeArrayOfObjectsForKey:] + 1779
	9   Foundation                          0x00007fff34906934 -[NSDictionary(NSDictionary) initWithCoder:] + 202
	10  Foundation                          0x00007fff349062c1 _decodeObjectBinary + 2421
	11  Foundation                          0x00007fff34905848 _decodeObject + 243
	12  Foundation                          0x00007fff349b0144 +[NSKeyedUnarchiver unarchiveObjectWithFile:] + 148
	13  QSCore                              0x00000001004ac037 recentDocumentsForBundle + 423
	14  QSCore                              0x00000001004b04be -[QSFileSystemObjectHandler loadChildrenForObject:] + 2894
[...]

PR #2403, opened 2017-10-16
Support for recent documents in High Sierra

skurfer:
…and various recent/favorite catalog presets.

Doing a string replacement on the extension feels ugly, but I don’t think we have a way to restrict presets by OS version, and that’s where the path is defined.

As discussed in #2396


pjrobertson:
Huh, that was simple! So it's basically exactly the same just it's `.sfl2` files that are NSDictionaries when unarchived. Nice detective work! Could well break again, but looks good to me.

PR #2404, opened 2017-10-16
redefine Relaunch as a trigger

skurfer:
As promised.

* enabled by default
* scoped to be active only in Quicksilver
* uses ⌃⌥⌘Q by default

Since it’s a trigger, users on 10.12 or older can change it back or change it to something else if they want.

### DIFF
--- a/Quicksilver/Code-App/QSController.h
+++ b/Quicksilver/Code-App/QSController.h
@@ -67,6 +67,7 @@
 - (NSString *)crashReportPath;
 - (void)showDockIcon;
 - (void)clearHistory;
+- (void)relaunchQuicksilver;
 
 @end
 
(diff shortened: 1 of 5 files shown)


pjrobertson:
Nice, looks good. We need to make it pretty clear for existing users since this will be a break in workflow. How bad would it be to make the trigger default ⌘⌃Q?

skurfer:
> How bad would it be to make the trigger default ⌘⌃Q?

Well, the new one has the advantage of working on all OS versions. It also makes it obvious that something has changed. People might assume it’s “broken” forever otherwise.

pjrobertson:
Sounds reasonable. I'm sure you'll make it clear in the changelog anyway ;-)

Issue #767, opened 2012-03-25
QSWebSource should be available in the Catalog UI


skurfer:
This would give users a way to supply a URL and have all the links on that page added to their catalog. All the plumbing is there, it just needs to be added to the drop-down for new entries (and it needs a NIB for configuring it).

PR #2408, opened 2017-11-23
URL Object Source

skurfer:
As discussed way back in #767, this adds some UI that will let you scan file contents from a remote URL.

One thing I can’t figure out is how to get the URL to “stick” when the text box loses focus. (I hate dealing with NIBs.) But other than that, it works.

There are some small bonus fixes at the end.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -981,7 +981,9 @@ NSSize QSMaxIconSize;
 		[icon setCacheMode:NSImageCacheNever];
         if (iconChange) {
             // icon is being replaced, not set - notify UI
-            [[NSNotificationCenter defaultCenter] postNotificationName:QSObjectIconModified object:self];
+			QSGCDMainAsync(^{
+            	[[NSNotificationCenter defaultCenter] postNotificationName:QSObjectIconModified object:self];
+			});
         }
 	}
 }
--- a/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.h
@@ -2,6 +2,8 @@
 #import <QSCore/QSClangAnalyzer.h>
 #import "QSObject.h"
 
+NSArray *recentDocumentsForBundle(NSString *bundleIdentifier);
+
 @protocol QSFileObjectCreationProtocol
 - (id)createFileObject:(QSObject *)object ofType:(NSString *)type;
 @end
--- a/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.h
+++ b/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.h
@@ -2,7 +2,7 @@
 #import <QSCore/QSClangAnalyzer.h>
 #import "QSObject.h"
 
-NSArray *recentDocumentsForBundle(NSString *bundleIdentifier);
+NSArray *QSGetRecentDocumentsForBundle(NSString *bundleIdentifier);
 
 @protocol QSFileObjectCreationProtocol
 - (id)createFileObject:(QSObject *)object ofType:(NSString *)type;
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
@@ -628,9 +628,7 @@ NSMutableDictionary *bindingsDict = nil;
 
 - (void)objectIconModified:(NSNotification *)notif {
     // icon changed - update it in the pane
-	QSGCDMainAsync(^{
-		[self setNeedsDisplay:YES];
-	});
+	[self setNeedsDisplay:YES];
 }
 
 #pragma mark -
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
@@ -628,7 +628,9 @@ NSMutableDictionary *bindingsDict = nil;
 
 - (void)objectIconModified:(NSNotification *)notif {
     // icon changed - update it in the pane
-    [self setNeedsDisplay:YES];
+	QSGCDMainAsync(^{
+		[self setNeedsDisplay:YES];
+	});
 }
 
 #pragma mark -
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
@@ -45,7 +45,7 @@
 - (NSView *)settingsView
 {
 	if (![super settingsView]) {
-		[NSBundle loadNibNamed:NSStringFromClass([self class]) owner:self];
+		[[NSBundle bundleForClass:[self class]] loadNibNamed:NSStringFromClass([self class]) owner:self topLevelObjects:NULL];
 	}
 	return [super settingsView];
 }
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
@@ -37,4 +37,65 @@
 	return YES;
 }
 
+- (BOOL)isVisibleSource
+{
+	return YES;
+}
+
+- (NSView *)settingsView
+{
+	if (![super settingsView]) {
+		[NSBundle loadNibNamed:NSStringFromClass([self class]) owner:self];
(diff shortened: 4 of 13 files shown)


Quicksilver/Code-QuickStepInterface/QSSearchObjectView.m
```diff
@@ -628,7 +628,9 @@ - (void)selectObject:(QSBasicObject *)obj {
 
 - (void)objectIconModified:(NSNotification *)notif {
     // icon changed - update it in the pane
-    [self setNeedsDisplay:YES];
+	QSGCDMainAsync(^{
```
tiennou: Technically, I'd prefer whatever sent the notification to be where that main-async call is added, but yeah 😉.

Quicksilver/Code-QuickStepCore/QSObject_FileHandling.h
```diff
@@ -2,6 +2,8 @@
 #import <QSCore/QSClangAnalyzer.h>
 #import "QSObject.h"
 
+NSArray *recentDocumentsForBundle(NSString *bundleIdentifier);
```
tiennou: Namespacing pretty please ?

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
```diff
@@ -38,4 +37,65 @@ - (BOOL)indexIsValidFromDate:(NSDate *)indexDate forEntry:(NSDictionary *)theEnt
...
+{
+	return YES;
+}
+
+- (NSView *)settingsView
+{
+	if (![super settingsView]) {
+		[NSBundle loadNibNamed:NSStringFromClass([self class]) owner:self];
+	}
+	return [super settingsView];
```
tiennou: This made me wonder, but #2255 is still unmerged. Oh well.

Quicksilver/Code-QuickStepCore/QSObject_FileHandling.h
skurfer: You mean rename the function to something more unique?

skurfer:
I’ve moved the download status strings to another PR.

I’ve also changed the icon stuff, but I’d still like to talk about it…

When you post the notification, you don’t know who might pick it up and what thread they need to do the work on, so doesn’t it make more sense to let the _observer_ handle that? Plus, for the sake of argument, let’s say `objectIconModified:` was being called from somewhere other than the notification. It needs to run on the main thread in that case too, but now there’s no assurance that it will.

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
```diff
@@ -38,4 +37,65 @@ - (BOOL)indexIsValidFromDate:(NSDate *)indexDate forEntry:(NSDictionary *)theEnt
...
+	[itemParserPopUp setMenu:parserMenu];
+	NSInteger parserEntry = [itemParserPopUp indexOfItemWithRepresentedObject:parser];
+	[itemParserPopUp selectItemAtIndex:(parserEntry == -1?0:parserEntry)];
+}
+
+- (IBAction)setValueForSender:(id)sender {
+	NSMutableDictionary *settings = self.selectedEntry.sourceSettings;
+	if (!settings) {
+		settings = [NSMutableDictionary dictionaryWithCapacity:1];
+		[[self currentEntry] setObject:settings forKey:kItemSettings];
```
skurfer: Speaking of updated classes, I get a warning here about `currentEntry` of course, but I don’t see an alternative. `QSCatalogEntry` doesn’t initialize `self.info[kItemSettings]`, and it doesn’t give us a way to do it either. Should we add a setter method to that class, or just do it in `init`?

Quicksilver/Code-QuickStepCore/QSObject_FileHandling.h
tiennou: Yeah, just to be on the safe side.

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
tiennou: I ought to be confused by #2255 not being in and having a bunch of expectations about *how* it should behave being wrong because it's not yet in. Hence, I don't really care if what you're doing is hacky, as the API is a mess. If this goes in before #2255, I'll just fix that code in the other PR (I will try to work on that soon™).

tiennou:
> When you post the notification, you don’t know who might pick it up and what thread they need to do the work on.

Given the amount of bugs we had because of the "originating" thread being not obvious, I'm really tempted to explicitly make that API main-thread *by default*, instead of the hodge-podge it is now. I mean, *everyone* who choses to respond to that notification is very likely to 1) be trying to update its GUI in some way in response to the icon being loaded 2) cause a main-thread violation because it's not obvious you've just been pinged *on* the background loader thread/whatever else sent it. Hence the stance I have now w.r.t to this : "every posting of that notification MUST be made on the main thread".

> Plus, for the sake of argument, let’s say `objectIconModified:` was being called from somewhere[…]

As per the rule above, I'd be tempted to say that it's "somewhere's" responsibility to make sure GUI calls are made on the main-thread. You could simplify out the whole notification out, and "somewhere", calling `-setNeedsDisplay:` from a background thread would still be incorrect without a main-thread dispatch.

Maybe I should have opened an issue to clarify, but that was why I started #2255 in the first place 😉 .

skurfer:
I’ve rebased this to take advantage of stuff in #2255 and renamed the “recent documents” function. Give it a look when you can.

On the icon notification…

> _everyone_ who choses to respond to that notification is very likely to 1) be trying to update its GUI in some way

Yes, in this case we know all the places it gets used and what it’s doing so it’s not the best example. But in general, you won’t always know.

> > Plus, for the sake of argument, let’s say `objectIconModified:` was being called from somewhere[…]
> 
> As per the rule above, I'd be tempted to say that it's "somewhere's" responsibility to make sure GUI calls are made on the main-thread.

But what if the GUI call is 4 methods deep into what you’re calling and you missed it because you didn’t follow every possible code path? What if the GUI call is conditional and only hit 2% of the time a method gets called?

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSWebSource.m
tiennou: I was doubly wrong then, I meant #2256 🤣.

Quicksilver/Code-QuickStepCore/QSObject.m
```diff
@@ -981,7 +981,9 @@ - (void)setIcon:(NSImage *)newIcon {
 		[icon setCacheMode:NSImageCacheNever];
         if (iconChange) {
             // icon is being replaced, not set - notify UI
-            [[NSNotificationCenter defaultCenter] postNotificationName:QSObjectIconModified object:self];
```
tiennou: nitpick: weird indent

tiennou:
Apart from a nitpick, code looks good, haven't tested though (I'd be interested in your use-case 😉).

About that GUI-call discussion, I don't think you can convince me 😉. In the specific case of icons, I'd really prefer the notification be sent from the main thread because that's the most logical thing to do. I can think of a few places that use Cocoa bindings to listen to icon changes *cough*Catalog > Scanned objects*cough*never worked*cough*, which makes the fix above incomplete. If something ever wants that notification for non-GUI stuff, it's up to it to not lockup the main thread or async it somewhere else, but I'd expect this to be exceptional.

> But what if the GUI call is 4 methods deep into what you’re calling and you missed it because you didn’t follow every possible code path? What if the GUI call is conditional and only hit 2% of the time a method gets called?

Just so we're clear, I consider fishy anything that call `-someNotificationObserver:(NSNotification *)` methods directly, so I'll usually take a closer look, extract a common method and ensure the observer behaves correctly w.r.t queues (or whatever). Granted, I might miss something, as it's true "in general you won't always know", but I think those few rules help to make that true.

skurfer:
> (I'd be interested in your use-case 😉).

Partly it was just a shame that all the pieces were there, but not hooked up. You could use this to grab bookmarks from a remote site, index [frequently used documentation](https://docs.python.org/3/library/index.html), etc.

One use for me personally: We have dozens of [Trac](https://trac.edgewall.org) instances at work. I have script that generates an index of 3 links for each one (the main page, my tickets, and a search URL). Right now, I have to copy this file to my local machine periodically so it can get indexed by QS.

> Just so we're clear, I consider fishy anything that call `-someNotificationObserver:(NSNotification *)` methods directly

OK, true. But even if such a method is getting called in response to a notification and not directly, it could have a UI update lurking 5 calls deep, or in a rarely triggered `else` clause.

PR #2420, opened 2018-01-10
hide the 3rd pane when the interface is empty

### LINKED ISSUE #2417: Only show two panes if nothing is selected
Frequently, when invoking Quicksilver with nothing selected, all three panes are showing. It looks weird.

Related: Clearing the first pane with ⌫ or ⌃U should hide the third pane.

skurfer:
Also removes a few complaints from the console.

fixes #2417


pjrobertson:
I've been seeing more of the 3rd pane lingering recently. You read my mind!

PR #2423, opened 2018-02-05
Action config

skurfer:
Allow more things to be defined for actions. Basically because I needed some of this for the new Audio plug-in.

I could have just created 12 actions by hand in less time than it took to figure out the block stuff, but I think it’s worth it. For an actual use-case, see https://github.com/quicksilver/Audio-qsplugin/blob/master/Audio/AudioAction.m#L31

Another place it can be used is Remote Hosts where many of the actions are just opening a URL with a different scheme.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSAction.h
+++ b/Quicksilver/Code-QuickStepCore/QSAction.h
@@ -13,6 +13,7 @@
 + (id)actionWithDictionary:(NSDictionary *)dict;
 + (id)actionWithDictionary:(NSDictionary *)dict identifier:(NSString *)ident;
 + (id)actionWithIdentifier:(NSString *)newIdentifier;
++ (id)actionWithDictionary:(NSDictionary *)dict identifier:(NSString *)ident bundle:(NSBundle *)bundle;
 + (id)actionWithIdentifier:(NSString *)newIdentifier bundle:(NSBundle *)newBundle;
 
 - (id)initWithDictionary:(NSDictionary *)dict identifier:(NSString *)ident bundle:(NSBundle *)bundle;
--- a/Quicksilver/Code-QuickStepCore/QSAction.h
+++ b/Quicksilver/Code-QuickStepCore/QSAction.h
@@ -33,6 +33,7 @@
 - (NSInteger)rank;
 - (void)setRank:(NSInteger)newRank;
 - (CGFloat)precedence;
+- (void)setPrecedence:(CGFloat)precedence;
 - (BOOL)enabled;
 - (void)setEnabled:(BOOL)flag;
 - (BOOL)menuEnabled;
@@ -46,6 +47,8 @@
 - (BOOL)canThread;
 - (BOOL)indirectOptional;
 - (void)setIndirectOptional:(BOOL)flag;
+- (BOOL)validatesObjects;
+- (void)setValidatesObjects:(BOOL)flag;
 
 // resolveProxy is a BOOL set in an action's dict to specify whether an object should be resolved
 // before being sent to an action. Action's like 'assign abbreviation...' should not resolve the proxy
@@ -67,6 +70,7 @@
 - (id)objectForKey:(NSString*)key;
 
 - (NSString *)commandFormat;
+- (void)setCommandFormat:(NSString *)commandFormat;
 @end
 
 @interface QSActionHandler : NSObject
--- a/Quicksilver/Code-QuickStepCore/QSAction.h
+++ b/Quicksilver/Code-QuickStepCore/QSAction.h
@@ -30,6 +30,38 @@
 - (SEL)action;
 - (void)setAction:(SEL)newAction;
 
+/**
+ * Use a predefined block as an action method. Useful for programatically creating many similar actions.
+ *
+ * @param actionBlock A block that does the action's work
+ * @param selectorName A selector name like "openThing:"
+ *
+ * @return YES if the action method was added to the provider, otherwise NO
+ *
+ * Example action block with no indirect object:
+ * QSObject *(^actionBlock)(id, QSObject *) = ^ QSObject *(id _self, QSObject *dObject) {
+ * 	   // action code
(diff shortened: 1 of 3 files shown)


Quicksilver/Code-QuickStepCore/QSAction.m
```diff
@@ -192,6 +199,32 @@ - (void)setAction:(SEL)newAction {
...
+- (BOOL)setActionUisngBlock:(QSObject *(^)(id, QSObject *))actionBlock selectorName:(NSString *)selName
+{
+	if (![self provider]) {
+		NSLog(@"define provider before setting a block as the action");
+		return NO;
+	}
+	IMP actionFunction = imp_implementationWithBlock(actionBlock);
+	SEL actionSelector = NSSelectorFromString(selName);
+	BOOL actionDefined = class_addMethod([[self provider] class], actionSelector, actionFunction, "@@:@");
+	[self setAction:actionSelector];
```
tiennou: Maybe logging the failure and returning `NO` if the method couldn't be added would be more fool-proof 😉.

Quicksilver/Code-QuickStepCore/QSAction.m
```diff
@@ -192,6 +199,32 @@ - (void)setAction:(SEL)newAction {
 		[[self actionDict] removeObjectForKey:kActionSelector];
 }
 
+- (BOOL)setActionUisngBlock:(QSObject *(^)(id, QSObject *))actionBlock selectorName:(NSString *)selName
+{
+	if (![self provider]) {
+		NSLog(@"define provider before setting a block as the action");
```
tiennou: I'd be tempted to NSAssert that, but maybe I'm being too paranoid…

skurfer:
> I'm just a little concerned about generating actions programmatically — I can't remember seeing it done anywhere

Services Menu Plugin.

I’ll address the other stuff later. Thanks!

Quicksilver/Code-QuickStepCore/QSAction.m
skurfer: I’ve made this a little better.

Quicksilver/Code-QuickStepCore/QSAction.m
skurfer: It’s mostly there so a developer will catch it during the plug-in’s creation. Wouldn’t an assert crash QS if the mythical developer doesn’t fix things, potentially punishing end users?

Quicksilver/Code-QuickStepCore/QSAction.h
```diff
@@ -30,9 +30,42 @@
...
+ *
+ * @return YES if the action method was added to the provider, otherwise NO
+ *
+ * Example action block with no indirect object:
+ * QSObject *(^actionBlock)(id, QSObject *) = ^ QSObject *(id _self, QSObject *dObject) {
+ * 	   // action code
+ * 	   return nil;
+ * };
+ */
+- (BOOL)setActionUisngBlock:(QSObject *(^)(id, QSObject *))actionBlock  selectorName:(NSString *)selName;
```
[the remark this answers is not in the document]
skurfer: Yes. Because strings are “normal” and easy to create, while selectors are “weird” and difficult to create. “Difficult” meaning I had to look it up. I didn’t say it was a good reason. 😄 I was just trying to make it as simple as possible on the calling end.

Are you saying it should keep the string but also require you to create and pass in a `SEL`? Or drop the string?

Quicksilver/Code-QuickStepCore/QSAction.m
tiennou: Well, if the mythical lets a crashing assert come up in the wild, isn't the mythical to blame for its sloppiness 😉 ?

NSLogs can and will be ignored by the developer, and by the user. Exceptions will not. I'd say document that "It is a programming error to use this method if the action doesn't have a provider".

Quicksilver/Code-QuickStepCore/QSAction.h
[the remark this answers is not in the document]
tiennou: I'd use `-setActionUsingBlock:(…)block selector:(SEL)aSelector`.

Take for example `-mySelector:`. If you mistype `"mySelectr:"` you will get a "missing method" at runtime. If you use `@selector(mySelectr:)`, you willl/should get a compiler warning that it doesn't know about that particular selector.

Quicksilver/Code-QuickStepCore/QSAction.m
skurfer: > Well, if the mythical lets a crashing assert come up in the wild, isn't the mythical to blame for its sloppiness 😉 ?

Of course, but it’s the user that’ll feel the pain. Not the developer. The creators of all these plug-ins aren’t exactly jumping on issues they day they get reported. Or even the _year_ they get reported.

> NSLogs can and will be ignored by the developer, and by the user.

I don’t expect the log message to be what gets their attention. But when they realize their new action doesn’t exist, the message will explain why.

If we’re assuming they’re going to fix the issue during development one way or another, I suppose an exception gets the job done. I’ll change it.

Quicksilver/Code-QuickStepCore/QSAction.h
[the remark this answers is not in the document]
skurfer: OK, but the only time you ever type the name is during creation. It can be misspelled and still work. Also, I don’t think the compiler is aware of these methods even if you spell them correctly, so you’ll always get a warning. (I haven’t tested that today, but I think I was seeing such warnings during early attempts.)

skurfer:
And it did occur to me after I added all those setters that I could have just used `+[QSAction actionWithDictionary:]`. 🤦‍♂️ I’ll probably update the plug-in to do that. But the block stuff was necessary. (It ended up how it is now because I started by copying what the Services plug-in does.)

skurfer:
I think I addressed everything you asked about @tiennou. Anything else? I’ll wait 72 hours for objections, then merge.

PR #2424, opened 2018-02-07
add an action for dumping object details

skurfer:
Disabled by default.

Should have done this years ago. Yes, there’s an item like this in the menu for debug builds, but what kind of animal has the Dock icon enabled? And what if you need to see something currently in memory and aren’t running a debug build? What if you need info from a user on another system?


pjrobertson:
Makes sense, we don't really ever use debug builds so having this in the main build will be handy for debugging

PR #2426, opened 2018-02-15
Don't store IDs and archives for a QSCommand (in the plist)

### LINKED ISSUE #2419: Some `.txt` files not recognized as text files
QS is showing some `.txt` files as an unknown type. [Called 'bad' files below.]

Some `.txt` files are recognized as text files. [Called 'good' files below.]

A screenshot of how this appears is attached below.

I cannot figure out the pattern for when a file is good or bad. I have many `.txt` files in the same directory. All the files have the same permissions (644). All the `.txt` files are recognized by Quick Look as text files. Some are good files, some are bad. Some have extended attributes and some do not, but this does not match up with being good or bad. 

When I tried to use a trigger which appended to this file in the very recent past, I sometimes got complete nothingness (no error, no appended text), twice got an error that text could not be added to a (null) file, and once got an error that text could not be added to a text file. [These are approximations to the error messages, as the actual error messages come up only sporadically.] 

After a nothingness event, the Console had the message
`default	13:59:39.204629 -0600	Quicksilver	unexpected object for file path: (null), object: QSObject <0x6040004d3400>, /Users/<somePathHere>/TextScraps/Ideas.txt`

I'm using macOS 10.13.2 and just upgraded to QS 1.5.7 when looking for help on this topic. The problem existed in QS 1.5.6. I do not know if it existed in earlier versions, but I know it did not exist back in 2013, because of how I used to use triggers. I have FileVault turned on and my filesystem in now APFS. 

Any tips would be much appreciated.

qs_txt_oops

pjrobertson:
Fixes issue #2419

This *should* fix #2419 (I think), it seems that in this case:

1. You create a trigger for an dObject that's not in the catalog
2. That dObject subsequently gets added to the catalog
3. You resave the trigger (I guess it happens on every quit of the app)

Then the trigger would have both a directArchive (created originally when the object isn't in the catalog) and then subsequently a directID (when the object gets added to the catalog). This causes problems in `QSCommand.m:-(QSObject *)dObject` (I think).

Now the ideal solution would be to store *both* the directID and directArchive, and (for the event that you add a trigger for something that's in the catalog, then it subsequently gets removed from the catalog) but that's another task for another day.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSCommand.m
+++ b/Quicksilver/Code-QuickStepCore/QSCommand.m
@@ -295,49 +295,40 @@ NSTimeInterval QSTimeIntervalForString(NSString *intervalString) {
 	[[NSDictionary dictionaryWithObject:[self dictionaryRepresentation] forKey:@"command"] writeToFile:path atomically:NO];
 }
 
+- (void)storeObject:(QSObject *)newObject forType:(NSString*)type {
+    
+    NSString *idString = [NSString stringWithFormat:@"%@ID", type];
+    NSString *archiveString = [NSString stringWithFormat:@"%@Archive", type];
+    
+    id rep = [newObject identifier];
+	if(rep != nil) {
+        [[self commandDict] setObject:rep forKey:idString];
+        [[self commandDict] removeObjectForKey:archiveString];
+	} else {
+        rep = [newObject dictionaryRepresentation];
+        if(rep)
+            [[self commandDict] setObject:rep forKey:archiveString];
+    }
+}
+
 - (void)setDirectObject:(QSObject*)newObject {
     if (newObject != nil && dObject != newObject) {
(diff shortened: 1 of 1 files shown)


skurfer:
I don’t see a problem in `- [QSCommand dObject]`. If there’s a `directID` and it can be used to find and object, the archive is ignored. But there’s obviously a problem somewhere. Rather than try to find it, I’ll just merge this locally and test it out.

Were you able to reproduce the problem prior to this change?

skurfer:
I haven’t tried to reproduce the problem to know if this fixes it, but it doesn’t seem to _break_ anything, so…

Please look at and adjust the release notes if I explained this incorrectly @pjrobertson.

pjrobertson:
looks good. It doesn't *break* anything... haha, best kind of code change out there ;-)

PR #2451, opened 2018-07-31
Disk Image improvements

skurfer:
So that wasn’t too hard. We don’t have a nice layout or instructions, but it’s still an improvement.

* Add a symbolic link to `/Applications`
* Sign the DMG


skurfer:
I found a (possibly stupid) way to customize the layout inside the DMG.

PR #2452, opened 2018-08-01
small changes to the scoring algorithm

### LINKED ISSUE #2450: Three bugs in the QSSense scoring algorithm
I've converted the Quicksilver scoring algorithm to JavaScript for various projects, and I believe I've found a few bugs in `QSSense.m`.  (I've only learned as much Objective-C as I needed to translate the algorithm, but I'm fairly sure these bugs are real.)

Several years ago I started with this [old version](https://web.archive.org/web/20100901004019/http://docs.blacktree.com/quicksilver/development/string_ranking) of the code, which used `NSString`.  That worked well, though I didn't really have a way to verify that the JS version was returning the same scores as the Objective-C version.

Recently, I found this repo and the `TestQSSense.m` file.  The scores in `testScoreSimpleNS()` from the [initial commit](https://github.com/quicksilver/Quicksilver/commit/0f2b4043fafbfc4b1263b7807504eb1b3baaeab8#diff-86c92ca75387e03f87312001fe115fb9) matched the scores from my JS version, but I couldn't get it to consistently match the new values, even when I added the `/2` to `score -= (matchedRange.location-strRange.location)/2;` (which wasn't in the original code above).  Some of the tests produced the same scores, but others didn't.  As far as I could tell by inspection, they all should've been the same, so it was very confusing.

I finally just gave in and ran the ObjC code in Xcode to see what was going on.  After stepping through it (and cursing the absurd complexity of simply dumping strings to the console in Cocoa/ObjC), I believe there are three issues:

1. The [line](https://github.com/quicksilver/Quicksilver/blob/551b8ce027bcd1bbaf150a91a5bc673149bf34a1/Quicksilver/Code-QuickStepFoundation/QSSense.m#L120) where the `/2` was added to increase the scores for longer strings throws away the fractional part of the result, since it's not dividing by `2.0`.  I suppose that could be intentional, like calling `floor()` on the result, but it seemed odd since `score` is a float.  Making the divisor a float brought more of the scores from my code into alignment with the test file.

2. The [inline buffer](https://github.com/quicksilver/Quicksilver/blob/551b8ce027bcd1bbaf150a91a5bc673149bf34a1/Quicksilver/Code-QuickStepFoundation/QSSense.m#L62) is filled with only part of the string, rather than all of it.  This throws off the [lookup](https://github.com/quicksilver/Quicksilver/blob/551b8ce027bcd1bbaf150a91a5bc673149bf34a1/Quicksilver/Code-QuickStepFoundation/QSSense.m#L99) of the current and previous chars, since the lookup is done from `matchedRange.location`, which is in relation to the start of the entire string.  But `inlineBuffer` is filled from `strRange`, which moves to the right with each recursion.  So the check of whether the previous char was a word separator or the current char is uppercase is looking at the wrong letters, which then causes the score to be adjusted incorrectly.  For instance, an abbreviation matching the initial letters of words will have a lower score than it should, since the previous char in the buffer won't be a space, while it is in the original string.  Fixing this in the ObjC code brought the remainder of the scores into alignment.

3. The mask of matching letters can contain more matches than there are letters in the abbreviation, which I mentioned recently in a comment on this [old PR](https://github.com/quicksilver/Quicksilver/pull/2342#discussion_r200827959) (someplace probably no one would see it).

    Consider matching `"test"` against the string `"t e s t tess!"`.  The full abbreviation doesn't appear as a substring in the first loop, so it's reduced to `"tes"` and the function recurses.  That substring is found near the end of the string, and those indices (`[8, 9, 10]`, I think) get set in the mask.   Then it looks for the remaining `"t"` in the rest of the string, but only finds `"s!"`, so it has to start over and shrink the abbreviation again.  Eventually, it finds the individual letters separated by spaces and returns a score.  But the mask still contains those first hits that it gave up on.

    My solution was to clear the mask just before the [`return 0;`](https://github.com/quicksilver/Quicksilver/blob/551b8ce027bcd1bbaf150a91a5bc673149bf34a1/Quicksilver/Code-QuickStepFoundation/QSSense.m#L127) at the end of the function.  I believe that's correct, as the 0 score means the algorithm will start over with a shorter abbreviation, and that shorter bit might hit in a different place, rendering the previous hits obsolete.  If the UI isn't underlining all those letters, then something else must be cleaning up the mask.  (I tried clamping it to the length of the abbreviation, but I don't think that gives the correct results in all cases.)

It looks like the inline buffer was added [9 years ago](https://github.com/quicksilver/Quicksilver/commit/516ba6f31a96e760e62f42531e070e5e4d741917), so maybe it doesn't make sense to fix the behavior now.  But it should be noticeable if you have two somewhat similar strings and an abbreviation that matches the start of words or uppercase letters in one, and random letters near the beginning of the other.  The first match might score lower than the second when it shouldn't.

Anyway, I'm just happy to solve the mystery of why my code wasn't returning the same scores when it looked like it should.  My rough test version of `QSSense.m` that I was using to step through the code and print out debug info is in [this gist](https://gist.github.com/fwextensions/2bbdd453bc941dcfa36b06065c760980).

skurfer:
I addressed the first two points here @fwextensions. On the third, I don’t doubt the bug, but the fix doesn’t seem right. At that point in the code, `mask` is going to be thrown away forever anyway. Clearing it wouldn’t make a difference. (I tried and the scores didn’t change.) I haven’t had a chance to dig in and figure out the actual fix.

closes #2450

### DIFF
--- a/Quicksilver/Code-QuickStepFoundation/QSSense.m
+++ b/Quicksilver/Code-QuickStepFoundation/QSSense.m
@@ -58,7 +58,7 @@ CGFloat QSScoreForAbbreviationWithRanges(CFStringRef str, CFStringRef abbr, id m
 	// Create an inline buffer version of str.  Will be used in loop below
 	// for faster lookups.
 	CFStringInlineBuffer inlineBuffer;
-	CFStringInitInlineBuffer(str, &inlineBuffer, strRange);
+	CFStringInitInlineBuffer(str, &inlineBuffer, CFRangeMake(0, CFStringGetLength(str)));
 	CFLocaleRef userLoc = CFLocaleCopyCurrent();
 
 	CGFloat score = 0.0, remainingScore = 0.0;
--- a/Quicksilver/Code-QuickStepFoundation/QSSense.m
+++ b/Quicksilver/Code-QuickStepFoundation/QSSense.m
@@ -114,7 +114,7 @@ CGFloat QSScoreForAbbreviationWithRanges(CFStringRef str, CFStringRef abbr, id m
 							score -= SKIPPED_SCORE;
 					}
 				} else {
-					score -= (matchedRange.location-strRange.location)/2;
+					score -= (matchedRange.location-strRange.location) / 2.0;
 				}
 			}
 			score += remainingScore * remainingStrRange.length;
--- a/Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
+++ b/Quicksilver/Tests/Tests-QSFoundation/TestQSSense.m
@@ -56,43 +56,43 @@
 	NSMutableIndexSet *indexes = [[NSMutableIndexSet alloc] init];
 	NSIndexSet *results = nil;
 
-	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	score = QSScoreForAbbreviationWithRanges(str, abbr, indexes, strRange, abbrRange);
 	XCTAssertEqualWithAccuracy(score, 0.901851, ACC);
 	results = [NSIndexSet indexSetFromArray:@[@(0), @(5), @(15), @(16), @(22), @(23)]];
 	XCTAssertEqualObjects(indexes, results);
 
 	strRange.length += STEP;
-	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	score = QSScoreForAbbreviationWithRanges(str, abbr, indexes, strRange, abbrRange);
 	XCTAssertEqualWithAccuracy(score, 0.901612, ACC);
 	results = [NSIndexSet indexSetFromArray:@[@(0), @(5), @(15), @(16), @(22), @(23), @(26)]];
 	XCTAssertEqualObjects(indexes, results);
 
 	strRange.length += STEP;
-	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	score = QSScoreForAbbreviationWithRanges(str, abbr, indexes, strRange, abbrRange);
 	XCTAssertEqualWithAccuracy(score, 0.901428, ACC);
 	results = [NSIndexSet indexSetFromArray:@[@(0), @(5), @(15), @(16), @(22), @(23), @(26)]];
 	XCTAssertEqualObjects(indexes, results);
 
 	strRange.length += STEP;
-	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	score = QSScoreForAbbreviationWithRanges(str, abbr, indexes, strRange, abbrRange);
 	XCTAssertEqualWithAccuracy(score, 0.72948, ACC);
 	results = [NSIndexSet indexSetFromArray:@[@(0), @(5), @(15), @(16), @(22), @(23), @(26), @(36)]];
 	XCTAssertEqualObjects(indexes, results);
 
 	strRange.length += STEP;
-	score = QSScoreForAbbreviationWithRanges(str, CFSTR("test"), indexes, strRange, abbrRange);
+	score = QSScoreForAbbreviationWithRanges(str, abbr, indexes, strRange, abbrRange);
 	XCTAssertEqualWithAccuracy(score, 0.69883, ACC);
 	results = [NSIndexSet indexSetFromArray:@[@(0), @(5), @(15), @(16), @(22), @(23), @(26), @(36), @(40), @(41)]];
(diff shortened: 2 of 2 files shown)


fwextensions:
> At that point in the code, mask is going to be thrown away forever anyway. 

Is it?  I thought the caller passed it in by reference.  I don't know what the `id` type is in the signature, but in the test file the mask param is defined as `NSMutableIndexSet *indexes`, which I assume is pointer to the set, so the end of the function wouldn't have any impact on the caller's value.  

> Clearing it wouldn’t make a difference. (I tried and the scores didn’t change.)

Yes, clearing it doesn't make a difference in the scores, just in the set of indices of the matched characters.  I'm guessing the UI just uses the first `abbr.length` indices and ignores the rest, but those won't necessarily correspond to the matches that were used for the score.  

I'd also recommend resetting the index set after each `strRange.length += STEP;`.  Otherwise, it's just accumulating indices from all the previous calls, which isn't how the scorer would normally be called. 

Otherwise, :+1:

skurfer:
> Yes, clearing it doesn't make a difference in the scores, just in the set of indices of the matched characters.

I see what you mean now. What’s odd is that @tiennou specifically wrote the tests as if he was expecting the abbreviation to be repeated within the same string. I’ve asked for clarification on that.

In the meantime, I’ve made the change to clear the mask and am testing it locally. Hoping to get a release out before Mojave.

Quicksilver/Code-QuickStepFoundation/QSSense.m
```diff
@@ -124,5 +124,6 @@ CGFloat QSScoreForAbbreviationWithRanges(CFStringRef str, CFStringRef abbr, id m
 		}
 	}
 	CFRelease(userLoc);
+	[mask removeAllIndexes];
```
tiennou: AFAIR, the mask-managing thing is supposed to be the responsibility of higher-level code (ie. it's not supposed to pass an already-populated/non-empty index set in here), so I'm somewhat suspicious that this could cause a spurious reset of a correctly matching mask in a (previous) recursive step in case a subsequent one doesn't. I'm failing to imagine an example of it though, so maybe the algorithm already caters to that ?

skurfer: > AFAIR, the mask-managing thing is supposed to be the responsibility of higher-level code (ie. it's not supposed to pass an already-populated/non-empty index set in here)

But it was passing a non-empty set. If you look at `testLongString` (before the changes here), the index set would have up to 10 matches for an abbreviation 4 characters long. Though now I’m wondering if that’s because the test was reusing the index without clearing it each time. You might be right about the behavior of the actual application never doing that.

fwextensions: There are two index set issues.  One is that the test code wasn't passing in a fresh set to each call, like you normally would, so it was caching hits from previous calls. 

The other is that the algorithm will leave behind hits from abandoned iterations if the set isn't cleared before returning 0, which is what this line changes.  I mentioned the example of matching `"test"` against `"t e s t tess!"` in [this comment](https://github.com/quicksilver/Quicksilver/pull/2342#discussion_r200827959) on the PR that originally added this test (a year later, so I'm assuming no one saw it). 

Other than this one file, I have zero familiarity with the Quicksilver code base, but I'm betting there's something that is picking just the first `abbreviation.length` chars as matches to render, so that the "extra" hits left in the index set aren't shown.  I think that works in most cases, but I believe there are edge cases where it would show hits that don't actually correspond to the match that was used to calculate the score.

fwextensions:
>  I would only expect indexes from a "valid match path" to be included, since we only add found things (at line 85).

If you match `"test"` against `"t e s t tess!"`, you'll see that the hit indexes are `[ 0, 2, 4, 6, 8, 9, 10 ]`.  The `[8, 9, 10 ]` are left over from the first iteration.

> Does only moving the indexset addition just before the return score at line 123 make sense, so we'd only add things that recursive calls would have considered valid ?

Good idea.  I tried doing that in my JS version of the algorithm and all the [test suites](https://github.com/fwextensions/quick-score/blob/develop/test/quick-score.test.js) passed.  The hits do get set in reverse order, though I assume that doesn't matter for a `mask`.  (It did for me, since I was pushing them onto an array, but using `unshift()` to push them onto the beginning of the array instead left them in the correct order at the end.)  So maybe that's the cleaner fix. 

> Shouldn't a 3-letter match from the start of a word be higher than a 2l + l ?

It's due to the lengths of the strings.  If you make them the same length, the `"Divinity - Or"` string scores slightly higher.

skurfer:
I haven’t been following this too closely, but if someone wants to add the proposed fix, I’ll look it over and try to get it into 1.6.1. I’d like to release it soon since we currently have a bug that prevents QS from launching under 10.10.

PR #2476, opened 2018-11-01
refer to the manual instead of the wiki

skurfer:
Probably anything useful in the wiki should be relocated to the manual. The idea behind the wiki was that anyone could contribute, but no one does. And the manual is a public git repo now, so it’s not as exclusive.


skurfer:
I merged this all into #2486

PR #2486, opened 2019-02-01
Point 'more info' button to Changelog page, not just website homepage

pjrobertson:
I went to actually update QS today (yes, I haven't updated for over 1 year. Shame on me!) only to find out that there is a new version (thanks @skurfer ).

I figured this 'More info' button could actually point to the changelog though :/

screenshot 2019-02-01 at 13 09 11


skurfer:
Agreed. Maybe we should also change `kHelpURL` to [the manual](https://qsapp.com/manual/), now that it’s online?

pjrobertson:
Yep, makes sense. Does that mean we want to deprecate the wiki? If not, we should at least have a link to the wiki from the manual (explained as a 'repository of user tips and scripts')

pjrobertson:
Oh how I've missed QS coding...! What you'd have thought was a simple one liner change turned into a can of worms! Turns out some of the `help` code was never used, so I've removed it.

Secondly, I couldn't get this to build because Xcode was complaining about certain string files (Thai, Portuguese) not being in the right encoding. I played around with changing the encoding to UTF-8 but didn't want to break anything anywhere else so left things as is.

screenshot 2019-02-03 at 09 44 06


@skurfer what are the file encodings of the Thai string files on your computer (go to the Quicksilver/Localised/th.lproj folder and run `file -I ./*`
@tiennou would also be interested on your input on the strings thing. I'd argue to convert everything to UTF-8 then tell Xcode they're all UTF-8, but I'm wondering if things have got messy because the encoding changes locally on my computer?

skurfer:
I’ve merged my changes, since there were some more files touched, and made some small fixes here. I’m good to merge this if you are.

pjrobertson:
I never replied, but all looks good. Thanks for doing the tidying up and typo fixing

PR #2487, opened 2019-02-04
fix encoding errors for Thai localization files

skurfer:


pjrobertson:
Just as an FYI: nice - I saw this problem when I was trying to build previously. Glad you fixed it!

PR #2489, opened 2019-03-19
Background proxy side-effects

skurfer:
I was trying to show nicer details for a proxy object by pulling them from the resolved object, but calling `resolvedObject` while getting details tends to happen before the object gets selected in the main interface. (I think because it appears in the results during a partial match.) That forces the object to be resolved and cached before it’s actually displayed.

There also seems to be something calling for details just after the interface is dismissed, which is even worse because of this sequence:

1. The user dismisses the interface
2. The proxy cache is intentionally cleared
3. Something calls `detailsOfObject:` which calls `resolvedObject` which grabs and immediately re-caches the current value of the proxy

Then, sometime later:

4. The user calls up the proxy
5. It has a (potentially very old) cached value, so that gets returned

I’m not sure what the best answer is here, other than for developers to avoid calling `resolvedObject` when they just want the already resolved value. So for now, this gives us a way to do that.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSProxyObject.h
+++ b/Quicksilver/Code-QuickStepCore/QSProxyObject.h
@@ -33,7 +33,10 @@
 + (id)proxyWithDictionary:(NSDictionary*)dictionary;
 + (id)proxyWithIdentifier:(NSString*)identifier;
 - (NSObject <QSProxyObjectProvider> *)proxyProvider;
-- (QSObject*)proxyObject;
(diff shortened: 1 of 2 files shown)


pjrobertson:
I'm assuming you're planning on doing something with `cachedProxyTarget` once it's merged, so I'll merge this and wait and see!

skurfer:
Yeah, I was going to use it in the Zoom plug-in. It turns out this didn’t really solve the problem, but including it is pretty harmless.

PR #2490, opened 2019-04-16
Make length of QSHistory editable

pjrobertson:
Nothing much here. I realised the QSHistory preset in the catalog (Catalog > Quicksilver > Recent Objects) is limited to 50. That's not enough for me.

I made it a hidden preferences to allow you to edit this value (ideally it'd be a setting in the Catalog UI but I don't have the time for this :(


skurfer:
OK, but I say we get rid of `MAXHIST` entirely and set the default in `QSDefaults.plist`, since that’s the standard way we handle such things.

tiennou:
And I say "add it to the Advanced prefs" 😉. IIRC it's just about a few lines in a plist (if it supports integer values).

skurfer:
I did the things. 🆒

pjrobertson:
Looks good. Can I merge a PR that I started? I say yes! :D

skurfer:
Yup.

skurfer:
Of course I just noticed this, but you can’t actually change the number in Extras. You can see what it is, though! 😁

tiennou:
Gosh, I wonder whose fault it is 😁.

I was pretty sure it was working… Oh well, at least it's somewhat auto-documenting.

pjrobertson:
😨 Ooops!

I think we should try and do away with the 'Extras' tab in the future anyway. It's not very self-explanatory, and some of the prefs should maybe even be deleted. Another job for another day :P

PR #2493, opened 2019-05-18
Update list of TLDs

pjrobertson:
Turns out there are quite a few more TLDs these days! Just tried to type 'something.watch' into QS, which didn't work. Updated based on latest list I just pulled from IANA


skurfer:
Cool. does this kill string sniff performance?

pjrobertson:
Good question. I haven't profiled it, but my guess is since all we're doing is one lookup on an NSArray for the TLD, we're not really going to be hampering performance. And still, this check only gets run on strings that:
a) already have a dot (`.`) in them
b) don't have any whitespace
c) don't have a `@` in them
d) don't have more than one `:` in them
...

the list goes on. I think the point here is that this is actually one of the last checks to get done, when it's quite likely the string is in fact a URL

Issue #2497, opened 2019-06-27
Catalina.  Any effort to make QS compatible with it?


juanfal:
For questions or tech support, please start with the [users’ group][1]. If you know the problem is in Quicksilver itself, report it here.

[1]: https://groups.google.com/forum/#!forum/blacktree-quicksilver

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**What version of macOS?**

**What version of Quicksilver?**

juanfal:
**Catalina**, MacOS 10.15 beta, rejects many of QS plugins as no from recognised developers.  Funnily some keyboard shortcuts still work, others not.  No sense for me.  Of course, other System triggers, like "change location" when starting up, do not work at all anymore.  Sad, very sad.  I couldn't live without QS after these many many years using so intensely!

It says (and repeat through alerts, as many times as you eventually need to quit QS)

     com.blacktree.Quicksilver.QSEventTriggersPlugIn.73.qsplugin

comes from an unidentified developer

juanfal:
Desperately trying to recompile sources, XCode raises the error:
`Unable to resolve build file: XCBCore.BuildFile (Named reference 'FMDB.framework' cannot be resolved)`
Ive tried to find the link, whatever, no luck.

juanfal:
It seems I've found a way to avoid the problem con QSEventTriggers plugging:

I've manually taken off the file:
`/Users/me/Library/Application Support/Quicksilver/PlugIns/com.blacktree.Quicksilver.QSEventTriggersPlugIn.73.qsplugin`
and then, from the Quicksilver plugins panels, installed `Event Triggers`

The problem seems solved

tiennou:
It seems Catalina introduced new codesigning checks when loading bundles (a.k.a plugins).

This is going to be as messy as it can get, as IMHO it entails reworking how the build is done, because `/tmp/QS` was a good idea 15 years ago, but now I think it's an impediment to having plugins codesigned.

Please realize that *unless someone stepped up* most if not all plugins are still using manual memory management, most of the things we do with aliases and resources forks have been deprecated for like 4 OS releases, most of our frameworks predate the "modular framework", so Xcode will now also complain about these when headers are missing.

[...]

PR #2503, opened 2019-09-19
fix some crashes under Catalina

skurfer:
This should force some lingering UI things onto the main thread. These have been warnings for a while, but under Catalina, they will crash QS.

Progress toward #2497

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/QSObjectView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSObjectView.m
@@ -182,14 +182,21 @@
   previousObjectValue = aValue;
 }
 
-- (id)objectValue { return [[self cell] representedObject];  }
+- (id)objectValue {
+	__block id objVal;
+	QSGCDMainSync(^{
+		objVal = [[self cell] representedObject];
+	});
+	return objVal;
+}
 - (void)setObjectValue:(QSBasicObject *)newObject {
     [self setPreviousObjectValue:[self objectValue]];
 	[newObject loadIcon];
 	[newObject becameSelected];
-	// [self setToolTip:[newObject toolTip]];
-	[[self cell] setRepresentedObject:newObject];
-	[self setNeedsDisplay:YES];
+	QSGCDMainSync(^{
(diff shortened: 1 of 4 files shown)


skurfer:
Any objections to how this was handled @tiennou @pjrobertson ?

I had originally changed the code that posts the notification that triggers `catalogIndexingFinished:`, but in the end, I decided that code should be on the main thread no matter how it’s called.

Quicksilver/Code-QuickStepInterface/QSObjectView.m
```diff
@@ -182,14 +182,21 @@ - (void)setPreviousObjectValue:(QSObject *)aValue
   previousObjectValue = aValue;
 }
 
-- (id)objectValue { return [[self cell] representedObject];  }
+- (id)objectValue {
+	__block id objVal;
+	QSGCDMainSync(^{
```
tiennou: That's hacky as possible, but… yeah. The argument I'm going through when trying to fix those kind of issues is mostly this :

> does the "function" performed by the code touches AppKit objects, (ie views) ? does it benefit from being able to be multi-threaded ?

In that case, this is a view, so there should be no need to dispatch anything to the main thread — you *are* in the view's code, so it's whatever controller-level background-receiving notification-handling code that caused that call to happen that needs some proxying. OTOH, those thread-safety issues are *everywhere*, so if that fixes most of the crashes, I'm fine with 🚢.

tiennou: Thanks for doing that @skurfer. I don't intend to sound harsh in any way shape or form, but…

> whatever controller-level background-receiving notification-handling code

could call any other AppKit object, and you'd still get a crash.

I do have a non-dying MacBook Pro again, so hopefully I'll be able to actually *open* the QS project, and help with some of the issues/PRs/cleanups.

skurfer: I did put a break point in there to see where `objectValue` was being called from so I could just make sure all of that stuff was on the main thread instead, but like you said, it’s _everywhere_.

PR #2510, opened 2020-02-10
index /System/Applications by default

skurfer:
For Catalina


skurfer:
This should be uncontroversial, so I’ll just merge.

PR #2521, opened 2020-05-26
don’t release Large Type windows on close

skurfer:
This prevents the app from crashing when dismissing the Large Type interface. I have a VM running Catalina, so I was able to test there too.


skurfer:
Oh, I also threw in a small casting fix I've had uncommitted locally forever.

skurfer:
Even better, I removed the array altogether. Everything still seems to work.

PR #2544, opened 2020-11-30
Update main app icon to the Big Sur style.

### LINKED ISSUE #2542: Big Sur Icon
Copied from [a mailing list email](https://groups.google.com/g/blacktree-quicksilver/c/QQ8t1RtM4nU).

Now that Big Sur is out, iOS-style rounded squares are the new icon style:
https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/
https://developer.apple.com/design/resources/#macos-apps

I haven't seen an updated icon for Quicksilver, so I thought I'd take a stab at it:

icon_512x512@2x

In context with the Big Sur dock (with the old icon on the right for comparison):

Screen Shot 2020-11-28 at 1 55 50 AM

I tried to match the outer guides of macOS icon template while keeping as much of the original icon shape as possible:

Screen Shot 2020-11-28 at 01 43 04

If I can get a signal in this issue from a core developer that this is a good direction, I'd be happy to pixel-tweak the full range of resolutions and send a PR to replace https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/Resources/Images/Quicksilver.icns !

lgarron:
This is based on the template from https://developer.apple.com/design/resources/ .

The @1x and @2x resolutions could really be in the same file, but to
avoid any breakage from the project configuration, the resolutions are
broken across the two existing files.

Closes #2542.


lgarron:
Here's how all the resolutions look:

icon_16x16
icon_16x16@2x
icon_32x32@2x
icon_128x128
icon_128x128@2x
icon_256x256@2x
icon_512x512@2x

If there is a `@2x` resolution with the same image pixel width as another `@1x` resolution, they share the same icon. I believe this is fairly common for `icns` files, and it works well in practice.

lgarron:
Ideally these two resources should be one icon, but I can't quite reverse-engineer 64767ed878ca85fd8b8b2021ac28e91e7ec4f78f to figure out why how to collapse them again.

skurfer:
It looks great. Thanks!

Issue #2543, opened 2020-11-29
ARM M1 compiling


juanfal:
M1 is there.  It's the greatest increase in speed in one step I've ever seen.  Quicksilver is, by far, the most useful utility made for OSX ever.  QS has several issues on Big Sur, specifically the copy-paste history… but apart from fixing bugs, it'd be great to know if is there any plans to recompile it for ARM

skurfer:
I can get it to compile in Xcode 12, so it should be possible in theory, but there are a lot of problems with the resulting application that you don’t see in the current version (built with Xcode 9), so it isn’t something we would want to release as is.

I would like to fix the bugs, but I don’t know if I’ll have time, or if I would even know how. I can only speak for myself though. It’s an open source project that anyone could contribute to.

pasiaj:
@skurfer, I'd love to start helping out on M1 with a long term goal of having a branch that can be compiled on latest Xcode without modifications and then start encouraging others to jump on. 

If someone was to help me get started with getting the source to compile on Xcode 12.4. 

Right now the build fails on all build targets with:
"unable to find sdk 'macosx10.13'"

juanfal:
I did try it too, and started having the same issue.  Not being involved in any other similar project now, sorry I can't help here.

skurfer:
@pasiaj I’ve just committed some changes that allow me to build it under Xcode 12. You’ll want to pick Quicksilver Distribution as the target if it’s not already selected. Let me know if you run into problems. I’ve sent you an invite for Slack as well.

totallyGreg:
Pulled latest check-in
❯ git hist
* 4374aea4 update the Xcode project version (HEAD -> master, origin/master, origin/HEAD) (19 hours ago) [Rob McBroom]
Updated the submodules
❯ git submodule update
Submodule path 'Quicksilver/Code-External/DisableSubviews': checked out 'f569e60e9e46220bce701dcfe7b92f4d8abe01ad'
Submodule path 'Quicksilver/Code-External/FMDB': checked out '6c0ff9c423e8b476c63f11413790443c3fe62f16'
Submodule path 'Quicksilver/Code-External/LaunchAtLoginController': checked out 'da5db7a0a20e76f83a10f7ecd34c81d1def7a64d'
Submodule path 'Quicksilver/Code-External/VDKQueue': checked out '503c187d0a418d0af8c90a6216435b3052b0b223'

and Xcode 12.4 Build fails with the same error 
Showing All Messages
[...]

PR #2557, opened 2021-04-13
Relaunch as ARM64 if running ARM64

n8henrie:
Without this, running as Apple architecture -> `Relaunch Quicksilver` -> running as Intel.

This is the first line of Objective C I've ever written, but seems to work.

Related issue: https://github.com/quicksilver/Quicksilver/issues/2543


skurfer:
Thanks for the fix. We probably shouldn’t default to `i386` any more, since that will never work. Can you change it to check for `arm64`, and default to `x86_64` otherwise?

PR #2560, opened 2021-05-25
Prefer running native on ARM64

n8henrie:


esoel:
Thank you so much for porting quicksilver to arm. I really appreciate it. After updating and relaunching, the intel version was relaunched, but closing it and reopening fixed the issue. Just thought I'd put it here in case someone else has the same issue. It's probably not worth it to create an issue (?), but if you want me to I can create one. Now I'll go donate...

n8henrie:
@esoel thanks for your kind words, glad it's working!

PR #2570, opened 2021-10-31
feat: Add Github Actions build

gliptak:

https://github.com/gliptak/Quicksilver/runs/4062461862?check_suite_focus=true

the build fails, please offer pointers on the right command line build command

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: ci
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [macOS-11, macOS-10.15]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: latest-stable
+    - run: |
+        cd Quicksilver
+        git submodule init; git submodule update 
+        xcodebuild -project "Quicksilver.xcodeproj"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,5 +20,6 @@ jobs:
         xcode-version: latest-stable
     - run: |
         cd Quicksilver
-        git submodule init; git submodule update 
-        xcodebuild -project "Quicksilver.xcodeproj"
+        git submodule update --init --recursive
+        cd Quicksilver
+        ./Tools/qsrelease
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,5 +21,4 @@ jobs:
     - run: |
         cd Quicksilver
         git submodule update --init --recursive
-        cd Quicksilver
         ./Tools/qsrelease
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,0 @@
-language: objective-c
-osx_image: xcode8.3
-xcode_project: Quicksilver/Quicksilver.xcodeproj # path to your xcodeproj folder
-xcode_scheme: "Quicksilver Distribution"
-script: |
-    ruby Quicksilver/Tools/travis/check_indent.rb


skurfer:
The official releases are created with [the `qsrelease` script][1], so maybe there’s something in there you can use. I’m pretty sure you have to manually change something in the SQLite project before it will build, though.

What are you trying to do?

[1]: https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/Tools/qsrelease

n8henrie:
I presume he's trying to get an automated build process through GitHub Actions, which would allow the build artifacts to be automatically generated for new tags / released and subsequently stored and distributed without requiring other work.

I [had mentioned](https://github.com/quicksilver/Quicksilver/issues/2543#issuecomment-887079401) that something like this could be really helpful once GitHub released MacOS 11 containers\* (looks like this has happened now), as I think it should allow building universal binaries to support M1 Macs.

\* <https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources>

skurfer:
That sounds ideal, but the project was designed to be built with the tools of 2003 and not much more than duct tape has been added since.

I think effort should be directed toward modernizing and cleaning up the Xcode project so it builds easily and predictably on a local machine before we even think about doing stuff on the GitHub side.

(This is the only Xcode project with any complexity I’ve ever worked with, so I have no context for the “right way” to lay it out or I would have fixed it a long time ago.)

gliptak:
@n8henrie the build is green with your last update, so maybe start with merging this PR

https://github.com/gliptak/Quicksilver/actions/runs/1417799713

There are no M1 builders. I don't have specifics on how to cross-compile

gliptak:
This what `1.6.1` from qsapp.com shows:
quicksilver
Would XCode project updates allow for this check?

n8henrie:
gliptak I think you just need to "show in finder" then right click and open.

n8henrie:
> I think effort should be directed toward modernizing and cleaning up the Xcode project so it builds easily and predictably on a local machine before we even think about doing stuff on the GitHub side.

@skurfer I see your point, but I think this will go a long way towards the modernizing process -- allowing CI to run tests, automatically provide a release stream / etc. Also helps provide a lot of transparency -- having the binary be created and hosted directly through GitHub Actions gives a little more confidence that the app they are downloading and installing matches the open source code they're (ideally) examining (makes it harder to do a bait-and-switch between the provided app and the open source code).

With a couple minor tweaks I was able to get it working -- and it looks like it's creating a universal binary that is running natively on Apple Silicon for me. Can you give it a shot and see if it works okay on an Intel mac? If so, I'll make another couple tweaks to cut a release on new tags and we should be able to merge.

AFAICT this should also close #2543!

https://github.com/n8henrie/Quicksilver/releases/tag/1.6.1

n8henrie:
Actually, since the only changes are in the `.github` folder (not directly affecting any application code), and since I don't think it will automatically create a release yet (not until a new tag gets pushed), I'm going to go ahead and squash and merge #2573 (which includes this PR as well).

skurfer:
> Can you give it a shot and see if it works okay on an Intel Mac?

I’m not on an Intel Mac any more. 😏 So I’ll hopefully be more active on this as a result. I can have a co-worker try it out or borrow an older machine from my kids, though.

jschwalbe:
> EDIT: Direct link: https://github.com/n8henrie/Quicksilver/releases/download/1.6.1/Quicksilver.1.6.1.dmg

@n8henrie, I cannot get this to run, unfortunately.

image

Thoughts?

n8henrie:
Hmm, thanks!

What is your hardware and OS?

jschwalbe:
MacBook M1 Pro
macOS Monterey 12.0

n8henrie:
@jschwalbe -- odd, I'm getting the same on my M1 MBP / Monterey as well now. I remember downloading and testing before I posted that...

If I directly execute the contained binary (`Quicksilver.app/Contents/MacOS/Quicksilver`) it works. Weird.

n8henrie:
I just cut a new release and it seems to be working. Can you try again?

<https://github.com/n8henrie/Quicksilver/releases/download/v0.0.8/Quicksilver.1.6.1.dmg>

jschwalbe:
That worked! odd

PR #2574, opened 2021-11-07
Upload artifacts to PRs and merges

gliptak:

uploaded to the Artifacts section

https://github.com/quicksilver/Quicksilver/actions/runs/1432475302

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - "v*.*.*"
+  pull_request:
 
 jobs:
   build:
@@ -15,7 +16,13 @@ jobs:
     - run: |
         cd Quicksilver
         ./Tools/qsrelease
+    - name: Upload document
+      uses: actions/upload-artifact@v2
+      with:
+        name: Quicksilver.dmg
+        path: /tmp/QS/build/Release/Quicksilver*.dmg
     - name: Release
       uses: softprops/action-gh-release@v1
+      if: startsWith(github.ref, 'refs/tags/')
       with:
         files: /tmp/QS/build/Release/Quicksilver*.dmg


n8henrie:
@gliptak What does this accomplish?

gliptak:
@n8henrie it allows to download QuickSilver specifically built for a PR for local testing

n8henrie:
I see, that makes sense, but I would be a little concerned about consuming *all* of the free CI time with PRs (and then having none left to make a release):

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes



Product | Storage | Minutes (per month)
-- | -- | --
GitHub Free | 500 MB | 2,000

> Jobs that run on Windows and macOS runners that GitHub hosts consume minutes at 2 and 10 times the rate that jobs on Linux runners consume. For example, using 1,000 Windows minutes would consume 2,000 of the minutes included in your account. Using 1,000 macOS minutes, would consume 10,000 minutes included in your account.

It looks like this run took about 3.5 minutes on your fork: https://github.com/gliptak/Quicksilver/actions/runs/1432457097 , so that would be a little under 2% of the *total* minutes allotted for a month.

I suppose we could enable it and see how it goes (perhaps needing to disable later), or consider filtering further so it only ran on specific tags (as opposed to every PR).

gliptak:
@n8henrie I didn't know the MacOS 10x multiplier ...

Even with that, you can have 50 PR runs a month. It might be worth enabling and observing how it goes

skurfer:
In practice, I would usually merge branches from every PR I was considering to make sure they worked well together and not just in isolation. I don’t think there’s a way to automate that, but then again, worrying about multiple PRs simultaneously might not be something we need to worry about these days.

n8henrie:
Makes sense.

Well why don't we merge as is for now and see how it goes. Seems like worst case scenario is this would run out of minutes and block CI for a month, potentially delaying a release (if this became the default workflow).

Still need to figure out whether automated signing is possible / desirable: #2576

skurfer:
At the current rate, it shouldn’t be an issue.

PR #2575, opened 2021-11-07
Add GHA badge to README

gliptak:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,5 +1,7 @@
 # About Quicksilver on Github #
 
+[![release](https://github.com/quicksilver/Quicksilver/actions/workflows/release.yml/badge.svg?branch=master)](https://github.com/quicksilver/Quicksilver/actions?query=workflow%3Arelease+branch%3Amaster)
+
 This repository contains the current source code of Quicksilver. If you're having issues with the latest Quicksilver version, feel free to log them at the [issue tracker](https://github.com/quicksilver/Quicksilver/issues). 
 
 The `master` branch contains the source for the currently released Quicksilver.


n8henrie:
I don't think using a badge for the `release` action makes sense, this is just cutting a release for new tags.

Using a badge in this way usually conveys some information to the reader (tests passing, test coverage percent, or just release version) -- this script currently is just creating a release.

n8henrie:
I think it might be worthwhile to add a badge for the test suite, but I don't know how to run that yet.

Did you have other ideas regarding the purpose of adding this badge?

I also think that as soon as @skurfer gives the OK changing the links in the readme might be a worthwhile PR.

We should also look into whether or not we can sign releases with this github action -- hoping there may be some way to use the github environment secrets to do so.

gliptak:
The approach might be considered here is to separate a build/test flow (validating PRs and indicating health of master branch and displaying badge) from the release flow (triggered when a v* tag is pushed onto master)

skurfer:
> I also think that as soon as @skurfer gives the OK changing the links in the readme might be a worthwhile PR.

Changes to which links?

> We should also look into whether or not we can sign releases with this github action -- hoping there may be some way to use the github environment secrets to do so.

Has anyone gotten signing to work locally? I haven’t seen it work for years. I’ve tried a lot of things to fix it, but wasn’t able to figure out the problem.

n8henrie:
> Changes to which links?

The official downloads can come from GitHub once this is working: <https://github.com/quicksilver/Quicksilver#where-to-download-quicksilver>

> Has anyone gotten signing to work locally? 

I assumed someone would need a (paid --$100 or so?) Apple developer signing certificate. I will probably get one in the next year but don't have one now. I can look for some kind of self-signed cert (which I think they generate for e.g. local iOS builds), but I assume it will give the same kinds of warning if I use it to sign an app that is distributed.

gliptak:
https://github.com/quicksilver/Quicksilver/issues/2576#issuecomment-964647533

skurfer:
> I assumed someone would need a (paid --$100 or so?) Apple developer signing certificate.

Correct. We (meaning me) have had one since Gatekeeper was introduced, but you will need one to try things out locally.

> I will probably get one in the next year but don't have one now.

If you end up getting one, let us know. @pjrobertson can reimburse you using donation money. We should probably still issue releases with mine for now so it will match prior releases.

gliptak:
@skurfer if you configure the signing cert into GHA following above, I will open a PR to update GHA script

Issue #2571, opened 2021-11-04
Update internal Python packages and scripts


skurfer:
* Update Markdown, etc. to the latest versions
* Use Python 3 to avoid the warning from Monterey

n8henrie:
I know next to nothing about Obj C*, but I am pretty familiar with bash and
python. I see you've self-assigned here -- is this something I could help
with? Don't want to start a PR and deal with merge conflicts if you've got
something already in progress.

* though I hope to contribute to converting some code to Swift, which I
also imagine will help encourage contributors going forward

>

n8henrie:
#2577

Needs additional work, I'll make a few more pushes.

> The way I got those Python packages into QS was by installing them to a virtualenv with pip, then just copying the resulting folder out of site-packages.

`pip download --no-binary=:all: markdown` gets a `tar.gz` (without the `--no-binary` gives a wheel). I'll see if that works, if not fallback to the site-packages workaround.

n8henrie:
@skurfer How does one run `bltrversion` (and what is it for)?

I'm trying from `Tools/` but there are several unset envvars.

```
PYTHONPATH=./python-support INFOPLIST_FILE=org.quicksilver.plist ./bltrversion 
```

n8henrie:
In the future, it looks like this works for vendoring dependencies:

```console
$ pip download beautifulsoup4
$ find . -maxdepth 1 -name '*.whl' -exec unzip {} \;
$ rm -r ./*.dist-info *.whl
$ python3 -c 'import bs4'
```

NB that this will be a platform-specific compiled wheel if any dependencies are not pure python code.

skurfer:
There’s a pretty good explanation of `bltrversion` in [the plug-in dev reference][1]. It gets run by Xcode against the current project when you build a plug-in.

The packages we’re using should be pure Python and one of them (Markdown) only runs on developer machines.

[1]: https://github.com/quicksilver/PluginDevelopmentReference/blob/master/QuicksilverPlug-inReference.mdown#documentation-and-bltrversion

PR #2577, opened 2021-11-09
Start migration to python3

n8henrie:


n8henrie:
Fixes #2571

skurfer:
OK, I hate the idea of complaining when someone is helping out, so I want to be clear that I appreciate you spending time on this.

Now, for my _actual_ first thought when I looked at this: Ugh. Looks like Black has ruined a couple more perfectly readable scripts. If you could convince your editor not to let Black take a dump in these files I’d appreciate it (but honestly, I’m not going to refuse the PR over it). 😏

n8henrie:
¯\\_(ツ)_/¯ Thanks for your honesty. I've gotten used to it (especially after working with gofmt and rustfmt); I value the consistency and find the readability usually better than when others do their own thing. (FWIW, the majority of the 65 files changed were never opened in my editor, black is *ubiquitous* at this point, especially as it's officially under the umbrella of the PSF now.)

Are there certain changes you're particularly turned off by? Double quotes? I'll see if I can configure it.

skurfer:
> FWIW, the majority of the 65 files changed were never opened in my editor

Oh, I know. I don’t care about those and they aren’t worth reformatting since we’ll just replace them again at some point. I’m just referring to the two scripts that we control within the project.

> Are there certain changes you're particularly turned off by? Double quotes?

Yes, mainly the double quotes. Single quotes are easier to read, easier to write, and they match the representation you’ll see in the Python REPL.

> I'll see if I can configure it.

If you’re using Black, you can’t and they’re very proud of it.

n8henrie:
> Single quotes are easier to read, easier to write

I consider the extra effort of pressing `Shift` to be offset by not having to escape apostrophes in strings, and mentally more consistent with Go, Rust, and Swift. But it seems like you have a well-formed opinion here so I'm happy to play ball :) Thankfully a `%s/"/'/g` hopefully did the trick.

> and they match the representation you’ll see in the Python REPL.

That's a good point that I hadn't considered.

> If you’re using Black, you can’t

`skip-string-normalization = true` seems to be a reasonable workaround. I've added to a minimal config file so future contributors won't run into the same issue.

I still need to go through and make sure these are working in XCode, just wanted to get the ball rolling.

n8henrie:
Fixed one more python script in preflight.

Looks like the changelogs are being generated without issue.

skurfer:
OK, I’ll try to give this a look soon.

skurfer:
Oh, I’ll look at this now since I see it’s holding your `master` branch hostage. Feel free to just set `origin` to the main repo and create your branches for PRs there going forward.

skurfer:
Hmm. I assume you don’t see this?

~~~
Traceback (most recent call last):
  File "/tmp/QS/build/Debug/Quicksilver.app/Contents/PlugIns/Core Support.qsplugin/Contents/Resources/QSURLExtractor.py", line 18, in <module>
    from bs4 import BeautifulSoup
  File "/private/tmp/QS/build/Debug/Quicksilver.app/Contents/PlugIns/Core Support.qsplugin/Contents/Resources/bs4/__init__.py", line 36, in <module>
    raise ImportError('You are trying to use a Python 3-specific version of Beautiful Soup under Python 2. This will not work. The final version of Beautiful Soup to support Python 2 was 4.9.3.')
ImportError: You are trying to use a Python 3-specific version of Beautiful Soup under Python 2. This will not work. The final version of Beautiful Soup to support Python 2 was 4.9.3.
~~~

I know `/usr/bin/env python3` gives me something reasonable in a terminal. I’m trying to think of a way to see what it refers to in Quicksilver’s execution context. I don’t see how it could result in a Python 2 executable in any case though.

n8henrie:
I have not seen that.

I did have to change a path in XCode (it wasn't respecting the shebang and
had a hard-coded `python2`), but I thought I included that change in the
PR. Will double check soon.

Are you seeing this at build time, startup, or during usage?

skurfer:
I see it after I build and try to use the HTML link parser.

See the last paragraph in my previous comment for what I did. I’m wondering if you can reproduce it. If not, it must be something on my system.

n8henrie:
> See the last paragraph

Sorry, was going from email which didn't register the edit.

> then tab back to the first pane and hit →

Yes, I see the same error. After ~13 years of continuous, daily, heavy use, with a number of [blog posts dedicated to QS][0] over that time frame, having evangelized it to anyone that will listen as my most-critical MacOS app... I had actually never used / didn't know about this functionality as far as I can recall. And I'm sure I will use it now -- QS is just such a deep rabbit hole.

[As an aside...](https://github.com/quicksilver/Quicksilver/issues/2584)

So [here's the issue][1]. Fixing the path still give some errors with regards to the path and soupsieve (which is needed due to an updated bs4) -- should be an easy fix.

I think hardcoding `/usr/bin/python3` would be expected to break compatibility with prior MacOS  versions that didn't ship with Python3 (was Big Sur the first?). ~~Perhaps just `python3`?~~ That doesn't work, doesn't do a `PATH` lookup because it's not running in a shell I guess, but using `/usr/bin/env` seems to work (since it is executable and has an appropriate shebang). This hopefully should mean pre-Big Sur users can use an e.g. homebrew-installed python3.

Also looks like much of the utilized NSTask API is deprecated. This might be a good candidate for me to [convert to Swift](https://developer.apple.com/documentation/foundation/process) as some point.

[0]: https://n8henrie.com/tags/#Quicksilver-ref
[1]: https://github.com/quicksilver/Quicksilver/blob/40f0f59c39797c251dfbd44563586663b6812ccc/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSHTMLLinkParser.m#L51

n8henrie:
Seems to be working with these.

n8henrie:
Also, I don't see any other user-facing hardcoded python paths (just for one of the build phases which should only be run on dev machines AFAICT), so I think if this works for you it should be ready for merge.

skurfer:
I just ran across something that might be related.

https://indiestack.com/2021/11/xcodes-environmental-pollution/

I can’t remember now, but I think I might have been running a Debug build from Xcode. Still, that wouldn’t explain why it’s running Python 2.

n8henrie:
Did you try again since https://github.com/n8henrie/Quicksilver/commit/ebcd08e53e044ffe146f280203e68e9d74174b69? I was able to reproduce your error and think this fixed it.

n8henrie:
🤦‍♂️ Sorry, I pushed those last changes to the n8henrie/n8henrie branch instead of n8henrie/master. Try again, the python2 issue should hopefully be gone.

Again, once these are merged I'll just branch from Quicksilver and set it up as my default remote.

PR #2582, opened 2021-11-13
cleanup qsrelease: Make POSIX, fail on errors

n8henrie:
- If errors are acceptable, explicity ignore them.
- Move cleanup to a trap instead of a long conditional
- Add and remove quoting where it can / can't make a difference
- Break long lines for readability
- Enable users to pass in a custom signing identity for local testing

PR #2586, opened 2021-12-06
Fail qsrelease if step fails

gliptak:


n8henrie:
Please follow https://github.com/quicksilver/Quicksilver/pull/2582, which contains this and many other changes.

Issue #2576, opened 2021-11-09
Is signing possible through GitHub Actions?


n8henrie:
Since GitHub actions seems readily able to compile and provide a universal binary using the `qsrelease` script, the next logical step would be to see if we can use something like a repository secret to allow signing (and hopefully avoid all the issues about being unable to open since Big Sur).

<https://docs.github.com/en/actions/security-guides/encrypted-secrets>

- [x] Is there a way to keep a signing identity in a GitHub secret?
- [x] Discussion: Is this desirable? Would allow a "rogue contributor" to create a signed copy by pushing a new tag
    - In practice, this doesn't seem much different than not signing them at all, which is what's happening now

gliptak:
https://localazy.com/blog/how-to-automatically-sign-macos-apps-using-github-actions

skurfer:
I didn’t want to create a second issue related to signing, but I wanted to document what we know and have tried (since most of it is probably only in my head) so we can start talking about solutions. Maybe we should rename this issue to “Code Signing Status” and move this to the description, but here goes…

Releases were signed for a long time after Gatekeeper was first introduced. Code signing stopped working somewhere around Xcode 10. For a while, we got away with building QS in a VM running 10.13 with Xcode 9, but even that stopped working. (The signing, not the builds.)

Most people have probably always used Xcode to sign their apps. That will not work for us, so we have always told Xcode not to sign anything, then done it on the command-line in the `qsrelease` script. Why?

  1. Xcode requires you to define the signing cert, but this project is worked on by several independent devs and only one of them (me) has the cert we use.
  2. This project is configured to build several things then put them together after the fact. Xcode will sign `Quicksilver.app` as soon as it’s built, then copy something (like the frameworks or the Core Support plug-in) into the application bundle, rendering the signature invalid. Is the way this project does things “wrong” or unusual? I have no idea, but it must be because Apple clearly assumes a different setup.

The first reason surely has a solution because there must be a ton of projects like this.

On the second reason, we’re pretty screwed right now. As far as I can tell there is no way to tell a recent version of Xcode not to do any signing at all, so the command-line method will simply complain that the app is already signed. (But the signature won’t be valid because of all the other things the build process does.) This is why I say we should be focusing our efforts on redesigning the Xcode project. Maybe if we _could_ let Xcode do the signing, it would actually work.

Even if we could still do it manually, like I said, the command line technique just stopped giving us valid signatures for no apparent reason. Here’s some detail on that.

[...]

PR #2587, opened 2021-12-08
Automate signing

n8henrie:
Making a PR to trigger a CI build so we can see if the signing setup is working.


pjrobertson:
@n8henrie  - does this need to stay open? Did the code-signing work as expected?

Next steps?

n8henrie:
There's more discussion in the associated issue: https://github.com/quicksilver/Quicksilver/issues/2576

Not ready yet: <https://github.com/quicksilver/Quicksilver/runs/5046062463>

```
 codesign --verify --deep --strict --verbose=1 Quicksilver.app
Quicksilver.app: valid on disk
Quicksilver.app: satisfies its Designated Requirement
+ spctl --assess --verbose --type open --type exec ./Quicksilver.app
./Quicksilver.app: rejected
+ true
+ spctl --assess --verbose --type open --context context:primary-signature 'Quicksilver 1.6.1.dmg'
Quicksilver 1.6.1.dmg: rejected
```

Using the artifact from <https://github.com/quicksilver/Quicksilver/actions/runs/1787301564>:

Trying to open the DMG:

Screen Shot 2022-02-03 at 08 49 01

Right click -> Open -> double click `Quicksilver.app`

Screen Shot 2022-02-03 at 08 49 32

Right click `Quicksilver.app` and open and it runs as expected.

So the developer certficate still hasn't made `spctl` happy.

> Next steps?

Tinkering with `spctl` locally to see if we can get both of these happy:

- `spctl --assess --verbose --type open --type exec ./Quicksilver.app`
- `spctl --assess --verbose --type open --context context:primary-signature 'Quicksilver 1.6.1.dmg'`

I hadn't been able to get them to both work, and I had assumed it was because I don't have a valid developer certificate, so @skurfer was kind enought to put his into GitHub secrets; I'd hoped the GA would "just work" but clearly not.

So next step is to resolve the seeming discrepancy between the results of these two commands:

```
 codesign --verify --deep --strict --verbose=1 Quicksilver.app
Quicksilver.app: valid on disk
Quicksilver.app: satisfies its Designated Requirement
+ spctl --assess --verbose --type open --type exec ./Quicksilver.app
./Quicksilver.app: rejected
```

pjrobertson:
Great, thanks for the update. Could it be that the plugins aren't getting signed? Just a thought.

When you're working on this, can you also add one line to qsrelease:

`xcodebuild test -project Quicksilver/Quicksilver.xcodeproj -scheme "Quicksilver Distribution" -destination 'platform=OS X,arch=x86_64' | xcpretty && exit ${PIPESTATUS[0]}`

So that we can also run tests at the same time, as mentioned in #2600. Thanks!

n8henrie:
Continuing at https://github.com/quicksilver/Quicksilver/pull/2647

PR #2588, opened 2021-12-20
Use create-dmg for packaging

gliptak:

https://github.com/create-dmg/create-dmg

https://github.com/quicksilver/Quicksilver/pull/2587
https://github.com/quicksilver/Quicksilver/issues/2576


n8henrie:
I think this might benefit from some discussion (usually in an issue) first -- sorry if I missed a conversation elsewhere. My general practice is to start an issue for conversation *before* embarking on any work towards a pull request, especially anything non-trivial.

In this case, it would probably be worth considering a few things:
- is this addressing a problem / bug, or adding a feature?
- what are the advantages and disadvantages of this solution?
- what alternatives might we consider?

Specifically, I don't known Perl, so something in a language I'm familiar with is appealing, but the current solution seems to work okay -- what makes it worth switching?

I just did a nontrivial amount making qsrelease POSIX, but this is full of Bashisms -- now that zsh is default on MacOS should this be reconsidered?

The shebang will lead to issues in users using homebrew bash vs MacOS bash.

There is a similar Python project I was looking at, but decided to not fix something that wasn't broke. If we want to replace the Perl, IMO Python seems like a better option than bash for nontrivial scripts.

gliptak:
@n8henrie this targets to address the `No space left on device` failure hence the linked issues from above. `create-dmg` has 1.3k stars so it is likely a better place than the current packaging script.

n8henrie:
> this targets to address the No space left on device failure 

Right, I figured so.

- Is replacing the script better than manually specifying a larger file size (which seems to work)?
- How does this script estimate size? Is it more accurate or does it suffer from the same issue? [0] 

> 1.3k stars so it is likely a better place than the current packaging script

Perhaps... though buildDMG.pl is actually hosted from apple.com, which might be more "proof" than 1.3k stars to some? <https://opensource.apple.com/source/AppleUSBCDCDriver/AppleUSBCDCDriver-4201.2.2/Package/buildDMG.pl.auto.html>

[0] Looks like it also has a somewhat fudgy process: https://github.com/create-dmg/create-dmg/blob/de4e0af87b32c982680af97e9c596b7aa2b9e4f6/create-dmg#L297

gliptak:
Thank you for the pointer, I actually couldn't find that original source for buildDMG.pl (I'm a bit surprised that a copyright by a person in 2004 script is considered "official" Apple tooling ...)

Please close the PR if not in line with preferred direction. I'm simply trying to move forward publishing new signed DMGs

n8henrie:
> Please close the PR if not in line with preferred direction

I'm not sure whether or not it's a good idea -- it may be, I've been inclined to look for a replacement just because I don't perl -- but I think my default would be to leave things as is unless the answers to the questions I posed above suggest that this would be an improvement.

pjrobertson:
I'm open to this, assuming that `create-dmg` does actually fix an underlying issue. @gliptak - can you reproduce the `No space left on device` issue, and does this fix it?

If so, we can merge it in - but please add `create-dmg` as a submodule, not directly copy/paste it over.

n8henrie:
I think @skurfer fixed "no space left":  https://github.com/quicksilver/Quicksilver/pull/2582#issuecomment-990463440 I haven't seen it since then.

The fix is a little kludgy, but the approach taken by create-dmg doesn't seem [any less kludgy](https://github.com/create-dmg/create-dmg/blob/de4e0af87b32c982680af97e9c596b7aa2b9e4f6/create-dmg#L297):

```bash
# Add extra space for additional resources
DISK_IMAGE_SIZE=$(expr $DISK_IMAGE_SIZE + 20)
```

https://github.com/create-dmg/create-dmg#alternatives also lists a Python option, but it looks like it will soon be looking for a maintainer: https://github.com/al45tair/dmgbuild . While I'd generally prefer anything more than trivial complexity be in Python rather than bash, the `create-dmg` codebase looks pretty well commented and maintained.

I'm not sure that there's a problem that needs fixing quite yet, but I'm all for removing perl we don't need :)

pjrobertson:
> @pjrobertson for the submodule approach, do you prefer a `pip install` https://pypi.org/project/dmgbuild/ ?

Normally, we'd do a `git submodule add`, but I'd say to hold on for now. If @n8henrie is right that Rob's already merged a fix, then we don't need to merge this after all.

Thanks for the pull request though @gliptak !

n8henrie:
Yes, the fix is the `-volSize 50` currently in `master`'s `qsrelease`.

PR #2590, opened 2022-01-30
Replace git:// URLs with https

l2dy:
GitHub is removing the unencrypted Git protocol, see https://github.blog/2021-09-01-improving-git-protocol-security-github/.


pjrobertson:
Looks good, thanks for your first contribution to Quicksilver!

PR #2600, opened 2022-02-04
Use github actions to run unit tests

### LINKED ISSUE #1852: XCTest Unit Tests failing on Jenkins
Initially this was thought to be a UI problem, but I don't think it is.

It may well be a problem with XCTest in general (which I recommend we change to for all unit tests in the near future anyway).

Here's some useful links:
http://dev-blog.xoom.com/2014/02/15/migrating-ios-continuous-integration-builds-unit-tests-xcode-5/
https://issues.jenkins-ci.org/browse/JENKINS-19955?page=com.atlassian.jira.plugin.system.issuetabpanels:all-tabpanel

I've just tried removing some of the possibly problematic code from the Unit Test I created, but Jenkins is still hanging: http://jenkins.henningj.com:8080/job/Quicksilver/77/console (sorry @HenningJ )

pjrobertson:
Fixes #1852

Hopefully this PR will trigger two GH actions - the existing one and the new unit tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: test
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+  pull_request:
+
+jobs:
+  build:
+    runs-on: macOS-latest
+    env:
+      SIGNING_IDENTITY: "-"
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+    - name: Select Xcode
+      run: sudo xcode-select -s "/Applications/Xcode_13.1.app" && /usr/bin/xcodebuild -version
+    - name: Run unit tests
+      run: |
+           cd Quicksilver
+           ./Tools/qstest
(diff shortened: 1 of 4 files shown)


pjrobertson:
Hmm, can't get the test phase to build properly.

Perhaps instead @n8henrie - as you're working on fixing up qsrelease, can you also add this line after the existing build phase has been complete?

`xcodebuild test -project Quicksilver/Quicksilver.xcodeproj -scheme "Quicksilver Distribution" -destination 'platform=OS X,arch=x86_64' | xcpretty && exit ${PIPESTATUS[0]}`

I'll close this now

n8henrie:
@pjrobertson 

It currently has this before the build phase, `|| true` because it's not passing. Is that adequate?

```
xcodebuild \
    -project Quicksilver/Quicksilver.xcodeproj \
    -destination "platform=macOS,arch=${CURRENT_ARCH}" \
    -scheme 'Quicksilver Distribution' \
    test || true
```

https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/Tools/qsrelease#L47

pjrobertson:
That doesn't run the tests. To run the tests, it needs `xcodebuild test` - the `test` word in there.

This PR almost got things working, I'll re-open so you can see what the issues were

pjrobertson:
Oh wait, I missed the `test` at the end of that command - so you are running the tests. In that case, just need to figure out the dependency problem.

pjrobertson:
Closing in favour of #2632

Issue #330, opened 2011-05-22
Watch directories for modification


fheckl:
It may be desirable to have QS watch file system directories for modification to trigger a catalog update for that directory instead of forcing the user to invoke a rescan.
It was among the things I missed most prominently when switching from LaunchBar. E.g. when you install a new application and then want to do something with that  application you now have to wait until QS has picked it up into its catalog after some time.

pjrobertson:
Adding an option in the catalog prefs to 'watch this folder for modifications' I think would be a great feature to add, and shouldn't be too much code.

tiennou:
IIRC, we need to wrap some of the lower-level fs primitives with something that better fits the bill (as per the points made by @daniels220 above). For example, having a "watch" checkbox and a "depth" that don't care about each other feels to me like good UX.

Aaand I have something about that too ;-).

PR #2601, opened 2022-02-10
Remove QSGlossyBarView

pjrobertson:
It's unlikely that the QSGlossyBarView causes the problems in #2537, but it's possible. And anyway, the 'glossy' look in the prefs is kind of 2000s. 2020 is much 'flatter'.

I've taken the liberty of completely removing this 'dead wood' (I checked, the class doesn't appear to be used by any other QS plugin), and then updating the QS prefs buttons a little to make them more modern.

Take a look at the icons in Catalog/Trigger/Plugins prefs.

Screenshot 2022-02-10 at 16 44 59
Screenshot 2022-02-10 at 16 45 12


pjrobertson:
One final commit to add a spinner and disable the 'check now' button in the prefs if it's already been clicked (and QS is already checking for updates).

Right now, I can click. the 'check now' button multiple times, which will lead to a race condition and potential crash in QS. Much better to avoid this. See below:

Screenshot 2022-02-10 at 18 05 41

pjrobertson:
Once this is merged, I'll work on #330 - want to avoid conflicts first.

skurfer:
Looks good. There are some small conflicts. Once those are fixed, feel free to merge.

pjrobertson:
Great, thanks. Done

PR #2632, opened 2022-02-12
Test Github's built-in Xcode action

pjrobertson:
This action is a standard template from GitHub. Creating this PR to see if it builds and tests successfully.

### DIFF
--- a/.github/workflows/objective-c-xcode.yml
+++ b/.github/workflows/objective-c-xcode.yml
@@ -1,0 +1,33 @@
+name: Xcode - Build and Analyze
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: Build and analyse default scheme using xcodebuild command
+    runs-on: macos-latest
+    defaults:
+      run:
+        working-directory: Quicksilver
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


n8henrie:
Just to mention, I've found e.g. https://github.com/nektos/act to be a bit of a pain to use for local simulation of GA, so IMO it is worthwhile trying to keep the GA build / tests suite as close as possible to what we use locally.

Also, keep in mind that [GA consumes its free minutes relatively quickly on MacOS](https://github.com/quicksilver/Quicksilver/pull/2574#issuecomment-963640820), which is another drawback to debugging GA *without* using something like `act`.

pjrobertson:
Ah, right:

**GitHub Actions usage is free for both public repositories and self-hosted runners**

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

Limits only apply to private repos. It's free for public projects, so go nuts :)

n8henrie:
> Not a whole lot to see here, except:

Sounds good. Going forward, I think it will pay dividends to ensure the two paths don't eventually diverge too much.

> Limits only apply to private repos. It's free for public projects, so go nuts :)

Good catch, I'd missed that! (Looks like I'm [not the only one](https://github.community/t/are-macos-actions-free-for-public-repositories/167315/4).) Excellent news!

PR #2633, opened 2022-02-13
Allow user to automatically folders of custom catalog entries

### LINKED ISSUE #330: Watch directories for modification

pjrobertson:
Fixes #330

I forgot how much of a mess QSLibrarian/QSCatalogEntry actually is :/ This works, but to implement it 'nicely' would require completely rewriting QSLibrarian. Not happening any time soon.

Added new option to sidebar for files & folder catalog type:
Screenshot 2022-02-13 at 14 12 20

To test:

1. Create a new folder catalog entry (e.g. for 'Desktop')
2. Enable the "Automatically update when files change" option
3. Keep your Quicksilver catalog prefs open so you can see the 'count' badge next to your new entry from 1.
4. Create a new file in the folder from 1 (e.g. 'Desktop')
5. Confirm number goes up/down

https://user-images.githubusercontent.com/150431/153741325-b5b10e71-e78e-4818-9f48-c9ce28b0c587.mov

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSCatalogEntry.m
+++ b/Quicksilver/Code-QuickStepCore/QSCatalogEntry.m
@@ -65,6 +65,7 @@ NSString *const QSCatalogEntryInvalidatedNotification = @"QSCatalogEntryInvalida
     return keyPaths;
 }
 
+
 + (instancetype)entryWithDictionary:(NSDictionary *)dict {
 	return [[QSCatalogEntry alloc] initWithDictionary:dict];
 }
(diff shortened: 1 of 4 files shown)


pjrobertson:
I'm thinking about this in hindsight: computers are way more powerful now. Should we just do away with the manual rescan for file object sources, and just make *all* of them update automatically? Or is that asking too much, if someone accidentally adds a folder with a gazillion sub-folders?

n8henrie:
> Or is that asking too much, if someone accidentally adds a folder with a
gazillion sub-folders?

The depth limit should still work, right? So this would only be an issue in
the case of a gazillion files in a flat structure?

skurfer:
> Should we just do away with the manual rescan for file object sources, and just make _all_ of them update automatically?

The last time someone asked about that, I heard there were limits in the OS for how many files a process was allowed to watch, but I don’t know if that’s still true or what the limit might be.

pjrobertson:
If this works, I say let's merge it for now. Automatically update *all* files/folders can come in another update, down the line ;-)

Issue #1235, opened 2012-11-21
Current search text invisible when QS is set to dark colors


daniels220:
The search text that is shown in the header of the results dropdown is basically invisible when QS's colors are set to something dark, because I can't change the color of the text. There's a box in the interface that looks like it should change that text, but it's unclickable. See [this screenshot](http://i.imgur.com/3Pwgq.png).

PR #2635, opened 2022-02-13
Remove QSBackgroundView

pjrobertson:
In the same light as removing QSGlossyBar, this removes QSBackground view, and its custom 'fancy' drawing.

**Edit:** removing the view was a bit premature, it is actually used by a couple of plugins. Instead, I just removed the code draws the gradient backgrounds.

It makes the Results Window more 'flat' and hopefully more 'modern':
Screenshot 2022-02-13 at 21 45 51

I also properly bound the results view header/footer text colour to the appearance prefs, fixing #1235

I'm a little bit more confident that these changes may help with the Catalina/Big Sur colour problems. Less archaic code is good, right?! :D

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/QSBackgroundView.m
+++ b/Quicksilver/Code-QuickStepInterface/QSBackgroundView.m
@@ -44,14 +44,8 @@
 		[cornerEraser appendBezierPathWithRoundedRectangle:[self frame] withRadius:4];
 		[cornerEraser addClip];
 	}
-	NSRect topRect, bottomRect;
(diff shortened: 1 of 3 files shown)


skurfer:
The contrast isn’t ideal depending on the chosen colors, but it’s still a big improvement.

pjrobertson:
> The contrast isn’t ideal depending on the chosen colors, but it’s still a big improvement.

Agreed. I'll take a look at tweaking the Default QS colours. It may make sense to spruce up QS's colour palette based on the new logo anyway.

pjrobertson:
@skurfer  - if you're updating the CHANGELOG - make sure to `pull --rebase` first - I've just updated it

PR #2637, opened 2022-02-15
Fix PYTHONPATH issues with markdown instructions

n8henrie:
[See](See) also: https://github.com/Python-Markdown/markdown/issues/884

Python can't find the extensions when vendored like this; this change
seems to resolve the issue.

Additionally, it looks like the `extra` package may be deprecated.


pjrobertson:
Looks good.

Issue #2640, opened 2022-02-16
Plugins have hex or float `CFBundleVersion`s


n8henrie:
It looks like `CFBundleVersion`s are [supposed to be `[MAJOR].[MINOR].[PATCH]` formatted](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion?language=objc), but currently the majority are using hex:

```console
$ awk '/CFBundleVersion/ { getline; split($0, arr, /[<>]/); print arr[3] }' **/Info.plist  | head
1C
1.0
12C
25
287
A9
47
45
2
120C
```

Further, `bltrversion` assumes hex and chokes on formats that can't be interpreted as hex: <https://github.com/quicksilver/Quicksilver/blob/9300d8893504ff06debfc61da1fbfe6bf54c4479/Quicksilver/Tools/bltrversion#L38>

For the moment, I propose that we:
- if ambiguous assume hex and increment by 1, which preserves backwards compatibility
- fallback to `[MAJOR].[MINOR](.[PATCH])?` if there are any `.`
    - auto-increment only the rightmost part

Sound OK?

pjrobertson:
Makes sense to me. I hate the hex numbers. But I guess they were considered easier to increment. It looks like your PR chooses to just increment the [PATCH] setting?

As much as I'd like to get rid of the hex strings, it'll be a fair amount of testing work because:

* Server handling - I'm pretty sure the qs0.qsapp.com server currently only handles Hex versions
* In-app updating/checking. The app right now compares version numbers based on hex. We'd need to edit the code to compare based on X.Y.Z format. Then any older QS versions that don't have this code would choke if they got an X.Y.Z format from the update server, but expected a hex.

skurfer:
As part of re-building all of the plug-ins, I started ripping out `bltrversion`. It was failing and has maybe always been more trouble than it’s worth because it took so much fiddling.

1. You would often have to manually adjust the version to avoid bumping it by 50 instead of 1.
2. Xcode changed the order it did things, so you would end up building _before_ the script ran. This meant that the plug-in you built was still using the old version, and that the version in the repo didn’t match the built product. You could get around it by building again, then manually putting the version back 1.

I think it has value as a documentation generator if we can get it working again (and get it to run _before_ the bundle is created).

n8henrie:
> It looks like your PR chooses to just increment the [PATCH] setting?

[...]

PR #2641, opened 2022-02-16
Tolerate CFBundleVersions in [MAJOR].[MINOR].[PATCH] format

n8henrie:
Fixes https://github.com/quicksilver/Quicksilver/issues/2640

### DIFF
--- a/Quicksilver/Tools/bltrversion
+++ b/Quicksilver/Tools/bltrversion
@@ -35,9 +35,21 @@ if qs_build >= min_build:
         info['QSRequirements'] = {'version': min_build_string}
 
 ## increment version
-version_hex = info['CFBundleVersion']
-version_int = int(version_hex, 16)
-new_version = hex(version_int + 1)[2:].upper()  # uppercase, without 0x prefix
+version = info['CFBundleVersion']
+
+try:
+    version_int = int(version, 16)
+    new_version = hex(version_int + 1)[
+        2:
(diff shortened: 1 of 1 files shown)


pjrobertson:
Based on the chat over at #2640 it looks like we won't merge this for the time being. Sorry Nate! :'(

I do like the idea of the X.Y.Z, but I don't think the risks of breaking our update server and having to change all plugins is worth it at the moment.

n8henrie:
No worries! Though I want to make sure it's clear that this PR didn't make any moves to `X.Y.Z`, it just fixes an existing problem where a few plugins *already use* `X.Y.Z` and are incompatible with `bltrversion`:

```console
$ awk '/CFBundleVersion/ { getline; if ( $0 ~ /\./) { print FILENAME "\n" $0 "\n" } }' QSPlugins/*/Info.plist
QSPlugins/XattrMetadataImporter-qsplugin/Info.plist
        <string>1.2</string>

QSPlugins/com.facebook-qsplugin/Info.plist
        <string>1.2</string>
```

But we should probably just fix these few instead of the script -- I had thought it was more prevalent of an issue because I initially recursively looked at `**/Info.plist` and picked up those of many sub-dependencies. My bad!

pjrobertson:
Right, that clarification is important.

If there are plugins using the X.Y.Z format, then they should be changed to a hex format, otherwise they’ll break when we upload them to the update server.

For those plugins, go ahead and make the changes then submit PRs!

PR #2642, opened 2022-02-16
Also clear text when clearing object view

### LINKED ISSUE #2289: Text entry text isn’t cleared
To reproduce:

1. Hit `.` to enter text entry mode
2. Type something
3. Hit ⎋ to exit text entry mode
4. Search for something else
5. Press ⌫ until the first pane is empty (showing “Type to search”)
6. Hit `.` again to enter text entry mode

Expected: There will be no text
Actual: The previously typed text is still there

In other words, it sounds like that text is getting replaced when appropriate, but never cleared.

pjrobertson:
Fixes #2289

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.h
+++ b/Quicksilver/Code-QuickStepInterface/QSSearchObjectView.h
@@ -90,6 +90,7 @@ typedef NS_ENUM(NSUInteger, QSSearchMode) {
 @property (strong) NSTextView *textModeEditor;
 
 - (void)clearSearch; // reset everything and be ready for a new search
+- (void)clearTextView; // reset the text view. Not to be confused with clearSearch - clear search is called every time a new object is presented into the SOV, whereas keeping the underlying text is important in these cases. clearTextView should only be used in `deleteBackwards` style situations
 
 - (void)clearAll;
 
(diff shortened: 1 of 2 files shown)


skurfer:
I’m still able to reproduce the original issue with this change.

It seems to me like the text mode string should be cleared as soon as you initiate a search and something else gets selected. Or maybe the current behavior is fine and we shouldn’t worry about it? 🤷🏻‍♂️

pjrobertson:
Hmm... am I following the right steps? See below:

https://user-images.githubusercontent.com/150431/154369571-5677edc1-fc8f-4947-9995-7a8e3b48bd98.mov

skurfer:
So, I noticed something strange in your video. As you hit Delete, the selection changes. I had never seen that happen, then I remembered you’re a _monster_ and use that one-character-at-a-time delete preference. 😜

If I disable double-Delete to clear the selection, I see what you see. So the behavior must be slightly different for that last keypress depending on that setting. I can look into it.

pjrobertson:
You caught me!
Screenshot 2022-02-23 at 11 19 41

> Maybe we should clear the text string inside clearSearch instead?

Should be a simple enough fix, yes. I'll look into it and update this PR.

skurfer:
That might be too heavy handed. It’s called from 11 places. Maybe we should create a `clearSearchAndText` that calls `clearSearch` and adds your new code? Or just repeat the code in that method.

pjrobertson:
OK, so you were right that clearing the text string in `clearSearch` was too heavy handed. I just tried that and it broke everything. That's because `clearSearch` is called every time you type *anything* into the search field (it's more akin to 'reset and get ready to redo another search').

I cleared the string in the 3 places I can think it makes sense:

1. Deleting backwards with 'Pressing the Delete key clears the whole search string' off (my case)
2. Deleting backwards with 'Pressing the Delete key clears the whole search string' on (Rob's case)
3. Pressing ⌃U to clear the search object

I think that's good enough for now :)

PR #2643, opened 2022-02-16
update credits

skurfer:
Mainly to add @n8henrie


pjrobertson:
Fully agree! Merged 😄

PR #2644, opened 2022-02-17
Performance tweaks

pjrobertson:
It helps having a 10 year old Mac ;-) I wouldn't say QS is 'slow', but it could definitely be snappier.

These changes are based on profiling using Instruments. Overall searching should be slightly faster, displaying action should be faster.

I propose this *isn't* merged into 2.0.0, but saved for a later release. We've got enough to work on right now.


ernstki:
From one ten-year-old-Mac owner to another, cheers to you, good sir.

Issue #2636, opened 2022-02-15
Sleep no longer selectable


Max-Intosh:
The Sleep function is no longer available to the input.

pjrobertson:
Hi @Max-Intosh  - can you please upload a screenshot of the problem you are facing, and follow the Issue form template, including providing the following information:

**What version of macOS?**

**What version of Quicksilver?**

jeutzsch:
I noticed the same thing yesterday after first updating to **Monterey 12.2.1** and then updating to **QS v1.6.1** on both my computers (MacMini 2018 and MBP 13" 2019).
I'm not sure a screenshot would be useful because this is something that is now missing, so a screenshot wouldn't show anything.
Before updating, opening QS and typing "SLEEP" (actually I had it set to be triggered with "SL" already), a script was run that put the computer into sleep mode. But now this script seems to be missing, and opening QS and typing "SLEEP" just doesn't result in anything that will put the computer to sleep.
It looks to me like it was/is part of the ["Extra Scripts" plugin](https://qsapp.com/manual/System/#extra-scripts).

pjrobertson:
Thanks for the clarification.

OK, so this is an issue with the 'sleep' script from the ExtraScripts plugin no longer being present. Likely to do with the latest release of the plugin @skurfer 

It seems that the Script files (which are comiled AppleScripts) are no longer being picked up as such by Quicksilver. Not sure if a re-build would work, or what's needed

Max-Intosh:
Hi PJ @pjrobertson,

Mac OS 10.14.6 on Macbook Pro (15 inch 2019) with the just updated Quicksilver 1.6.1 (4026).

Can not recall how Quicksilver had been triggering the Sleep function as have had it operation for so many years now.
Birth Restart and Sleep are missing out of the catalogue of text inputs to the Quicksilver window.

In Catalog>Plugins the item "Scripts (QS Extras)" is present (with 24 to the right hand side) and ticked, however the drop down triangle is not present on the left hand side (indicating that no scripts are present?).  Screen shot attached.

In Catalog>Scripts the item "Scripts (All Users)" is present (with 86 to the right hand side) and ticked, however the drop down triangle is not present on the left hand side (indicating that no scripts are present?).  Screen shot attached.

Catalog>Custom is empty.

Hope that that helps.
[...]

PR #2647, opened 2022-02-18
fix code signing

skurfer:
This small change to `qsrelease` gives me a valid application. The DMG is rejected, though. Still trying to figure that one out.


pjrobertson:
Looks good to me. I'm assuming when you say 'gives me a valid application' you mean it passes the test @n8henrie posted over in [this comment](https://github.com/quicksilver/Quicksilver/pull/2587#issuecomment-1029135191)?

`codesign --verify --deep --strict --verbose=1 Quicksilver.app`

 I was investigating code-signing this morning. Trying to get Xcode to automatically codesign.

I figured that the plugins have an `INSTALL_PATH = /Quicksilver.app/Contents/PlugIns` which meant the built-in plugins (Bezel, Core Support etc.) were getting copied *into* the Quicksilver.app bundle after they were built, but weren't being code-signed. That's a problem.
My solution was to delete the line `INSTALL_PATH = /Quicksilver.app/Contents/PlugIns` in QSPlugins.xcconfig, then add a 'Copy Files' script stage that copied the plugins into the App and signed them at the same time. This worked, but caused dependency issues.

I like your approach of just doing a deep codesign manually on the entire package once it's built. Much simpler.

For the DMG, maybe there's something in Create-DMG that was proposed by @gliptak in #2588 that fixes it? Maybe we *should* try using Create-DMG?

```
"${SOURCE_ROOT}"/Tools/create-dmg \
  -volname Quicksilver \
  -volicon "${SOURCE_ROOT}"/Resources/Images/QuicksilverDMG.icns \
  -disk-image-size 50 \
  Quicksilver \
```

pjrobertson:
This might help: https://lessons.livecode.com/m/4071/l/1120307-codesigning-and-notarizing-your-lc-standalone-as-dmg-for-distribution-outside-the-mac-appstore#creating-dmg

If that doesn't work, then maybe you need the `sudo xattr -cr "<path_to_app_bundle>"` from step  2?

pjrobertson:
Just came across this thread that recommends against using `--deep` :( Not sure if thats why the DMG doesn't work properly. https://developer.apple.com/forums/thread/129980

n8henrie:
> This small change to qsrelease gives me a valid application. The DMG is rejected, though. Still trying to figure that one out.

@skurfer What are you seeing that's different?

[CI from this PR](https://github.com/quicksilver/Quicksilver/runs/5247027750?check_suite_focus=true):

```
+ codesign --verify --deep --strict --verbose=1 Quicksilver.app
Quicksilver.app: valid on disk
Quicksilver.app: satisfies its Designated Requirement
+ spctl --assess --verbose --type open --type exec ./Quicksilver.app
./Quicksilver.app: rejected
+ true
+ spctl --assess --verbose --type open --context context:primary-signature 'Quicksilver 1.6.1.dmg'
Quicksilver 1.6.1.dmg: rejected
```

[CI from before the change](https://github.com/quicksilver/Quicksilver/runs/5219420954?check_suite_focus=true):

```
+ codesign --verify --deep --strict --verbose=1 Quicksilver.app
Quicksilver.app: valid on disk
Quicksilver.app: satisfies its Designated Requirement
+ spctl --assess --verbose --type open --type exec ./Quicksilver.app
./Quicksilver.app: rejected
+ true
+ spctl --assess --verbose --type open --context context:primary-signature 'Quicksilver 1.6.1.dmg'
Quicksilver 1.6.1.dmg: rejected
```

> the built-in plugins (Bezel, Core Support etc.) were getting copied into the Quicksilver.app bundle after they were built, but weren't being code-signed.

That's what `find ./Quicksilver.app -exec codesign ...` was for, which this PR removes (which is fine since it still wasn't passing spctl). Come to think of it, perhaps this `find ... --sign` everything followed by a `codesign --force --deep` is a problem? Perhaps  the signatures are different and thereby invalidating something? I'll look again.

> For the DMG, maybe there's something in Create-DMG that was proposed by @gliptak in https://github.com/quicksilver/Quicksilver/pull/2588 that fixes it?

I'll test it out locally, but I've looked into this (when I was trying to figure out what `buildDmg.pl` or whatever it's called did) and I don't think it does anything that would affect codesigning. They're both just wrappers around `hdiutil` et al. that facilitate creation of a dmg without having to go through `Disk Utility`'s GUI.

> `xattr -cr `

I think I'd looked into this and `xattr -lr` didn't show any issues (though I did have to use `xattr -cr` to get ExtraScritps to build in https://github.com/quicksilver/Quicksilver/issues/2636).

skurfer:
Oh, I also generated a new signing cert, but the previous one was valid, so I don’t think that should matter. I’ll try to get the one GitHub uses updated to match, though. Maybe I should export the cert and put it on `qs0` so other can experiment with this?

n8henrie:
> So for the first time in years, spctl is happy.

`spctl --assess --type execute` has not been difficult to make happy locally, I've reliably been able to get this working with a self-signed cert (I think this is what you were referring to), it's just the `--type open` with the DMG that's been problematic. 

- https://github.com/quicksilver/Quicksilver/issues/2576#issuecomment-1037368736
- https://github.com/quicksilver/Quicksilver/pull/2582#issuecomment-973656957

> The DMG fails, but I don’t get any warning when I open it like you do with the ones built by GitHub.

Are you downloading it? Using the same file, `python3 -m http.server` in the parent directory, then download, then open? Something about downloading triggers the warnings.

n8henrie:
I still keep seeing things about needing to [`stapler`](https://github.com/quicksilver/Quicksilver/issues/2576#issuecomment-1013946466) something, I think this is our missing step but will require a valid cert AFAICT.

> Maybe I should export the cert and put it on qs0 so other can experiment with this?

That would be great!

skurfer:
I tried downloading the DMG and still don’t get a warning. 🤷🏻‍♂️

I played around with notarization last night, but it’s not happy. Even with a build that `spctl` accepts.

```json
{
  "logFormatVersion": 1,
  "jobId": "68c7a00c-6889-4871-9aa6-61d661fcbf4a",
  "status": "Invalid",
  "statusSummary": "Archive contains critical validation errors",
  "statusCode": 4000,
  "archiveFilename": "Quicksilver.zip",
  "uploadDate": "2022-02-18T15:29:18.508Z",
  "sha256": "0a35e321e031389b5ea644794dbae3a95c1befd42b97af4e113e00e2d2d06747",
  "ticketContents": null,
  "issues": [
    {
      "severity": "error",
      "code": null,
      "path": "Quicksilver.zip/Quicksilver.app/Contents/MacOS/Quicksilver",
      "message": "The executable does not have the hardened runtime enabled.",
      "docUrl": null,
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Quicksilver.zip/Quicksilver.app/Contents/MacOS/Quicksilver",
      "message": "The executable does not have the hardened runtime enabled.",
      "docUrl": null,
      "architecture": "arm64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Quicksilver.zip/Quicksilver.app/Contents/Resources/QSDroplet.app/Contents/MacOS/QSDroplet",
      "message": "The binary is not signed with a valid Developer ID certificate.",
      "docUrl": null,
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Quicksilver.zip/Quicksilver.app/Contents/Resources/QSDroplet.app/Contents/MacOS/QSDroplet",
      "message": "The signature does not include a secure timestamp.",
      "docUrl": null,
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Quicksilver.zip/Quicksilver.app/Contents/Resources/QSDroplet.app/Contents/MacOS/QSDroplet",
      "message": "The binary is not signed with a valid Developer ID certificate.",
      "docUrl": null,
      "architecture": "arm64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Quicksilver.zip/Quicksilver.app/Contents/Resources/QSDroplet.app/Contents/MacOS/QSDroplet",
      "message": "The signature does not include a secure timestamp.",
      "docUrl": null,
      "architecture": "arm64"
    }
  ]
}
```

n8henrie:
I copied over [the signing action](https://github.com/n8henrie/Quicksilver/blob/3f2395082dbbe239a708a8202f30d1f580248991/.github/workflows/release.yml#L32) from the other PR: https://github.com/quicksilver/Quicksilver/pull/2647

n8henrie:
Continuing from [here](https://github.com/quicksilver/Quicksilver/pull/2587#issuecomment-1029135191), `spctl --type exec` is happy with `Quicksilver.app` on CI for what I think is the first time:

https://github.com/quicksilver/Quicksilver/runs/5250494318?check_suite_focus=true

```
+ spctl --assess --verbose --type open --type exec ./Quicksilver.app
./Quicksilver.app: accepted
source=Developer ID
```

Here's the artifact: <https://github.com/quicksilver/Quicksilver/suites/5363424203/artifacts/168117212>

For me the downloaded `.dmg` opens without warning (!!!).

I still get a warning when trying to run the `.app` -- have to right click -> open.

n8henrie:
@skurfer, can you try the `notarytool` command from <https://github.com/quicksilver/Quicksilver/pull/2647#issuecomment-1044753771> to see if that works for `dmg` notarization?

n8henrie:
One other thought -- once we get this to a satisfactory spot, we should probably delete [this line](https://github.com/n8henrie/Quicksilver/blob/3f2395082dbbe239a708a8202f30d1f580248991/.github/workflows/release.yml#L7) in order to only run the `sign` action on tags and not all PRs; it seems like the secrets themselves [can't be leaked by a malicious PR](https://github.com/quicksilver/Quicksilver/issues/2576#issuecomment-974704757), but I imagine someone could submit a PR to build a malicious copy of the app carrying @skurfer's signature, which could jeopardize your good standing with Apple.

skurfer:
I have been trying `xcrun notarytool submit` with my dev account. The JSON output above comes from running `xcrun notarytool log` after it’s found invalid. I was just using a ZIP of the application since the DMS wasn’t signed correctly and would have surely been rejected. I did try the DMG just to see and got basically the same result.

skurfer:
> Continuing from https://github.com/quicksilver/Quicksilver/pull/2587#issuecomment-1029135191, spctl --type exec is happy with Quicksilver.app on CI for what I think is the first time:
> 
> EDIT: To clarify, I think this means that @skurfer set up the organization secrets correctly for signing with the cert. Huzzah!

Did you update the secrets with my new cert? I wonder why it would suddenly start working if it didn’t work when I initially opened this PR.

n8henrie:
> Did you update the secrets with my new cert?

No, I don't seem to have access to manage secrets here (the repo `Settings` tab that I see on my personal repos just doesn't appear).

> I wonder why it would suddenly start working if it didn’t work when I initially opened this PR.

This branch didn't have the signing action, my old one did, so I [copied it over](https://github.com/quicksilver/Quicksilver/pull/2647#issuecomment-1044791138) this morning.

n8henrie:
>       "message": "The signature does not include a secure timestamp.",

Added timestamp to the `.dmg` codesign

skurfer:
> Was this on a release or debug build? Currently looks like we have hardened runtime enabled only on release builds.

That was after running `qsrelease`, so it should be a release build. It must be partially respecting that setting because before I enabled it, there were a lot more files with that error. I’ll look at the individual targets to make sure it’ not being overridden there.

skurfer:
Merging this since it at least fixes the app signing. We can continue to work on the DMG and create a separate PR.

pjrobertson:
Sounds good to me. For a new release - do we *need* to sign the DMG? I've seen reports online that DMGs don't have to be signed, just the app inside. If that's the case, then I suggest we make DMG signing a 2.1.0 target, so that we can get 2.0.0 out asap

n8henrie:
I agree -- I think it's a nuisance to not have the DMG signed, but with
some good instructions should not prevent people from people able to use it.

After opening the DMG, they'll get a warning on first use, will have to
reveal in finder, then right click and open to run the first time.

pjrobertson:
Oh, OK. If not code-signing the DMG still throws up an error for people when opening the DMG, then I think it should be fixed for 2.0.0. It's not ideal releasing 2.0 which is tagged as "Catalina ready" but then letting people deal with this problem.

Right now, other than re-building plugins, I think this is the only thing holding back the 2.0.0 release. Let's try and concentrate our efforts on the DMG fix. I *should* be able to help now that I have access to the signing Certs.

pjrobertson:
Created new corresponding issue: #2654

Issue #2119, opened 2015-10-09
Quicksilver service really needs to be renamed…


devburke:
For years now, Quicksilver has had a service in the OS X services. It's the same entry that allows you to send a selection to QS by pressing cmd-esc. However, it's got an awful, awful name: "Get Current Selection (Do Not Change Shortcut)".

There is so much wrong with that. It's completely un-descriptive of what it does. It doesn't even give any indication that it involves Quicksilver. In addition, putting "(Do Not Change Shortcut)" in the title is really grossly ugly.

Take a look in the services menu:

![screen shot 2015-10-09 at 1 21 46 am](https://cloud.githubusercontent.com/assets/1630645/10387136/637a6228-6e24-11e5-8ee6-544321a58ff0.png)

Why not change it to something like, "Send to Quicksilver"?

Quicksilver has made some wonderful strides in recent years, and this strikes me as one of the most inelegant things left in it.

Just wanted to point it out. Hopefully it'll be remedied in a future update :)

skurfer:
> It's completely un-descriptive of what it does.

Well, not *completely*. :smiley:

> It doesn't even give any indication that it involves Quicksilver.

Good point. That’s a problem.

> In addition, putting "(Do Not Change Shortcut)" in the title is really grossly ugly.

Changing the shortcut breaks the Current Selection proxy object completely, and I think prevents the service from working. But the number of people that will try that is probably pretty low.

devburke:
Touché. Well, I’d say un-descriptive in the context in which you encounter it. From within Quicksilver, it would make sense. But you only encounter it outside of Quicksilver.

That is an interesting problem about changing the shortcut. It’s unfortunate that that breaks the proxy objects. I’d ask why that is, but I think the explanation would go over my head anyway :P

Sesquipedalian:
Agreed. There used to be two functionally identical services, "Send to Quicksilver" and "Get Current Selection (Do Not Change Shortcut)". I used to disable the ugly-named one and just use the pretty one. Now that we only have one service (which makes sense), I would love it if the name were changed to "Send to Quicksilver".

TrevorSayre:
https://github.com/quicksilver/Quicksilver/blob/fe603d1f9811eff21ddc517996f8d9d2e909010b/Quicksilver/PropertyLists/Quicksilver-Info.plist#L230

squm:
how to remove “Send to Quicksilver” ?

Issue #1416, opened 2013-03-04
"Select Current Selection in command windows" hotkey trigger resets all the time


sadhuroura:
I like to assign "Select Current Selection in command windows" trigger to CTRL + ALT + G.  I change the hotkey from CMD + ESC (default one) to my desired one but all the time it returns to its default configuration.

I haven't done further investigation, I don't know when it happens. If I discover something new about this weird and annoying behaviour I'll make you know

Issue #409, opened 2011-07-03
Two separate services (one internal and one not) can be confusing


pjrobertson:
We should try and create a single service, that is clever enough to detect if the service was called by Quicksilver (when CMD+G is pressed or the 'current selection' proxy is used) or if it was called by the user (using the global hotkey or App->Services->Send to Quicksilver.)

I'm not currently aware on the differences between the code for the two services, but it could easily be integrated into a single service with a simple `if ... else ...` statement

In implementing this, we'd need to make sure we do not mess with the user's shortcut key for the 'send to quicksilver' service. This would probably mean making the `invokeSercive` method in QSGlobalSelectionProvider.m dynamically set the keys to press depending on the user's choice.
Or set two hotkeys if this is possible.

PR #2651, opened 2022-02-21
Fix 'Get Current Selection' service menu

### LINKED ISSUE #2646: Get Current Selection Service no longer works
**Describe the bug**

I noticed ⌘⎋ wasn’t working. I checked the Services prefs and saw “033” instead of the expected shortcut. I wiped it and attempted to set it again, but the system doesn’t allow you to use ⎋ in shortcuts. (I don’t remember if it used to.)

**Screenshots**

Screen Shot 2022-02-17 at 9 00 42 AM

**What version of macOS?**

12.2.1

**What version of Quicksilver?**

It’s complicated. 😂

pjrobertson:
This fixes the hotkey to ⌘⎋ and fixes #2646

I've also added another change here to rename the service to "Send to Quicksilver" as per #2119

I send to agree that 'get current selection' is not that descriptive. I also decided to remove the `(Do not change)` part:  it is ugly, and in the small edge case that users change it, they'll find that it breaks and likely just click the 'restore defaults' button in the menu any. 

Still, the fact you cannot change the hotkey, and how to fix it if you do (click 'Restore Defaults') should be documented in the manual.

### DIFF
--- a/Quicksilver/PropertyLists/Quicksilver-Info.plist
+++ b/Quicksilver/PropertyLists/Quicksilver-Info.plist
@@ -207,7 +207,7 @@
 			<key>NSMenuItem</key>
 			<dict>
 				<key>default</key>
-				<string>Get Current Selection (Do Not Change Shortcut)</string>
+				<string>Send to Quicksilver</string>
 			</dict>
 			<key>NSMessage</key>
 			<string>getSelection</string>
--- a/Quicksilver/PropertyLists/Quicksilver-Info.plist
+++ b/Quicksilver/PropertyLists/Quicksilver-Info.plist
@@ -202,7 +202,7 @@
 			<key>NSKeyEquivalent</key>
 			<dict>
 				<key>default</key>
-				<string>\033</string>
+				<string>⎋</string>
 			</dict>
 			<key>NSMenuItem</key>
 			<dict>


pjrobertson:
Oops, I accidentally just pushed these changes to master. This PR is more to discuss and confirm that we're ok with renaming the service to 'Send to Quicksilver'

n8henrie:
SGTM, `Send to Quicksilver` seems more parallel to what other apps use.

skurfer (approved):

skurfer:
Looks good. So should we just close this?

pjrobertson:
There was one more commit I didn't push to master (this one [5b6b2e0](https://github.com/quicksilver/Quicksilver/pull/2651/commits/5b6b2e0e8349c29e10bf97166e931e7f7d662784)) so I've done that now

skurfer:
I’m seeing a similar problem with the **Command Window with Selection** trigger. That one uses a key code and has the correct value, but it appears as ⌘5. And it’s not truncating from 53. I discovered this because ⌘5 was invoking QS. I’m not sure what the fix is, but I also don’t remember why we even need that trigger. I disabled it and ⌘⎋ still works as expected in every app.

pjrobertson:
Hmm.... I do see a problem with setting a trigger that uses ⎋.

E.g. if I set a trigger to ⌥⎋ it appears as ⌥5 and reacts to ⌥5.

RE the 'Command Window with Selection', I found this in the Changelog for v1.0.0 (9 years ago!)

Simplify Services and set the "Command Window with Selection" trigger to use ⌘⎋ by default (#409, #1295, #1416, #1418)

I will create a new issue for this and look into it

pjrobertson:
@skurfer - please try the latest master to see if that fixes the 'Command Window with Trigger' issue.

Oh, and to answer this:
> but I also don’t remember why we even need that trigger

You can see your own comment [here](https://github.com/quicksilver/Quicksilver/pull/1295#issuecomment-11463027), hehehe ;-)

Issue #2654, opened 2022-02-23
Quicksilver is not codesigned


pjrobertson:
For previous discussions, see #2555 

Quicksilver is currently not codesigned on release, causing issues with users opening it on their computer.

Attempts were made to fix codesigning in #2647, however those changes only codesign the *app* and not the DMG. Currently outstanding is codesigning the DMG.

This is the main blocking issue before we can get 2.0 out.

n8henrie:
I'd like to continue working on this.

```console
$ xcrun notarytool submit Quicksilver\ 1.6.1.dmg --wait --team-id '...' --apple-id '...' --password '...'
Conducting pre-submission checks for Quicksilver 1.6.1.dmg and initiating connection to the Apple notary service...
Error: HTTP status code: 403. Invalid or inaccessible developer team ID for the provided Apple ID. Ensure the Team ID is correct and that you are a member of that team.
```

@skurfer would you add me to your Apple "team," which should allow me to sign using my own ID and password? 

Based on <https://help.apple.com/developer-account/#/dev3e8818774> it should be:

<https://developer.apple.com/account> ->  `People` -> `Manage Users`; my name and Apple ID from email address we've used for correspondence.

skurfer:
I don’t see “People”. Maybe because I have an individual membership and not a business?

n8henrie:
Ah, that makes sense.

> If you’re enrolled in the **Apple Developer Enterprise Program**...

n8henrie:
WOOHOO! Got it.

n8henrie:
What a relief. 

@skurfer @pjrobertson can you guys take a look at <https://github.com/quicksilver/Quicksilver/actions/runs/1913370728>?

I think the hard part (of signing) is done, but I'm sure there are a few more tweaks before merging. Specifically, I'm going to do a little tinkering tomorrow with (fake) repo secrets and convince myself of the security here. Again, I think limiting the `qssign` step to tags instead of all pull requests should go a long way here.

Also, I'd like some feedback on <https://github.com/quicksilver/Quicksilver/pull/2655#issuecomment-1054379715> -- specifically I think we'll probably want to continue the current logic for naming the DMG (the version part), but I think we'll likely need to *manually* keep track of the `git tag` to have it properly named in the file uploaded to releases. Will tinker with this as well.

Good progress though, I'm psyched.

skurfer:
[...]

Issue #2583, opened 2021-11-13
Tests fail with cycle in dependencies


n8henrie:
Current `qsrelease`:

```
Testing failed:
        Cycle in dependencies between targets 'Core Support Tests' and 'Bezel Interface'; building could produce unreliable results.
Cycle path: Core Support Tests → Core Support → Core Support Tests → Quicksilver → Quicksilver Tests → Core Support Tests → Bezel Interface → Core Support Tests
Cycle details:
→ Target 'Core Support Tests' has a command with output '/tmp/QS/build/Release/Core Support Tests.xctest/Contents/MacOS/Core Support Tests'
○ Target 'Core Support Tests' has link command with output '/tmp/QS/build/Quicksilver.build/Release/Core Support Tests.build/Objects-normal/arm64/Binary/Core Support Tests'
○ Target 'Core Support Tests' has target dependency on Target 'Core Support'
→ Target 'Core Support' has compile command with input '/Users/n8henrie/git/Quicksilver/Quicksilver/Nibs/QSUserDefinedProxySource.xib'
○ Target 'Core Support' has symlink command from '/tmp/QS/build/Release/Core Support.qsplugin' to '/tmp/QS/build/Release/Quicksilver.app/Contents/PlugIns/Core Support.qsplugin'
→ Target 'Core Support Tests' has target dependency on Target 'Quicksilver'
→ Target 'Quicksilver': CodeSign /tmp/QS/build/Release/Quicksilver.app
→ Target 'Quicksilver Tests' has create directory command with output '/tmp/QS/build/Release/Quicksilver.app/Contents/PlugIns/Quicksilver Tests.xctest'
→ Target 'Core Support Tests' has target dependency on Target 'Bezel Interface'
→ Target 'Bezel Interface' has symlink command from '/tmp/QS/build/Release/Bezel Interface.qsplugin' to '/tmp/QS/build/Release/Quicksilver.app/Contents/PlugIns/Bezel Interface.qsplugin'
```

pjrobertson:
@n8henrie - looks like we have multiple issues/PRs on this qsrelease business. Can we close this one as well?

n8henrie:
I'll take another look, but I think this is still active. The other `qsrelease` tasks were mostly about modernizing the scripts, I haven't delved into how testing works with Quicksilver (or Xcode in general, for that matter), and I think this is more related to the latter.

n8henrie:
Issue is still there:

```
[...]

PR #2655, opened 2022-02-23
Split signing out from qsrelease

n8henrie:
Issues addressed:

- Provides separate GitHub Actions for building and signing (https://github.com/quicksilver/Quicksilver/issues/2583#issuecomment-1035702305)
- Provides separate script for debugging signing so you don't have to rebuild every time (requires exporting a few variables normally set in `qsrelease`
- By default will still build *and* sign (for local builds) unless `QS_BUILD_ONLY` is set -- preserves current behavior
- Uses GitHub Actions' artifacts to avoid re-building the entire project twice
- Removes the "arbitrary volume size and hope it's big enough" workaround
- Adds what I think should be the necessary changes for automatic
  notarization of the DMG

Other changes:

- Removes need for `buildDMG.pl` with no new dependencies
- Reorders test *after* build, since the tests depend on `/tmp/QS/Configuration/Quicksilver.pch`

TODO: 
- @n8henrie Ensure no secrets are exposed on logs (I think the keychain approachshould work)
- @skurfer Update the fake envvars in GA with a real ID and app-specific pass (I'm happy to do this if you're able to add me to the team as per <https://github.com/quicksilver/Quicksilver/issues/2654#issuecomment-1048970733>)

Fixes https://github.com/quicksilver/Quicksilver/issues/2654 -- hopefully


n8henrie:
https://github.com/actions/upload-artifact#zipped-artifact-downloads

> One of the consequences of this limitation is that if a zip is uploaded during a workflow run and then downloaded from the UI, there will be a double zip created.

A little annoying.

n8henrie:
Blocked on https://github.com/actions/upload-artifact/issues/300

n8henrie:
OK, I think this is as ready as I can get it for now. I've squashed and rebased.

- I'm blocked on debugging locally unless @skurfer can add me to the apple "team": <https://github.com/quicksilver/Quicksilver/issues/2654#issuecomment-1048970733>
- Whether or not I'm added to the team for local debugging, I don't have access to manage the repo's GitHub Secrets. In addition to the 3 [added here](https://github.com/quicksilver/Quicksilver/issues/2576#issuecomment-988454102), we will need the 4 listed [here](https://github.com/quicksilver/Quicksilver/pull/2655/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR48):
    - `SIGNING_IDENTITY`
    - `NOTARIZING_ID`
    - `NOTARIZING_PASS`
    - `NOTARIZING_TEAM_ID`

n8henrie:
Actually, the `TEAM_ID` and `SIGNING_IDENTITY` can be the same thing (I think), as long as it's set to the `1234ABCD` below, so just 3 secrets to add:

```
$ security find-identity -p basic -v
"Developer ID Application: Your Name (1234ABCD)"
```

pjrobertson:
Looks ike build / sign is failing:

`Error: Team ID must be at least 3 characters`

n8henrie:
Correct -- the 3 secrets [listed above](https://github.com/quicksilver/Quicksilver/pull/2655#issuecomment-1050221913) and [I don't have access](https://github.com/quicksilver/Quicksilver/pull/2655#issuecomment-1050220064) to modify repo secrets.

This is blocked pending @skurfer's review and feedback, since it's his Apple ID and cert.

skurfer:
I’ve made you an Owner on the Quicksilver org @n8henrie, so you should be able to make whatever changes are necessary now.

skurfer:
I’ve added the secrets and updated the signing cert to match the one I generated recently.

n8henrie:
I'm making some progress here.

Current errors with notarization:

```json
{
  "severity": "error",
  "code": null,
  "path": "Quicksilver 1.6.1.dmg/Quicksilver.app/Contents/Resources/QSDroplet.app/Contents/MacOS/QSDroplet",
  "message": "The binary is not signed with a valid Developer ID certificate.",
  "docUrl": null,
  "architecture": "x86_64"
},
{
  "severity": "error",
  "code": null,
  "path": "Quicksilver 1.6.1.dmg/Quicksilver.app/Contents/Resources/QSDroplet.app/Contents/MacOS/QSDroplet",
  "message": "The signature does not include a secure timestamp.",
  "docUrl": null,
  "architecture": "x86_64"
},
```

I'm about to push a PR that might fix the timestamp issue, which I'm hoping might fix the issue with `QSDroplet`. Alternatively, it might need some different handling because it is an app itself.

What does `QSDroplet` do? Does it provide the ability to drag and drop something onto the QS icon?

Pending tasks:
- ﻿[ ] Fix QSDroplet signing
- [ ] Investigate opportunities to harden security and ensure no private info can be leaked in logs or by a maliciously crafted PR
    - I think limiting the `sign` action to only run on tags and not PRs will go a long way here, another advantage of splitting it into two actions
- [ ] Determine how we want the DMG to be named
    - Taking `Quicksilver 1.6.1.dmg` as an example, I think continuing to get the `1.6.1` from `Info.plist` is fine
    - As an alternative, we could take `1.6.1` from the tag, which would give us some flexibility to e.g.: `git tag 1.6.1-hotfix && git push --tags` -> `Quicksilver 1.6.1-hotfix.dmg`

skurfer:
> What does `QSDroplet` do? Does it provide the ability to drag and drop something onto the QS icon?

It has a different icon and would be a separate drop target from the main app (or potentially many targets). It’s sort of like a trigger, but you drop something on it.

https://qsapp.com/manual/Preferences/#command-objects-and-droplets

n8henrie:
```console
$ codesign --verify -vvv Quicksilver.app/Contents/Resources/QSDroplet.app/
Quicksilver.app/Contents/Resources/QSDroplet.app/: valid on disk
Quicksilver.app/Contents/Resources/QSDroplet.app/: satisfies its Designated Requirement
```

Quicksilver/Tools/qssign
```diff
@@ -0,0 +1,72 @@
...
+  -format UDZO \
+  -imagekey zlib-level=9 \
+  -ov \
+  -o "${QS_VERSIONED_DMG}"
+
+codesign --deep --timestamp --options runtime --sign "${SIGNING_IDENTITY}" "${QS_VERSIONED_DMG}"
+codesign --verify --deep --strict --verbose=1 "${QS_VERSIONED_DMG}"
+
+## Easy access to plist and app
+cp -a "${DMG_TEMP}"/Quicksilver.app{,/Contents/Info.plist} "${BUILT_PRODUCTS_DIR}"
```
skurfer, line 55: Why are we copying the app back to $BUILT_PRODUCTS_DIR here? That’s where it originates from, and it still exists, so the script fails for me as a result. We should just need `Info.plist` from this step.

n8henrie: Doing a codesigning dance. The unsigned app is copied into `BUILT_PRODUCTS_DIR`; if `QS_BUILD_ONLY` is set, this copy is ad-hoc signed so that it can be run locally, and I left a copy in `BUILT_PRODUCTS_DIR` as a convenience to preserve existing behavior.

It is copied into `DMG_TEMP` so that this directory can be uploaded as a built artifact that can them be downloaded and manually explored if desired, but more importantly because artifacts within a single workflow can then be downloaded in other actions, avoiding the need to re-build everything.

In `qssign`, the app is codesigned, and the the "finished" app is subsequently re-copied to `BUILT_PRODUCTS_DIR`, overwriting the unsigned copy that was there previously.

I'm surprised that the script fails for you (thanks for testing) -- shouldn't `cp -a` just overwrite it?

Ahh --

```console
$ cd $(mktemp -d)
$ mkdir ./dmg
$ /bin/cp -a /tmp/QS/build/Release/Quicksilver.app .
$ /bin/cp -a ./Quicksilver.app ./dmg/
$ /bin/cp -a ./dmg/Quicksilver.app .
cp: cannot overwrite directory ./
... lots of errors ...
$ which cp
/opt/homebrew/opt/coreutils/libexec/gnubin/cp
$ cp -a ./dmg/Quicksilver.app .
$ echo $?
0
```

The issue didn't come up on GitHub Actions because the environments are cleared between the two actions (`BUILT_PRODUCTS_DIR` is unpopulated in `qssign`).

I reran with `PATH=$(getconf PATH)` to make sure no more GNU / BSD coreutils issues.

n8henrie:
A few other conveniences provided by releases:

The canonical URL to the asset is `https://github.com/quicksilver/Quicksilver/releases/download/<tag-name>/<asset-name>`, but you'll get evergreen redirects with:

- <https://github.com/quicksilver/Quicksilver/releases/latest>
- <https://github.com/quicksilver/Quicksilver/releases/latest/download/Quicksilver 1.6.1.dmg>

For better or worse the version number in the DMG prevents one from getting an "always working direct download url."

skurfer (approved):

PR #2660, opened 2022-02-25
Typo: ocurred -> occurred

n8henrie:
I can probably push something this simple straight to master, right?


pjrobertson:
Yes :) 

Although we should set out rules on what can be pushed straight to maser. I'd say a good one to start with is:

Anything text or language only can be pushed straight to master. This includes:
* translation updates
* typo fixes
* Error message / log improvemts
* Code comments

However, user-facing strings that want to be changed (e.g. prefs, the 'Send to Quicksilver' name etc. should still have PR opened so at least one other person can OK the language

PR #2665, opened 2022-03-05
Make donation links more prominent + add alert

pjrobertson:
1. Adds donation links to the main QS menu, and to the menu bar icon
2. Adds a donation alert, which opens on Quicksilver startup
 
The donation alert appears for any of the following:
* User upgraded Quicksilver
* User hasn't seen the alert for 1 week
    
The alert will be suppressed for the following:
* It's the first time the user's launched Quicksilver (we don't want to bother them if they're trying out Quicksilver for the first time)
* The user checked 'Do not show again for this version', and they haven't just upgraded
* For devs who've set a hidden pref (see here)

The only point worth making is that I guess a lot of users (like myself) only shut down their Macs every 2-3 weeks or months. That which means QS is only launched every 2-3 weeks or months, so the donation alert won't appear that often. I'm OK with that, and anyway - we give users the option to disable the alert anyway. The way to get the alert to appear more often is to make more releases of QS ;-)

### DIFF
--- a/Quicksilver/Code-App/QSController.h
+++ b/Quicksilver/Code-App/QSController.h
@@ -22,6 +22,7 @@
 + (id)sharedInstance;
 - (IBAction)runSetupAssistant:(id)sender;
 - (NSProgressIndicator *)progressIndicator;
+- (IBAction)openDonatePage:(id)sender;
 - (IBAction)showPreferences:(id)sender;
 - (IBAction)showGuide:(id)sender;
 - (IBAction)showPlugins:(id)sender;
--- a/Quicksilver/Code-App/QSController.h
+++ b/Quicksilver/Code-App/QSController.h
@@ -15,6 +15,7 @@
 	BOOL versionChanged, runningSetupAssistant;
 	NSObject *dropletProxy;
     NSString *crashReportPath;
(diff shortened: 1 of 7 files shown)


skurfer:
I’ve been running a build with this included for a while without issue. I haven’t had time to look over the code closely, but I trust you. 😄

It’s been a while… [inhales deeply] MERGE!

skurfer:
I just remembered that we released 1.0.0 on March 25, 2013 so if you want to wait a couple of days… I doubt that anyone cares though.

pjrobertson:
> Looks pretty good to me. The links work.
> 
> How to I force it to pop-up again? I didn't see the "Do not show again" button and wanted to check that out. Tried deleting caches, cleaning and rebuilding,
> 
> ```
> $ grep -r -i -a donation ~/Library/Application\ Support/Quicksilver ~/Library/Preferences/com.blacktree.Quicksilver.plist
> $
> ```

It'll only remind you once per week, maximum. If you want to reset and get reminded instantly:

```
defaults delete com.blacktree.Quicksilver donate.reminder.suppress.next 0
defaults delete com.blacktree.Quicksilver donate.reminder.last
```

PR #2669, opened 2022-03-12
Update QSFadingView

### LINKED ISSUE #1215: Primer doesn't like 'indirectOptional' arguments
This one courtesy of @tiennou 

For some strange reason Primer places the iSelector on top of the dSelector for indirectOptional commands (try Run […])

pjrobertson:
Updates QSFadingView to drop using a hacky `_recursiveDisplayAllDirtyWithLockFocus` and use `NSAnimationContext` instead.

Also fixes #1215


pjrobertson:
Merging this as it's a small fix and mostly only related to the primer interface

PR #2670, opened 2022-03-12
Fix enabling/disabling of proxy objects in catalog

### LINKED ISSUE #1660: Can't check or uncheck proxy objects
From:
https://groups.google.com/forum/?fromgroups#!topic/blacktree-quicksilver/F5UOCbFGY7k

When trying to uncheck items from the Catalog > Quicksilver > Proxy Objects, nothing happens. The boxes stay checked. See screenshot:

I think it does actually remove the item from the catalog, because after trying to uncheck it, it doesn't show up in the global search. But the opposite is not true, I can't get them back. 

I can still double click the items to show them in the main Quicksilver interface. So creating proxy based triggers still works.

There are no issues with the other catalog groups, like Internal Commands and Internal Objects.

Quicksilver 1.1.2 (4006)
OS X 10.8.5

pjrobertson:
Fixes #1660

Pretty simple fix. `QSProxyObject` (a subclass of QSBasicObject) override the `enabled` method, and did not call through to `super`.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSProxyObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSProxyObject.m
@@ -105,7 +105,7 @@
     NSNumber *e = [[self proxyDict] objectForKey:@"enabled"];
     if (e)
         return [e boolValue];
-    return YES;
+    return [super enabled];
 }
 
 - (BOOL)hasChildren {return YES;}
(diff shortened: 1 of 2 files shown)


pjrobertson:
Again, another small fix which I'll merge straight into master.

I'm keen to get 2.1.0 out

Issue #2668, opened 2022-03-10
Crash updating UI on background thread


pjrobertson:
**Describe the bug**
Xcode gives a `Unexpected outstanding background CATransaction` message and a crash when opening the UI while also checking for updates.

**To Reproduce**
1. Turn on 'Check for Updates' → 'On Launch'
2. Restart Quicksilver
3. Immediately open up the preferences, and then scroll on any scroll view
4. See Quicksilver Crash

**Screenshots**

https://user-images.githubusercontent.com/150431/157686741-264db3cc-ff5c-4331-b4b8-37d852ba9b54.mov

**What version of macOS?**
10.14.6

**What version of Quicksilver?**
Master + 1.6.1 
I thought it was because of my recent changes to add a spinning wheel in the UI to show when updating, but I went back to the current release and I still got the same problem.

skurfer:
I remember consistently seeing a crash when checking for updates because it was trying to show the alert window on a background thread. I thought we fixed it, but can’t find the PR. Maybe it was fixed in an OS update?

In any case, I can’t reproduce the crash using either of the methods described here.

pjrobertson:
Reopening as I found out the reason (plugin updates were running on a background thread and then causing issues when updating the interface).

Fixed by #2671

pjrobertson:
Fixed by #2671

pjrobertson:
Thanks for the debugging.

OK, so I think this is a macOS bug then. I debugged, and the NSAlert *is* getting called on the main thread, but what’s happening is this:

1. NSAlert gets called on main thread, blocking it
2. The OS decides to use NSConcurrentScreenUpdate to update the UI (I can’t find a way not to do this)
3. The OS tried to do an update on the background thread
4. Boom

I’ve only noticed this bug now because I turned on ‘Check for updates on launch’ to debug some other stuff. I think this is a pretty rare edge case, so I won’t spend any more time on it.

PR #2671, opened 2022-03-12
Update plugin updater to use blocks and better background thread management

pjrobertson:
* Fixes a crash where some NSView updates were done on the background thread
* Removes a lot of the NSNotification calling
* Reduces/removes some obsolete code

The code changes look pretty big and scary, but ultimately all I did was:

Switched to using the new async `NSURLSession` instead of the synchronous `NSURLConnection` to make HTTP requests. This in turn meant that I had to pass `completionHandler` blocks around to know what to do once the request had completed.

The plugin code is really messy/complicated. I tried to tidy it up and remove calls to delegates etc. - blocks are much cleaner


pjrobertson:
Fixes #2668

skurfer:
Should this be in 2.0.0 since it causes crashes pretty consistently?

pjrobertson:
Yes and no. I'd say it's fairly high priority, but I don't want it to delay 2.0.0. I'm almost thinking 2.1.0 would come out a few just a few weeks after 2.0.0. We'll put 2.0 out there, wait for some feedback for 1-2 weeks, make any changes then release 2.1.0

skurfer:
lgtm, and I used a test build that included these changes to grab my updated Clipboard plug-in.

skurfer (approved):

PR #2672, opened 2022-03-13
Disable library validation

pjrobertson:
Allows loading un-signed plugins

Test pull request to have GH Actions build a signed copy to debug.

Related to https://github.com/quicksilver/Quicksilver/issues/2634#issuecomment-1059642928

### DIFF
--- a/Quicksilver/Quicksilver.entitlements
+++ b/Quicksilver/Quicksilver.entitlements
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.cs.disable-library-validation</key>
+	<true/>
+</dict>
+</plist>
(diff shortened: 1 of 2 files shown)


pjrobertson:
OK, so this builds fine and verifies fine:

```
/Downloads ❯❯❯ spctl --assess --verbose --type open --context context:primary-signature "Quicksilver 1.6.1.dmg"
Quicksilver 1.6.1.dmg: accepted
source=Notarized Developer ID
/Quicksilver ❯❯❯ spctl --assess --verbose --type open --context context:primary-signature "Quicksilver.app"
Quicksilver.app: accepted
source=Notarized Developer ID
```

Since this is such a quick and easy change, and removes the need for us to re-download and re-built all other plugins, here's my proposal:

1. This gets merged into 2.0, we release 2.0 'as is' now that most plugins have already been rebuilt for M1
2. Once 2.0 is out, we can take our time to code-signing all other outstanding plugins and release them post 2.0 release
3. For 2.1, we re-enable Library Validation, that means only plugins signed by us can be loaded.

In my mind, this + the Donation PR should be all we need to get 2.0 out. There are a few outstanding plugins, but I'm not that concerned - they're mostly ones that don't work now anyway.

skurfer (approved):

PR #2675, opened 2022-03-14
Library validation

pjrobertson:
Turns out I was a bit premature in my last release. I set up the XCode proj to disable library validation, but of course we're signing using `codesign`. 

These changes should fix that.


pjrobertson:
OK, this works. Sorry @skurfer  - I should have double checked. I'll merge this directly.

Issue #2685, opened 2022-03-25
Fix Quicksilver icon for more sizes


pjrobertson:
Some icon sizes are missing in the newest QS icon, created by vadimrob.

-----

From @lgarron:

Would it make sense to add more resolutions to the `.icns` files?

In https://github.com/quicksilver/Quicksilver/pull/2544 I put a lot of care into the dimensions to match Apple's guidelines/icons, and also to make sure that the horizontal short line at the bottom remains visible in Finder at 1x monitor resolution (right side of screenshot).

![Screen Shot 2022-03-16 at 17 41 39](https://user-images.githubusercontent.com/248078/158714144-92421703-29a1-4692-a7a2-573264a40f92.png)

_Originally posted by @lgarron in https://github.com/quicksilver/Quicksilver/issues/2595#issuecomment-1069784894_

Original `.afdesign` files are available in that post as well for editing

@lgarron, would you be able to update the icons?

lgarron:
Okay, here's a go that tries to match @vadimrob's design and dimensions. I'll make a PR for it.

![icon_16x16](https://user-images.githubusercontent.com/248078/160120524-0eb7dbd7-afc6-4733-bf95-95bca18ee7eb.png)
![icon_32x32](https://user-images.githubusercontent.com/248078/160120525-53cd2068-d06d-4c9a-931a-af678602286d.png)
![icon_32x32@2x](https://user-images.githubusercontent.com/248078/160120526-ad456dc6-de3d-45e1-b550-44827417a016.png)
![icon_128x128](https://user-images.githubusercontent.com/248078/160120529-db51ed66-8775-4f80-b573-fe634def19c8.png)
![icon_256x256](https://user-images.githubusercontent.com/248078/160120530-83b8d6e6-d575-43ee-a226-ffcb1e9e528f.png)
![icon_256x256@2x](https://user-images.githubusercontent.com/248078/160120531-e55fe4da-dd4b-4a36-b8db-bcc4d1a75a8f.png)
![icon_512x512@2x](https://user-images.githubusercontent.com/248078/160120534-caa3c181-9662-4b79-b407-db1397267538.png)

If we want a bigger purple surface, I'd be more partial the preserving the traditional Quicksilver style of using a purple inset, also matching current icons for apps like Safari:
![icon_512x512@2x](https://user-images.githubusercontent.com/248078/160121853-58fccfe1-4e87-4582-8124-5e16d1b5fddc.png)
[...]

PR #2595, opened 2022-02-03
Update Quicksilver.icns 2022

vadimrob:
new quicksilver icon 2022
what do you think?

qs
Screenshot 2022-02-03 at 12 18 48
2

PR #2686, opened 2022-03-25
Update the white-on-purple icon to a version with more sizes.

lgarron:
For https://github.com/quicksilver/Quicksilver/issues/2685

This is an adaptation of the design from https://github.com/quicksilver/Quicksilver/pull/2595 with some of the tweaks from https://github.com/quicksilver/Quicksilver/pull/2544 to better match the macOS icon template sizes from Apple and make the icon look sharper at small resolutions. In particular, the `16x16(@1x)` icon contains a lot of manual adjustments:

Dark mode:
Screen Shot 2022-03-25 at 19 38 17

Light mode:
Screen Shot 2022-03-25 at 19 38 35

icon_16x16
icon_32x32
icon_64x64
icon_128x128
icon_256x256
icon_512x512
icon_512x512@2x

Quicksilver still has `.icns` files with an unusual spread of resolutions (old on the left); I've tried to keep it close (new on the right):
Screen Shot 2022-03-25 at 05 41 12
It would still be nice to update to a single icon in the future.

Happy to revise this if it's still not quite right.


lgarron:
Looks like the `build / sign` CI job doesn't like my face. 🤷

pjrobertson:
Thanks @lgarron - these look really good! No problems here :)

You don't need to worry about the build/sign failing. It fails for anybody who is not an owner of this repository. Perhaps the only thing I would ask is - can you add the original files (AI/PSD/afdesign) that you used to the repository, in the same folder. That way we always have them there as a record.

Thanks!

lgarron:
> Perhaps the only thing I would ask is - can you add the original files (AI/PSD/afdesign) that you used to the repository, in the same folder. That way we always have them there as a record.

Unfortunately I'm not really about the situation with this. I accidentally ended up using a trial of the subscription version of Sketch (it is sneakily presented as a routine update from the last perpetual license version). I'll have to figure out a more sustainable format approach. :-/

lgarron:
> > Perhaps the only thing I would ask is - can you add the original files (AI/PSD/afdesign) that you used to the repository, in the same folder. That way we always have them there as a record.
> 
> Unfortunately I'm not really about the situation with this. I accidentally ended up using a trial of the subscription version of Sketch (it is sneakily presented as a routine update from the last perpetual license version). Also, the file is 6.7MB, which I feel isn't really an appropriate weight. I'll have to figure out a more sustainable format approach. :-/

Here's a 4.9MB slightly pared down version, if you still want:
https://www.dropbox.com/s/7lemwwwu89vg9j2/Quicksilver%20Big%20Sur%20Inverted%20v0.8.4.1.sketch?dl=

pjrobertson:
Great! I'll merge this and then add the sketch file to the repository as well :)

On behalf of the Quicksilver community - thanks!

pjrobertson:
P.S. - note that I only meant for this file to be included in the QS repository for future needs, not in the app itself. So it doesn't really matter how big it is. It can just sit in the repository until it's needed again.

I've uploaded your 4.9MB version here: https://github.com/quicksilver/Quicksilver/commit/4e37a9caa67b042d77614c60712859da950faece

But please make a PR for the full 6.7MB if there are significant differences between the two versions.

lgarron:
> But please make a PR for the full 6.7MB if there are significant differences between the two versions.

There are no differences that affect the final icons.

Adding the final file takes the checked-out files from 12.5MB to 17.2MB (and adds a few percent to the `.git` dir), which I think isn't a *great* use of bytes. If I'd have known this would be checked in, I would have found a way to generate these from something much lighter. But I'll certainly defer to you on whether to keep this in tree!

PR #2697, opened 2022-03-31
Update Primer interface for dark mode

### LINKED ISSUE #2695: Primer interface doesn't look great in dark mode
The primer interface interface in 2.0.0 is pretty much unusable in dark mode. Not only is it screamingly bright, also the borders disappear and the actual action is not readable.

This is on macOS Catalina and I've included screenshots in dark and in bright mode for comparison:

Screenshot 2022-03-30 at 18 51 46
Screenshot 2022-03-30 at 18 56 16

pjrobertson:
Fixes #2695

Note that this is also explicitly disables using custom colours for the Primer interface. It was already pretty messy before: you could only change the window background colour, and that was done by setting the "Result Headers" background colour. Strange.

I figured since Primer is a... 'primer' interface, then hard-coding colours is preferable.


pjrobertson:
Sorry, not sure what happened there. Have just re-pushed. Once the build checks are successful I'll merge :)

Issue #2681, opened 2022-03-23
App hangs with beach ball of death and can't be easily restarted


lucian303:
**Describe the bug**
QS starts to hang and needs to be force quit constantly. After being killed, it still won't restart for quite some time. After I kill it 
with
```
pkill Quicksilver && open /Applications/Quicksilver.app
```

I get:

```
The application cannot be opened for an unexpected reason, error=Error Domain=NSOSStatusErrorDomain Code=-600 "procNotFound: no eligible process with specified descriptor" UserInfo={_LSLine=380, _LSFunction=_LSAnnotateAndSendAppleEventWithOptions}
```

It takes quite a few minutes, at least enough to think about and write half of this bug report before it can be restarted. I'm using no plugins and only have keyboard shortcuts triggering the opening/switching of apps. 

**To Reproduce**
Use system normally. Quicksilver beach balls of death at least once every day or two. 

**Expected behavior**
It shouldn't crash but if it does crash, it should be able to be killed and restarted immediately even if the crash cannot be solved. 

**Screenshots**

**What version of macOS?**
11.6.4 (20G417)

**What version of Quicksilver?**
1.6.1 (4026)

pjrobertson:
Hi @lucian303 - thanks for the bug report!

We will be releasing Quicksilver 2.0 in the next 24 hours, which may well fix this issue. Please wait to upgrade to that version and try it out.

If you still experience a problem with Quicksilver 2.0, then the next time you see this issue, please open Activity Monitor, highlight Quicksilver and choose 'sample process' or 'spindump':
![Screenshot 2022-03-25 at 08 53 54](https://user-images.githubusercontent.com/150431/160033756-e87be340-1087-42ea-b2df-0fc0025ffe6b.png)


I will close this issue until we hear back from you regarding how you do with Quicksilver 2.0

Thanks!

kzykx:
Quicksilver hangs with a . 
I have to kill the app using Activity Monitor (quit or force quit)
[Spindump.txt](https://github.com/quicksilver/Quicksilver/files/8355090/Spindump.txt)
[Sample of Quicksilver.txt](https://github.com/quicksilver/Quicksilver/files/8355091/Sample.of.Quicksilver.txt)

n8henrie:
@kzykx -- a couple thoughts

- We'll probably need you to reproduce this on the final release build
- Have you gone through the steps in [the FAQ](https://qsapp.com/manual/FAQ)?

[...]

PR #2698, opened 2022-04-01
Make QSTask main thread calls async

### LINKED ISSUE #2661: Hangs related to QSTask
I’ve been seeing QS “Not Responding” with the latest Universal builds from Xcode 13.

I’ve sampled the process and gotten some spindumps, but in every case, it appears to be related to the `@synchronized` calls in `QSTask`.

This happened the first 3 times I tried to install plug-in updates, but not the 4th. I’ve also seen it happen when rescanning the `QSProcessObjectSource`, but that thing gets rescanned every time you switch applications in macOS, so it’s probably a potential issue with all catalog entries and that one is just the most likely to hit it.

I’ve attached a sample. I plan to look into this when I can, but if anyone else is bored, feel free. 

[Sample of Quicksilver.txt](https://github.com/quicksilver/Quicksilver/files/8154608/Sample.of.Quicksilver.txt)

pjrobertson:
Fixes #2661

This is a simpler fix than I thought (I hope...)

It turns out we're already running scans on a background thread. See `QSCatalogEntry.m` at `- (NSArray *)scanAndCache;`

A further look at the spindumps in #2681 showed that the specific problem is with QSTask, and locking on showing/hiding tasks:

Main Thread:

```
                                                                    1000  __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 12 (CoreFoundation + 479292) [0x7ff81a95203c]
                                                                      1000  -[QSLibrarian reloadSource:] + 562 (QSCore + 92849) [0x106e8eab1]
                                                                        1000  -[QSTask stop] + 24 (QSCore + 314328) [0x106ec4bd8]
                                                                          1000  objc_sync_enter + 27 (libobjc.A.dylib + 74862) [0x7ff81a73146e]
                                                                            1000  __ulock_wait + 10 (libsystem_kernel.dylib + 12522)
```

2nd thread:
```
              1000  __42-[QSLibrarian scanCatalogIgnoringIndexes:]_block_invoke + 649 (QSCore + 105232) [0x106e91b10]
                1000  -[QSTask stop] + 85 (QSCore + 314389) [0x106ec4c15]
                  1000  -[QSTaskController taskStopped:] + 149 (QSCore + 316325) [0x106ec53a5]
                    1000  _dispatch_sync_f_slow + 170 (libdispatch.dylib + 67222) [0x7ff81a6e8696]
                      1000  __DISPATCH_WAIT_FOR_QUEUE__ + 296 (libdispatch.dylib + 68172) [0x7ff81a6e8a4c]
                        1000  _dispatch_thread_event_wait_slow + 40 (libdispatch.dylib + 14741) [0x7ff81a6db995]
                          1000  __ulock_wait + 10 (libsystem_kernel.dylib + 12522) [0x7ff81a85b0ea]
```

QSTask was posting notifications to the main thread **synchronously** which is what was causing the locking. This PR changes them to be called async, so the scan can finish.

Still, I think the point of making `reloadSource:` run on a background thread (and not the main thread) serially makes a lot of sense as well. If two things invalidate a catalog item within 1ms of each other, then it's better to have then scan one after the other, to pick up any more changes.

I'll implement this 2nd point if @skurfer  - you think it still makes sense after seeing the above fix.

### DIFF
--- a/Quicksilver/Code-App/QSTaskViewer.m
+++ b/Quicksilver/Code-App/QSTaskViewer.m
@@ -150,27 +150,30 @@ static QSTaskViewer * _sharedInstance;
     QSTask *task = notif.object;
 //    NSLog(@"Adding task: %@", task);
 
+    QSGCDMainAsync(^{
     if (!self.taskControllers[task.identifier]) {
         self.taskControllers[task.identifier] = [QSTaskViewController controllerWithTask:task];
     }
 
-    QSGCDMainAsync(^{
         [self updateTaskView];
     });
 }
 
 - (void)removeTask:(NSNotification *)notif {
+	
     QSTask *task = notif.object;
 //    NSLog(@"Removing task: %@", task);
 
-    QSTaskViewController *removedController = self.taskControllers[task.identifier];
 
-    QSGCDMainDelayed(STOP_DELAY, ^{
-        [self.taskControllers removeObjectForKey:task.identifier];
-        [removedController.view removeFromSuperview];
-
-        [self updateTaskView];
-    });
+	// delay removing the task for 0.5s, so that the user can see
+	usleep(STOP_DELAY*1000000);
+	
+	QSGCDMainAsync(^{
+		QSTaskViewController *removedController = self.taskControllers[task.identifier];
+		[self.taskControllers removeObjectForKey:task.identifier];
+		[removedController.view removeFromSuperview];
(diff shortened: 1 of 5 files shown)


pjrobertson:
OK, a few more commits here to use a background queue for all QSLibrarian scans, and also tidy up a couple of other issues.

Issue #2545, opened 2020-12-01
On Big Sur, my triggers (keyboard shortcuts) stopped working


captainmacnl:
I have tried adding Quicksilver to the Security&Privacy SystemPreferences-pane, adding it under Input Monitoring (it was already added to Full Disk Access, Files and Folders, Accessibility [although not to Automation?] ).
I have tried restarting Quicksilver, rebooting, but absolutely no reaction.

An example of a trigger that i use often is to trigger a search from text that i have selected. The trigger uses the selected text and does a search on a database that i have to check like 20 times a day, so this would help a lot in my workflow.

I have been using QS from the start and am refusing to go with any other utility because i love it.
But this really needs to be fixed!
Thanks.

skurfer:
I’ve never used (or even noticed) Input Monitoring and triggers are working fine for me.

I suppose the first thing I would check is that you have the current version (1.6.1). If you’ve been using it from the start, I know there were at least two versions where the auto-updates broke, so a lot of people are running old copies.

juanfal:
They have been working for me since first beta of Big Sur, and now on an Air M1.  What is failing, is Clipboard History

skurfer:
Not to hijack this issue, but what about Clipboard history? I have the panel docked to a screen edge and it sometimes gets confused, but other than that, it functions.

captainmacnl:
OK, I have no idea why I hadn't tried reinstalling the app yet.. *facepalm*
Everything works now! YAY I'm so glad i can still be using this amazing tool!!

captainmacnl:
OK. So i realised that all my triggers work, EXCEPT the one i use most: the 'find with' trigger, which i use to search a website.
I have even fully uninstalled quicksilver and all its preferences, reinstalled everything, but same result.

I've got the trigger setup in the following way:

Select an item:
 Current Selection
Action:
 Find With...
Target:
 The website address with a search query where i have it end in *** to be replaced with the current selected text.

Somehow, this trigger won't do anything.
All the other triggers do work.
It's independent of what key combo i assign it to.
If i set the trigger to 'Show Window' when triggered, it does show that quicksilver tries to do something, but nothing actually happens..
I have also tried setting up a new trigger with as basic a setting as any (ie: using the standard https://www.google.com/search?q=*** )
but still no dice

Anyone can help out?

skurfer:
[...]

PR #2703, opened 2022-04-03
Request access to accessibility and input monitoring on 10.15+

### LINKED ISSUE #2684: "Input Monitoring" required for Triggers to work.
@debugDS says:

I had to add Quicksilver also to **Input Monitoring** for Triggers to work.

_Originally posted by @debugDS in https://github.com/quicksilver/Quicksilver/issues/2545#issuecomment-1077700041_

pjrobertson:
Fixes #2684

### DIFF
--- a/Quicksilver/Code-App/QSController.m
+++ b/Quicksilver/Code-App/QSController.m
@@ -12,7 +12,7 @@
 #import "QSCrashReporterWindowController.h"
 #import "QSDownloads.h"
 #import "QSDonationController.h"
-
+#import <IOKit/hidsystem/IOHIDLib.h>
 
 #import "QSIntValueTransformer.h"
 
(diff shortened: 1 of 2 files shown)


pjrobertson:
@skurfer @n8henrie  - I would appreciate one of you testing this code. I've written it blind, since I'm on 10.14 and can't run the 10.15 code.

Basically, it uses a new 10.15 call to check if QS has access to listen to events (needed for mouse triggers) and post events (to allow QS to 'paste' and imitate pressing keys). It would be worth checking a couple of scenarios:

1. Standard scenario, just launch QS
2. Scenario where you've gone into System Preferences > Security > Privacy and specifically denied QS access in the 'Accessibility' and 'Input Monitoring' panes (by unticking 'Quicksilver' in the list).

For scenario 2, QS should show an alert and then open up the System Preferences for you.

skurfer:
It appears to work, but

1. It shows 2 messages that look like they’re from the app and 1 that looks like it’s from the OS. I think maybe the claret should be moved outside the `for` loop.
2. It adds QS to Input Monitoring (unchecked) automatically, but not to Accessibility.
3. I don’t think I ever had Input Monitoring allowed before, but **Paste** and **Type Text** always worked. From some quick testing, it looks like **Type Text**, **Paste**, and Keyboard Triggers all fail if I only enable Input Monitoring. If I only enable Accessibility and disable Input Monitoring, they all work. So I think this check/alert only needs to work about Accessibility.

pjrobertson:
Thanks for the debugging.

I think the original issue was filed RE mouse triggers. Can you confirm that mouse triggers work without enabling 'input monitoring'?

See: https://github.com/quicksilver/Quicksilver/issues/2545#issuecomment-1077700041

skurfer:
I’m messing with Mouse Triggers, but I can’t get them to run at all. I see the glow on the edge of the screen, so it looks like QS is able to see the state of the mouse. Need to do more debugging.

skurfer:
Scratch that. The Mouse Trigger I was testing with had some conditions I had forgotten about. Once I figured that out, I can confirm that it works (even if QS isn’t listed under Input Monitoring).

pjrobertson:
OK, so if you're able to use mouse triggers no problem, then I say there's no need for this, and no need to add an extra bit of code.

I'll mark #2684 as cannot reproduc

Issue #2727, opened 2022-04-10
Switch to using Github Release Page for archives


pjrobertson:
Replace the existing qsapp.com/archives and github.qsapp.com/ pages with GH Releases. Here are the steps we need to do:

- [x] Add checksum to GH Releases (@n8henrie )
- [x] Upload all previous versions from https://github.qsapp.com/ to GH Releases
- [x] Make sure each release that sets a new "min OS version" states it in the Release notes
- [x] Redirect qsapp.com/archives to github.com/quicksilver/quicksilver/releases
- [x] Delete the `gh-pages` branch in the main repo and turn of GH Pages integration.
- [x] Update Release Process in wiki to remove steps for adding to the 'archives'.

n8henrie:
>  Find a way to sign GH Releases

Can you clarify what you mean? 2.0.3 is both codesigned and the tag / release is GPG signed as well (green checkbox):

<img width="138" alt="Screen Shot 2022-04-11 at 13 29 48" src="https://user-images.githubusercontent.com/1234956/162815281-35f32f58-4146-41aa-a421-32715bbc0e72.png">

Can we check that one off?

> Upload all previous versions from https://github.qsapp.com/ to GH Releases

It looks like you can cut a release from an existing tag: <https://github.com/quicksilver/Quicksilver/releases/new>, so it *looks* like we should be able to select e.g. v1.0.0 and create a release et voila. However, I assume this would mess with efforts like https://github.com/Homebrew/homebrew-cask/pull/121662; I'm pretty sure Homebrew will be relying on the chronological order of a release irrespective of a tag name, and this would bump all Homebrew cask users (likely a nontrivial chunk of the userbase) back to an earlier version. @timvisher is that your understanding? @pjrobertson @skurfer do you think migrating old releases is worth possibly disrupting Homebrew users?

@timvisher you asked about pre-releases -- is there an escape hatch where we could add a suffix like `v1.0.0-archive` and avoid having Homebrew try to use that version?

The last 3 tasks should be relatively easy!

pjrobertson:
> Can you clarify what you mean? 2.0.3 is both codesigned and the tag / release is GPG signed as well (green checkbox):

Sorry, I think I meant a checksums. I saw somewhere you said you wanted to find a way to add MD5 checksums for homebrew or something.

timvisher:
[...]

PR #2737, opened 2022-04-15
Automatically Create Homebrew-Cask Bump PR on Release

timvisher:
This potentially uses GitHub Actions to automatically create a bump PR on
the Cask whenever we push a new semantic version tag.

It's currently untested and is meant mainly to ask for feedback on whether
it's desired or not.

Thoughts?

Related:

- https://github.com/Homebrew/homebrew-cask/pull/121662/
- https://github.com/quicksilver/Quicksilver/issues/2727

### DIFF
--- a/.github/workflows/bump_homebrew_cask.yml
+++ b/.github/workflows/bump_homebrew_cask.yml
@@ -1,0 +1,21 @@
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+- name: Update Homebrew formula
+  uses: dawidd6/action-homebrew-bump-formula@v3
+  with:
+    # Required, custom GitHub access token with the 'public_repo' and 'workflow' scopes
+    token: ${{secrets.TOKEN}}
+    # Optional, will create tap repo fork in organization
+    org: ORG
+    # Optional, defaults to homebrew/core
+    tap: USER/REPO
+    # Formula name, required
+    formula: FORMULA
+    # Optional, will be determined automatically
+    tag: ${{github.ref}}
+    # Optional, will be determined automatically
+    revision: ${{github.sha}}
+    # Optional, if don't want to check for already open PRs
+    force: false # true


n8henrie:
Thanks for bringing up the conversation! I've been using homebrew for a *long* time. I'm a little on the fence about this.

QS has its own update mechanism; aside from the recent <2.0.3 update glitch, it's worked well for a long time. Even if the brew-available version lags, QS should prompt them to update with minimal hassle. 

https://github.com/dawidd6/action-homebrew-bump-formula seems to have a pretty low bus factor, and looks like [the only commit in the last year was a README update](https://github.com/dawidd6/action-homebrew-bump-formula/commits/master); perhaps that's because it's stable and doesn't need anything, but it would be silly if it became unmaintained in 6 months and we just removed this process (and risk breaking workflows people had built around it in the meantime).

I don't really want to commit to maintaining a tap under the QS org; it looks like one can remove all of the `Optional` configuration and have it just submit a PR to `homebrew/core`? Is that your understanding?

Most of the work seems to be automated, which is great, not a huge loss if it doesn't work out, add some convenience for homebrew users if it does. If we merged this and the Actions dependency subsequently breaks or is deprecated, I'd probably recommend ripping it out over fixing it.

Strong feelings either way @skurfer @pjrobertson ?

pjrobertson:
1. Homebrew shouldn't replace our built-in updater, that's for sure. We need the ability to send out developer, pre-release and release builds, and a way for non-tech-savvy users to update.
2. I'm fine with us 'bumping' the homebrew project when we make a new release - like a notification. But as @n8henrie said - we unfortunately don't have the capacity to maintain it ourselves.
3. If all this PR is doing is making that 'bump', and not expecting anything else, then I'm fine with this being merged.
4. My only other concern is that we give the homebrew guys too much noise as we create/delete tags.

n8henrie:
SGTM

> I don't really want to commit to maintaining a tap under the QS org; it looks like one can remove all of the Optional configuration and have it just submit a PR to homebrew/core? Is that your understanding?

@timvisher if confirmed, would you mind removing all the optional config, and I'll figure out adding an access token.

> custom GitHub access token with the 'public_repo' and 'workflow' scopes

I don't see an option for access tokens at the organization level; should this be an access token generated from my personal account? (<https://github.com/settings/tokens>)

n8henrie:
bump @timvisher

timvisher:
@n8henrie Thanks for the bump. This is in my todo list and I do _really_ want to get to it but I've got a lot going on at the moment. If you'd like we can just close it and I'll reopen if I find a few spare cycles or we can leave it open. Sorry it's taking me so long. :\

PR #2751, opened 2022-04-26
QSDownloads: support Safari download folders containing spaces.

lgarron:
`[NSURL URLWithString:downloads]` fails when `downloads` contains spaces. Apparently the solution is to percent-encode the string.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSDownloads.m
+++ b/Quicksilver/Code-QuickStepCore/QSDownloads.m
@@ -45,7 +45,7 @@
     
     if (downloads) {
 		downloads = [downloads stringByResolvingSymlinksInPath];
-		return [NSURL URLWithString:downloads];
+		return [NSURL URLWithString:[downloads stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLHostAllowedCharacterSet]]];
     } else {
 		// fall back to the default downloads folder if the user settings couldn't be resolved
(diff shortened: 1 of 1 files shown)


pjrobertson:
Nice, this is a good fix to have. I'm looking at the code now - I can't get:

```
NSDictionary *safariPrefs = [defaults persistentDomainForName:@"com.apple.Safari"];
```

to work for me. It just returns `nil` every time. I think it has something to do with sandboxing (I confirm that the key exists at `defaults read com.apple.Safari DownloadsPath`).

Also, looking at Firefox, it seems it has its own download setting in `~/Library/Application Support/Firefox/Profiles/XXXXX.default/prefs.js -> browser.download.dir`. Seems like each browser may well have their own path.

To avoid us trying to figure out every browser's downloads dir, I suggest we have a setting in Quicksilver to set the downloads folder. A couple of options:

* Set it using `defaults write com.blacktree.Quicksivler downloadsPath XXXX` (easiest)
* Add a preference to Quicksilver Prefs > Extras (fairly easy)
* Add it as a 'source option' for the 'Latest Download' proxy object catalog source (hardest).

lgarron:
> I think it has something to do with sandboxing (I confirm that the key exists at `defaults read com.apple.Safari DownloadsPath`).

Ah, alas. :-/

This was more of a drive-by fix, so unfortunately I don't think I can contribute more work to it. But feel free to take it and run with it!

pjrobertson:
I think the right way to fix this is to use `-[NSURL fileURLWithString:]`. I've pushed straight to master. Thanks for spotting the bug!

Issue #2707, opened 2022-04-04
Change main development branch to `main`?


lgarron:
A lot of projects have renamed their main branch to `main`[^1], and GitHub makes it easy to do this in the repository settings to that the branch dropdown on the repo features a little popup about the change: https://github.com/github/renaming

<img width="501" alt="Screen Shot 2022-04-04 at 16 02 54" src="https://user-images.githubusercontent.com/248078/161646397-789fc015-ac25-4d0f-b960-9801618a963e.png">

Would you have any interest in switching over?

[^1]: In fact, so many projects have switched that `master` has left my muscle memory at this point.

skurfer:
I’m not opposed, but it’s not high priority right now.

lgarron:
> In addition to avoiding unintentional harm, the other benefit of migrating _now_ would be to avoid revisiting this topic going forward, as I imagine it will come up more than once.

This tends to be a good reason to do it. Basically everyone agrees that the name `main` is unproblematic. It's the new default, and many big projects have already tackled it earlier rather than later, so that they can just use the unproblematic name rather than debating about the old one.

> If we change the branch, we should probably also rename, reword, or redact the handful of other uses of `master` in the codebase -- looks like there's not too many: https://github.com/quicksilver/Quicksilver/search?q=master. @lgarron it seems like you've seen many more projects migrating than I have -- I assume that other projects changing their default branch are doing this as well?

To some extent. At this point, I'd say by far the most useful is renaming the development branch to match the modern default. Sometimes there are followup initiatives around removing other potentially exclusive language (`master`/`slave` databases, `whitelist`/`blacklist` instead of `allowlist`/`denylist`), but the Quicksilver codebase doesn't really seem to be using those. It looks like https://github.com/quicksilver/Quicksilver/blob/4374aea4eb98f8d5e22a447b2f2bc48dfa3736ed/Quicksilver/Tools/travis/check_indent.rb#L6 is the only change that would be needed at a technical level, although I'd be happy to send a PR to update the other references as well.

Some projects keep around a `master` branch for a short while (e.g. automatically kept in sync with `main` using GitHub Actions), but I don't think much would break if the `master` branch disappeared from this repo entirely. Pull requests would be automatically changed over, new checkouts would use `main`, and the popup would direct folks to rename their branch locally. Even if based off a local `master` branch from an older checkout, new pull requests will also still work.

pjrobertson:
[...]

PR #2759, opened 2022-05-01
Change the main branch reference in the main `README` to `main`.

lgarron:
There are a few other places that could be changed from `master` to
`main`, but they look mostly harmless to me.

The change to `check_indent.rb` may not actually be useful for anything:
https://github.com/quicksilver/Quicksilver/issues/2707#issuecomment-1094361203
But I've gone ahead and included it just in case.

Addresses https://github.com/quicksilver/Quicksilver/issues/2707

### DIFF
--- a/Quicksilver/Tools/travis/check_indent.rb
+++ b/Quicksilver/Tools/travis/check_indent.rb
@@ -3,7 +3,7 @@
 
 require 'set'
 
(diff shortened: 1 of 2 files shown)


pjrobertson:
I think it's probably worth changing the reference in `./SharedSupport/CHANGELOG`, just for consistency. I'll do that.

Thanks for all your hard work!

Issue #2868, opened 2022-05-09
Make it possible to run `qssign` outside of CI


pjrobertson:
It would be great if we were able to sign our own builds locally using `qssign`. A use case is that I want to build a debug version of the app, or I want to run a signed version of the debug app for testing.

Right now, it seems possible, with a few mods to the qssign script. Here's what I had to do:

```
xcodebuild -showBuildSettings -configuration Debug -scheme "Quicksilver Distribution"  >> /tmp/qs_build_settings 
cp Quicksilver.entitlements /tmp/
mkdir dmg
cp Resources/Images/QuicksilverDMG.icns /tmp/QS/build/Debug/dmg/.VolumeIcon.icns
cd /tmp/QS/build/Debug
mv ./* dmg
```

I *think* that was it. It would be great if the `qssign` script could detect where it's being run, and do this itself though!

@n8henrie - I'm hoping this is an issue you can handle :D

n8henrie:
Makes sense.

We'll continue to get some error message upon opening because the DMG won't be signed -- we don't have local credentials for that (which is a different kind of signature than that used for the app itself).

Currently a Debug build can be transferred to another computer and run without any signing at all, but at least on MacOS 12.3 the detection that it is a downloaded file leads to a "this app is broken" error with no obvious **GUI** workaround. CLI-savvy users can `xattr -c Quicksilver.app` and it will run, but obviously that's asking a bit much if you're providing a debug version for another user to test.

I think the easiest / best approach is to add a flag for configuration to `qsrelease`, which automatically runs `qssign`. I've tested #2869 and it seems to fit the bill. Current behavior is unchanged and produces a signed release build.

When I run as:

```console
$ SIGNING_IDENTITY='ABCD' Tools/qsrelease Debug
```

It correctly produces a Debug build, including a DMG that I've transferred to my wife's computer and run. There is the expected dialog requiring one to cmd-click and open (which has been documented before, due to unsigned DMG) but no "broken application" error as one gets with a completely unsigned build. It runs and shows it is indeed a debug build as expected.

pjrobertson:
Much simpler, I didn't realise I could just run `qsrelease` locally - no need to run qssign. The only other reason I had for running `qssign` was if I wanted to sign plugins locally. Do you think that's worth it?

I'm wondering now if we should build a release *and* debug build on every commit, or just a release build?

n8henrie:
[...]

PR #2869, opened 2022-05-09
Allow specifying the configuration for qsrelease

n8henrie:
Fixes https://github.com/quicksilver/Quicksilver/issues/2868

### DIFF
--- a/Quicksilver/Tools/qsrelease
+++ b/Quicksilver/Tools/qsrelease
@@ -1,22 +1,19 @@
 #!/bin/sh
 
 set -euf
+set -x
 
 TOOLSDIR=$(cd "$(dirname "$0")" && pwd)
 
-# Default to x86_64 for CI builds
-if [ -n "${CI:-}" ]; then
-  CURRENT_ARCH=x86_64
-else
-  CURRENT_ARCH=$(arch)
-fi
+CURRENT_ARCH=$(uname -m)
 
-## Quicksilver Release build
+CONFIGURATION=${1-Release}
+
 ## Cache settings to save time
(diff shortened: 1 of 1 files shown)


pjrobertson:
LGTM. One issue I have is that `CURRENT_ARCH=$(arch)` gives me `i386`. Apparently this is a quirk with `arch`(see [here](https://apple.stackexchange.com/questions/140651/why-does-arch-output-i386)) - I think what we actually want is `uname -m`. Can you confirm and make the change if necessary?

P.S. Right now, to get around this I define `CI` when I run `qsrelease` like this:

`CI="blahblahblah" ./qsrelease`

n8henrie:
> One issue I have is that CURRENT_ARCH=$(arch) gives me i386

Interesting! https://superuser.com/a/835519/321164

> Right now, to get around this I define CI when I run qsrelease like this:

😆 Sorry about that.

PR soon.

n8henrie:
Converted to a draft -- added some debug stuff to see what the values are in CI (which I will remove before ready to merge).

n8henrie (commented):
I guess I can't turn it back from a draft into a PR without reviewing it.

n8henrie:
@pjrobertson @skurfer would you guys mind pulling and building locally to make sure the `uname -m` change works okay locally? Seems like it's doing the right thing in CI.

n8henrie:
Closing in favor or https://github.com/quicksilver/Quicksilver/pull/2871, which contains this and a bit more.

pjrobertson:
OK. I was just about to review this :)

Issue #2604, opened 2022-02-12
Custom AppleScript actions: support JavaScript for automations (JXA)


n8henrie:
Copied from: https://groups.google.com/g/blacktree-quicksilver/c/I6dkTGN0bI8/m/ebbi1s46AAAJ

Low priority. I will take another look at this after 1.7.0

---

After a couple days of StackOverflow threads, I can't figure out whether I can use JavaScript for Automation (JXA) in Quicksilver actions.

I have a ton of AppleScript actions, and they're one of the best parts of Quicksilver IMO, but I really hate working with AppleScript. I'm not huge on JavaScript either, but having `map` / `filter` and a more traditional syntax really is a blessing compared to AppleScript. (For context, most of my current AppleScript actions just shell out to python or what have you, which is workable.)

If you open up Script Editor, set the language to JavaScript, then open the Quicksilver dictionary, it *looks* like it supports the same handlers, just with slightly different names (instead of `on process text` it's `processText`).

Unfortunately I've had no luck! I've tried a million or so different configurations of the below.

```javascript
const Quicksilver = Application("Quicksilver")
Quicksilver.includeStandardAdditions = true

function processText(text) {
    return "bar"
}

function openFiles(files) {
    return
}

function getIndirectTypes() {
    return [$.NSStringPboardType]
}

function getDirectTypes() {
    return [$.NSStringPboardType]
}

function getArgumentCount() {
    return 3
}
```

Looking through the source, I don't know enough Obj C to say what the issue may be. I do see `"AppleScript Action: No handler? Aborting..."`` in the logs, and poking around that place in the source and adding a few debug logs, I can see that it isn't finding any handlers in my test file.

Any ideas on this? I'd sure rather be writing these in JS (or python, or go, or swift, or rust, or ...) rather than AppleScript!

![image](https://user-images.githubusercontent.com/1234956/153687356-499b0d90-ef72-498e-98e2-d8ae6deba440.png)

n8henrie:
> Are these better documented somewhere by Apple?: @"aevtoapp", @"DAEDopnt", @"aevtodoc", @"DAEDopfl"]

No, these are defined in <https://github.com/quicksilver/Quicksilver/blob/master/Quicksilver/Scripting/Quicksilver.sdef>; each app that wants to implement "scriptability" can apparently define their own magic strings in this way.

[...]

PR #2883, opened 2022-05-30
Fix direct and indirect type handlers for JXA

n8henrie:
Fixes https://github.com/quicksilver/Quicksilver/issues/2604

Would love any feedback from @pjrobertson or @skurfer (on anything but variable names, but I guess that's fine too 😆).

### DIFF
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSAppleScriptActions.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSAppleScriptActions.m
@@ -321,9 +321,11 @@
     if ([hType isEqualToString:@"direct"]) {
         appleScriptTypeArray = [self typeArrayForScript:path forHandler:@"DAEDgdob"];
         jxaTypeArray = [self typeArrayForScript:path forHandler:@"function getDirectTypes()"];
-    } else {
+    } else if ([hType isEqualToString:@"indirect"]) {
         appleScriptTypeArray = [self typeArrayForScript:path forHandler:@"DAEDgiob"];
         jxaTypeArray = [self typeArrayForScript:path forHandler:@"function getIndirectTypes()"];
+    } else {
+        NSLog(@"Expected 'indirect' or 'direct' for hType, found %@\n", hType);
     }
 
     NSMutableArray *dTypesArray = [NSMutableArray array];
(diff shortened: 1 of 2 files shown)


pjrobertson:
Other than a few minor styling suggestions, I don't see any glaring problems with this. Again, the hard-coding of function names (with half a bracket now `(`) isn't ideal, but if it works, then why not? :D

n8henrie:
> Other than a few minor styling suggestions

Would you mine elaborating? I think I have a rudimentary sense of style and preferences in bash and python, and rust and go do some hand-holding with their linters, but this is totally new territory and feels very foreign. For example I was hoping to find a more satisfying way to concatenate the arrays.

> Again, the hard-coding of function names (with half a bracket now () isn't ideal

Agreed! If you come across a better solution let me know. As far as I can tell it is identical to what we've been doing with AppleScript, just that AppleScript obfuscates the function names (which is why it [also works if you just add the AppleScript handler name in a JavaScript comment](https://github.com/quicksilver/Quicksilver/issues/2604#issuecomment-1104535189), just feels gross). I figure trying to get as specific as a match as possible is the best approach, which is why functions that accept zero arguments include both parentheses, and the rest accept up to the opening parenthesis.

pjrobertson:
> Would you mine elaborating?

Sorry, I wrote them directly on the code again, but I think last time you had trouble seeing them?

Screen Region 2022-05-31 at 08 36 04

n8henrie:
> Sorry, I wrote them directly on the code again, but I think last time you had trouble seeing them?

Yes, they're still not showing up for me. Did you hit the `Submit Review` button?

pjrobertson (changes requested):
Hi Nate. Hopefully you can see these now?

n8henrie:
@pjrobertson Yes, thanks for getting the review fixed!

I incorporated your feedback, but also updated a method that has been deprecated for a *long* time to use what I think is the current approach. I'm not sure about OSANull for the storage options, but it seems to be working for me.

Could you and @skurfer run this locally to make sure your applescripts are also still working as expected?

pjrobertson:
Code looks good to me. 

By 'deprecated method' you mean:
```
[[OSAScript alloc] initWithScriptDataDescriptor:aed
            fromURL:[NSURL fileURLWithPath:scriptPath]
            languageInstance:[language sharedLanguageInstance]
            usingStorageOptions:OSANull error:&err];
```

Not deprecated for me on 10.14 ;-) `AVAILABLE_MAC_OS_X_VERSION_10_6_AND_LATER` :D

I'll run with this now

pjrobertson:
Tested, LGTM. I'll leave open in case @skurfer has anything to add :)

pjrobertson:
I'm going to merge this now, so we can get it out there :)

pjrobertson (approved):

PR #2887, opened 2022-06-12
Give file objects a string value

### LINKED ISSUE #1673: Unable to combine objects of different types (Previously: Append Text actions don’t handle combined objects correctly)
When you mix manually entered text objects with certain other objects using the comma trick, only the first object selected is passed to the action.

To reproduce:
1. Create `~/Desktop/Input.txt`
2. Invoke QS in text-entry mode
3. Enter a string, like “test”
4. Hit `⎋` then `,`
5. Select a file
6. Run Append To… Input.txt

The string “test” will be appended, but not the path to the file. If you select the file first, then hit `,` and `.` to enter a string, only the file’s path will be appended.

This definitely happens with file objects and tags (in the third pane). Combining text-entry objects with URLs appears to work as expected.

pjrobertson:
Useful for when a string type is required e.g. when combining multiple objects (string + file)

Fixes #1673

-----

I'm opening a PR for this since I note that originally @skurfer had earmarked it to fix. The fix I implemented is essentially what I said in [this comment](https://github.com/quicksilver/Quicksilver/issues/1673#issuecomment-29866876). 

I'm not sure of the consequences of giving file objects a 'string type' from the start, so I think it's worth @skurfer running with this build for a while.

P.S If you can download a debug version from here without having to build yourself ;-)
https://github.com/quicksilver/Quicksilver/actions/runs/2483657221

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
@@ -698,6 +698,7 @@ NSArray *QSGetRecentDocumentsForBundle(NSString *bundleIdentifier) {
 		if ([paths count] == 1) {
 			NSString *path = [paths lastObject];
 			[[self dataDictionary] setObject:path forKey:QSFilePathType];
+			[[self dataDictionary] setObject:path forKey:NSPasteboardTypeString];
 			NSString *uti = [self fileUTI];
 			id handler = [QSReg instanceForKey:uti inTable:@"QSFileObjectCreationHandlers"];
 			if (handler) {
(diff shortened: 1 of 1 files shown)


skurfer:
The impact of this probably varies from user to user. Basically, anyone that has an action that applies to text higher up on the list than Open (or whatever they want as the default for files) could get an unpleasant surprise. Easily fixed, but not sure if people would know that. It’s probably worth it to make this change, though.

skurfer (approved):

pjrobertson:
So I've been thinking about this more recently, and this would fix the issue for File Objects, but not other objects types that don't have a string type still wouldn't work (in the issue above you mentioned tags).

Here's what I propose:

Change `-[QSObject stringValue]` to add:

```
id handler = [self handlerForType:[self primaryType] selector:@selector(stringValueForObject:)];
if (handler) {
stringValue = [handler stringValueForObject:self];
}
```

That way, any object handler (including plugins) can decide what to return for the 'string value' for a given object type on a case by case basis. This is arguably clearer than having plugin devs remember to set `[obj setObject:myString forType:QSTextType]` on every object.

I'm working on implementing this to see how it runs

n8henrie:
It doesn't seem like this is going anywhere, some I'm going to close for now -- can reopen at any time.

PR #2916, opened 2022-08-08
Fix cycle in dependencies to get tests running

n8henrie:
- Revert "Disable parellalize builds for preflight"
- Fix comment prefix
- Remove multiple deps from Quicksilver Tests to fix cycle in dependencies

See also: https://github.com/quicksilver/Quicksilver/issues/2583

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSLibrarian.m
+++ b/Quicksilver/Code-QuickStepCore/QSLibrarian.m
@@ -313,23 +313,26 @@ static CGFloat searchSpeed = 0.0;
 
 
 - (void)reloadSets:(NSNotification *)notif {
-    [self.objectDictionary removeAllObjects];
-	//NSLog(@"cat %@ %@", catalog, [catalog leafEntries]);
+	QSThreadSafeMutableDictionary *newDict = [[QSThreadSafeMutableDictionary alloc] initWithCapacity:self.objectDictionary.count];
+	//NSLog(@"cat %@ %@", catalog, [catalog leafEntries]);dls
     for(QSCatalogEntry * entry in [catalog leafEntries]) {
         for (QSObject *obj in [entry contents]) {
-            if ([obj identifier]) {
-                [self setIdentifier:[obj identifier] forObject:obj];
+			NSString *identifier = [obj identifier];
+            if (identifier) {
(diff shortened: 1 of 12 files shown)


n8henrie:
Removed several dependencies from `Quicksilver Tests` and they seem to at least run now:

Screen Shot 2022-08-08 at 17 40 36

pjrobertson:
The tests on CI are still failing. See: https://github.com/quicksilver/Quicksilver/runs/7736472992?check_suite_focus=true#step:3:33

Is there a reason you reverted "[Revert "Disable parellalize builds for preflight"](https://github.com/quicksilver/Quicksilver/pull/2916/commits/7f74f6e54ad431cf55e655de84565a5d828aa67d)? I'm wondering if that's the problem

I tested this on my local Mac though, and can confirm things still build.

n8henrie:
> The tests on CI are still failing.

Quick push before had to run for dinner. I'll keep iterating on it -- but for the first time (ever) the tests are actually running on my machine, so I'm excited. There are several test failures -- which also makes me excited.

> Is there a reason you reverted "[Revert "Disable parellalize builds for preflight"](https://github.com/quicksilver/Quicksilver/pull/2916/commits/7f74f6e54ad431cf55e655de84565a5d828aa67d)?

Because it didn't seem to change anything on my end, and I'd rather isolate the problem rather than change multiple variables, and I'd rather have faster builds if it ends up not being an issue.

> I'm wondering if that's the problem

I don't think so -- in fact I doubt it even has an any effect on newer machines (since it's not a user-facing setting anymore).

pjrobertson:
Sure thing, as long as we can get the tests running successfully on CI, then I'm happy! 

FYI: the 3 tests you had that failed in QSFileObjectTests all pass for me. So it may just be a naming convention on newer Macs. Here's mine:

Screen Region 2022-08-09 at 09 01 49


Also - I pushed two more commits here since I added my own unit test, and thought it'd be easier to add it here than to master :)

pjrobertson:
All these tests pass for me (except for one). Amazing! I'm going to merge this :)

This one test fails, I think it's because I'm running unsigned. Should probably work on signed.

Screenshot 2023-08-20 at 14 53 38

PR #2927, opened 2022-10-13
Enable Renovate

gliptak:
https://github.com/marketplace/renovate


n8henrie:
Thanks! We should definitely get the tests passing first, but I think that's close.

gliptak:
please reopen when build is green

PR #2938, opened 2022-12-18
keep debug version

n8henrie:
- Remove the key when we're done with it... just in case.
- This should upload a debug artifact with each release, yet to be tested.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,3 +133,4 @@ jobs:
             -o StrictHostKeyChecking=no \
             "${GITHUB_REPOSITORY}" \
             <Quicksilver/SharedSupport/ChangesBare.html
+          rm /tmp/key
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,11 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           path: /tmp/QS/build/Release/checksum.txt
+      - name: Download debug artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: Quicksilver-debug.zip
+          path: /tmp/Quicksilver-debug.zip
       - name: Release
         uses: softprops/action-gh-release@v1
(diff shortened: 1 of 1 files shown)


n8henrie:
@skurfer we're getting:

>Error: HTTP status code: 403. A required agreement is missing or has expired. This request requires an in-effect agreement that has not been signed or has expired. Ensure your team has signed the necessary legal agreements and that they are not expired.

Are we okay to keep using your certificate? If so, do you get any notification when you log into your Apple Dev account?

skurfer:
Sorry, just now getting to this. The developer account had expired. I just renewed it. I’ll e-mail you my cell so you can get to me faster for stuff like this.

n8henrie:
I'll probably cut a fake release to make sure this works and if so merge.

I'm really hoping a debug version available for download helps sort out these intermittent crashing issues.

pjrobertson:
The code here looks good to me. However it appears the developer cert still needs updating.

@n8henrie did you manage to get the login details from @skurfer for the dev account?

If yes: Rob - you need paying for this year.
If no: We can set up a new developer account to get builds signed (regardless, Rob should still be reimbursed).

n8henrie:
@pjrobertson @skurfer I did not get account details, Rob renewed and I suspected it would just start working again, but the signing is giving an error:

> A required agreement is missing or has expired. This request requires an in-effect agreement that has not been signed or has expired. Ensure your team has signed the necessary legal agreements and that they are not expired.

Rob, if you log into the account, does it present you with some kind of EULA?

skurfer:
I accepted the agreement and `qssign` is working locally for me.

n8henrie:
Signing is not working, thanks @skurfer 

Looks like I need to fix the debug artifact: `Error: Unable to find an artifact with the name: Quicksilver-debug.zip`

n8henrie:
Hope nobody minds the quickfix pushes to main for this, CI stuff is kind of painful to work with via branches (since the branch itself often determines the CI process).

PR #2952, opened 2023-08-20
Switch to NSRunningApplications in QSProcessMonitor. Fixes #2919

### LINKED ISSUE #2919: [Bug]: App specific triggers sometimes do not work right after starting the app.

- [X] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [X] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [X] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

I have a trigger who's scope is limited to one app.
When I start this app, the trigger does not reliably work.
It does work after I switched to a different app and returned to the specific app.

### Steps to reproduce

1. Add a trigger
2. Limit its scope to an App
3. Start the App
4. Try the trigger

### Expected behavior

The trigger should work each time.

### MacOS Version

macOS 12

### Quicksilver Version

Checked out main/Released 2.4.0 (4039)

### Relevant Plugins

-/-

### Crash Logs or Spindump

-/-

### Screenshots

-/-

### Additional info

1) `QSProcessMonitor.m` handles both `appChanged` (line 51) and `appLaunched` (line 70) carbon process events.
2) The `appLaunched` handler schedules an async task on the main thread that posts a notifications on the default notification center.
3) This notification is handled _in the future_ on the main thread and updates the `processesDict` (in `reloadProcesses`, line 362).
4) The `appChanged` handler might be called before step 3 takes place. `appChanged` uses `processObjectWithPSN` to get the process information for the newly active app which is not yet stored in `processesDict`. The result of this call is attached as user info to another notification posted (one of the listeners to this notification updates which triggers are active).

Since in step 4, the `processesDict` has not been updated in all cases, the newly active app is not known and the triggers do not get updated.

Note: `appLaunched` calls `processObjectWithPSN` (in line 80) using the PSN of the new app. This call always returns `nil`.

I see two ways to fix this:

1) Lazy fix: Drop the call to `processObjectWithPSN` in `appLaunched` carbon event handler and move the call to `processObjectWithPSN` onto the main thread in `appChanged`.

2) Clean fix: `reloadProcesses` mentions a `addProcessWithPSN` which I could not find. If this existed, it should be called in `appLaunched` (instead of line 80, if it returns the new dictionary) to update the `processesDict` right away. Everything else could stay the same.

pjrobertson:
This replaces #2921

Code could still be tidied up, but this at the very least should avoid race conditions with reloading processes.

@n8henrie if you're able to test, please do! This should hopefully fix what we looked at together late last year ;-)


pjrobertson:
Confirmed working by dhoepfl . Meging

Issue #2953, opened 2023-08-20
[Bug]: Tests fail due to using `~/Library/Application Support`


n8henrie:
### Before submitting your bug report, please confirm you have completed the following steps

- [X] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [X] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [X] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

Continuation of some work from https://github.com/quicksilver/Quicksilver/issues/2583

Some tests -- specifically `testAddingFileObjectToPasteboard` in my case -- fail for my locally because they refer to QSObjects that have already been cached somewhere in `~/Library/Application Support`. This error will likely not show up in CI because it will be using an empty directory there regardless.

When running tests, having the tests use a fake directory for App Support helps ensure reproducibility.

### Steps to reproduce

Run the `testAddingFileObjectToPasteboard` while you already have an object for `/Applications/Safari` in your `~/Library/Application Support` directory

### Expected behavior

Tests should pass

### MacOS Version

None

PR #2954, opened 2023-08-20
Mock Application Support when running tests

n8henrie:
Fixes https://github.com/quicksilver/Quicksilver/issues/2953

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSPaths.m
+++ b/Quicksilver/Code-QuickStepCore/QSPaths.m
@@ -11,6 +11,13 @@
 
 NSString *QSGetApplicationSupportFolder() {
 	static NSString *QSApplicationSupportPath = nil;
+
+	// https://github.com/quicksilver/Quicksilver/issues/2953
+	#ifdef TESTING
+		NSLog(@"Override QSApplicationSupportPath for testing");
+		QSApplicationSupportPath = [NSTemporaryDirectory() stringByAppendingPathComponent:@"QSTest_Application_Support"];
+	#endif
+
 	if (!QSApplicationSupportPath) {
 		NSString *path = [[[NSUserDefaults standardUserDefaults] stringForKey:@"QSApplicationSupportPath"] stringByStandardizingPath];
 		if (!path) {
--- a/Quicksilver/Quicksilver Tests/Quicksilver_Tests.m
+++ b/Quicksilver/Quicksilver Tests/Quicksilver_Tests.m
@@ -72,48 +72,9 @@
     XCTAssertEqualObjects(textString, pboardString);
 }
 
-// TODO - these tests currently fail
-- (void)testAddingCombinedObjectsToClipboard {
-    return;
-    // create three temporary files in a subfolder of the temp folder, created with a random string
-    NSString *tempDir = [NSTemporaryDirectory() stringByAppendingPathComponent:[[NSProcessInfo processInfo] globallyUniqueString]];
-    [[NSFileManager defaultManager] createDirectoryAtPath:tempDir withIntermediateDirectories:YES attributes:nil error:nil];
-    NSString *str1 = [tempDir stringByAppendingPathComponent:@"test1.txt"];
-    NSString *str2 = [tempDir stringByAppendingPathComponent:@"test2.txt"];
-    NSString *str3 = [tempDir stringByAppendingPathComponent:@"test3.txt"];
-    // create the 3 files
-    [@"" writeToFile:str1 atomically:YES encoding:NSUTF8StringEncoding error:nil];
-    [@"" writeToFile:str2 atomically:YES encoding:NSUTF8StringEncoding error:nil];
-    [@"" writeToFile:str3 atomically:YES encoding:NSUTF8StringEncoding error:nil];
-
-    QSObject *obj1 = [QSObject fileObjectWithPath:str1];
-    QSObject *obj2 = [QSObject fileObjectWithPath:str2];
-    QSObject *obj3 = [QSObject fileObjectWithPath:str3];
-    
-    QSObject *combinedObj = [QSObject objectByMergingObjects:@[obj1, obj2, obj3]];
-    XCTAssertEqual([combinedObj count], 3);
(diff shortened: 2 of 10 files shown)


n8henrie:
This PR adds one "me-too" commit that unblocks tests that are currently not running. We should either fix these or remove them!

@pjrobertson for the life of me I can't get `testAddingCombinedObjectsToClipboard` to pass. I think maybe `[pboard.pasteboardItems[index] types]` gives the intersection of types instead of the union? Unsure. It looks like you added this test, can you see if it should be removed or fixed?

Quicksilver/Code-QuickStepCore/QSPaths.m
```diff
@@ -25,6 +25,15 @@
 		}
 		QSApplicationSupportPath = path;
 	}
+
+
+	bool isTest = NSClassFromString(@"XCTestCase") != nil;
```
pjrobertson, line 30: Hmm... we shouldn't be putting test code into the live app code. This isn't good practice. If the test isn't working, then the test needs changing, not the code. Behaviour should be the same between tests *and* live code.

n8henrie: I agree in the general case!

In this case, the test in question is an integration test that seems to load a full normal load of Quicksilver, which means that it is subject to all of the impurities and side effects of loading whatever plugins and settings the runner of the test has configured.

To the extent possible, I think we *should* have a way to ensure that our test results are not contaminated by our individual testing environments. Test failures that aren't reproducible seem like a problem!

I usually would approach this with some kind of dependency injection (for example rewriting this into a function that accepts an optionally passed in path to `Application Support` and defaults to `~/Library/Application Support`). I also looked into trying to override and restore the NSUserDefaults in a test setup / teardown function, but it looks like the path for doing this is in ObjC is relatively complicated (with some even creating [small GitHub projects](https://github.com/jakob/NSUserDefaultsOverride) for the overrides they use). What's worse it is seems that they could potentially mess up someone's Quicksilver setup, for example if XCode crashes (which never happens, right?) before the test teardown code and the NSUserDefaults were not restored.

The more I think about it, it would make sense for us to have a way to test QS with different configurations / settings. What other ideas do you have about an approach we could take here?

pjrobertson: You are spot on that running tests with the user configuration isn't ideal/shouldn't be done. Using a fixed, standard group of settings is much better. In this case we should really use `setUp` and `tearDown` to adjust the location of the application settings. More info: https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/UnitTesting/03-Writing_Test_Case_Methods/writing_tests.html

We already have a method for getting the app support folder (see `QSGetApplicationSupportFolder()` in QSPaths.m). If you look there, it first checks the userDefaults for `QSApplicationSupportPath`. If it finds it, then that is used as the application support path. As such, what I suggest is:

```
NSString *oldAppSupportPath = nil;
NSString *testAppSupportPath = @"/my/temp/test/folder";

- (void)setUp {
oldAppSupportPath = [[NSUserDefaults standardUserDefaults] stringForKey:@"QSApplicationSupportPath"];
[[NSUserDefaults standardUserDefaults] setObject:testAppSupportPath forKey:@"QSApplicationSupportPath"]
}
 
- (void)tearDown {
[[NSUserDefaults standardUserDefaults] setObject:oldAppSupportPath forKey:@"QSApplicationSupportPath"]
}

```

Alternatively you could just modify `QSApplicationSupportPath()` to do a compile-time test (using a macro) like this: https://innovaptor.com/blog/2013/09/02/xcode-preprocessor-macros-for-test-code/

Note: looking back at this now, I really don't think it's actually worth our time to try and get existing unit tests working. We never used them, so there's not really much point going back and trying to fix them (I spent an hour trying to do that today, which could have been spent fixing other bugs). I think we should just start writing unit tests when we come across issues, and delete any that don't work.

n8henrie: > Note: looking back at this now, I really don't think it's actually worth our time to try and get existing unit tests working.

I agree! I've spent several *months* trying to get this working (not sure if you or @skurfer saw my email from May), as having a passing test suite would make me feel much more comfortable working with ObjC, since I know so little of it.

I'll submit a separate PR to remove failing tests and close this one.

n8henrie: Also, my concern with the setup / teardown `NSUserDefaults` changes is that I *think* `NSUserDefaults` is persistent (which is why we need the teardown code); if there were a crash that prevented teardown from running, then one would be stuck with a Quicksilver that would try to use the test config (until someone manually fixed this setting or ran the test suite again). Right?

pjrobertson: True, best not to use that. Go with modifying `QSApplicationSupportPath()` at compile time. Very similar to what you have, but don't do a runtime check `if (XXX)` but a compile time check `#ifdef XXXX`. See:

https://innovaptor.com/blog/2013/09/02/xcode-preprocessor-macros-for-test-code/

n8henrie: Thanks for the tip! Done.

n8henrie:
Rebased on main. @pjrobertson does this look ready for merge?

n8henrie:
As a bonus, with this PR builds for me are *tremendously* faster and seem to be somewhat more reliable (fewer "I guess I'll clean and try again"s), likely due to enabling parallel build and automatic build dependency ordering.

n8henrie:
@pjrobertson if reasonable, I'd like to merge this before fixing https://github.com/quicksilver/Quicksilver/issues/2958 to avoid having to backtrack to fix a few merge conflicts.

pjrobertson:
I wrote a comment but forgot to "submit for review". Once they small change is made, feel free to merge!

n8henrie:
AFK until Sunday, should be able to merge this and resolve the ChangesBare.html issue then / soon. Thanks!

PR #2955, opened 2023-08-28
Debug uploading the debug artifact

n8henrie:
This branch was to help debug CI for https://github.com/quicksilver/Quicksilver/pull/2938#issuecomment-1694694593

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,17 +17,17 @@ jobs:
     env:
       QS_BUILD_ONLY: 1
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: recursive
       - name: Build debug version
         working-directory: Quicksilver
         run: ./Tools/qsrelease Debug
       - name: Upload debug version
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
-          name: Quicksilver-debug.zip
-          path: /tmp/QS/build/Debug/Quicksilver*.dmg
+          name: Quicksilver-debug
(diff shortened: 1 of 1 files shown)


n8henrie:
I think this is ready for merge. It uploads a debug executable along with the release artifacts, which should hopefully empower end users to help us debug issues.

n8henrie:
I've also updated the CI dependencies in this PR, which silences several annoying warnings.

pjrobertson:
I see Quicksilver-debug here :) https://github.com/quicksilver/Quicksilver/actions/runs/6007687787

Can it be given an extension in the name. Is it a .zip or .dmg?

pjrobertson:
@n8henrie  - waiting on this change (Adding an extension) then I'll push a new release out :)

n8henrie:
@pjrobertson -- I actually removed the `.zip` extension due to: https://github.com/actions/upload-artifact#zipped-artifact-downloads:

> There is currently no way to download artifacts after a workflow run finishes in a format other than a zip or to download artifact contents individually. One of the consequences of this limitation is that if a zip is uploaded during a workflow run and then downloaded from the UI, there will be a double zip created.

So I had originally made it so that one saw `foo.zip` in the web interface, but once downloaded it showed as `foo.zip.zip` which I thought seemed uglier than having `foo` in the web interace and `foo.zip` once downloaded. I still think the latter is less ugly, but if you think it's more important to show the `.zip` in the web interface I'm happy to revert that.

pjrobertson:
Makes sense. I see that when I download it it is correctly named Quicksilver-debug.zip, so I think this change is fine.

Merging :)

PR #2959, opened 2023-09-06
Fix version string task

n8henrie:
Add quoting to argument to `defaults`, fixing local `Archive` tasks


n8henrie:
This just quotes a previously unquoted bash argument, so I'm going to go ahead and merge.

Issue #2971, opened 2023-09-30
[Bug]: Multiple Quicksilver_Tests fail when run as part of Quicksilver scheme


n8henrie:
### Before submitting your bug report, please confirm you have completed the following steps

- [X] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [X] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [X] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

Of the tests for the Quicksilver scheme, these three fail when run directly:

- `testClearingFirstPane`
- `testClearingSearchStringOnTrigger`
- `testThirdPaneClosingBehaviour`

All tests, including these, pass when run as part of the `Quicksilver Distribution` scheme.

Why is this?

### Steps to reproduce

1. In Xcode, select the `Quicksilver` scheme
2. Run tests
3. Observe that these three tests fail

### Expected behavior

These tests should pass when run in isolation.

### MacOS Version

None

### Additional info

These tests fail as part of the `Quicksilver Distribution` depending on the order in which they are run; if one selects the option to `randomize test order`, they fail frequently.

The common theme between these three is:

```objc
QSInterfaceController *i = [(QSController *)[NSApp delegate] interfaceController];
QSCollectingSearchObjectView *dSelector = (QSCollectingSearchObjectView *)[i dSelector];
```

When run as part of the `Quicksilver Distribution`, `*i` is non-nil. When run as part of `Quicksilver`, this is a nil pointer.

Similarly, `[QSReg preferredCommandInterface]` is a nil pointer when run from `Quicksilver` but non-nil with `Quicksilver Distribution`:  https://github.com/quicksilver/Quicksilver/blob/main/Quicksilver/Code-App/QSController.m#L784

Having a hard time chasing this down, it *seems* like these tests should be set up to pass from this scheme (since they pass when run from `Quicksilver Distribution`); if there is some fundamental reason why they won't, perhaps they should be moved?

Any pointers from @pjrobertson @skurfer ?

n8henrie:
The `Quicksilver` test `testActionsForURLObject` fails if run immediately after a clean but succeeds if I first change schemes and run tests for `Core Support`.

n8henrie:
[...]

PR #2979, opened 2023-10-18
Run all tests

n8henrie:
Fixes https://github.com/quicksilver/Quicksilver/issues/2971

This PR includes a number of changes mostly to the tests, CI, and build toolchain.

It does add a few `ifdef TESTING` to avoid things like the accessibility checker on CI, which otherwise blocks forever and then fails.

It add a new plist file `QSCorePlugIn-Info-Testing.plist` to try to avoid indexing "sensitive" directories like `~/Desktop` and `~/Documents` on startup, which also result in popups that can block further execution.

In several places the default build interface was supposed(?) to be Bezel but was set to Primer; this changes it to Bezel.

It changes several build steps to allow automatic dependency resolution to work, specifically by removing dependencies like the Bezel interface from the `Quicksilver` target (that I think I put there in the first place) and putting them under `Quicksilver Tests`. (They are still under `Quicksilver Distribution`, so I don't think this should change peoples dev workflow.) It also enables parallel testing, which in addition to parallel builds *really* speeds up the build process on my machine, and configures the testing order to be randomized, to hopefully avoid issues with race conditions between tests that just *happen* to work in our favor when build alphabetically.

There are a lot of commits, several of which could be dropped (later reverted), but as I went through rebasing and squashing I decided it would be best to leave them separate to enable bisecting in the future if any of them end up being problematic (as opposed to squashing them together and trying to tease the changes apart later).

Finally, this adds a nice little build status badge to the readme and removes a few whitespace errors.


n8henrie:
@pjrobertson @skurfer I'd like to merge this soon, if you get a chance to take a look, so I can move on to several of the Sonoma bugs, while hopefully ensuring that the fixes aren't unintentionally resulting in backwards incompatible changes.

n8henrie:
Merging this in to avoid rebase hell.

Issue #2970, opened 2023-09-29
[Bug]: Extensions appear on applications


skurfer:
### Before submitting your bug report, please confirm you have completed the following steps

- [X] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [X] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [X] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

After installing 2.4.1, a few applications in the catalog were showing up with the `.app` extension. At this point, they all show it.

I’m looking through the diff between 2.4.0 and 2.4.1 to see if I can figure out what happened, but if someone who worked on it has any immediate theories, feel free to chime in.

### Steps to reproduce

1. Invoke Quicksilver
2. Search for an application

### Expected behavior

the name of the application will appear without `.app`. I forget if this respects the Finder preference for extensions or not. All I know is I didn’t see them before and now I do.

### MacOS Version

Other

### Quicksilver Version

2.4.1

### Screenshots

<img width="408" alt="Screenshot 2023-09-29 at 4 35 01 PM" src="https://github.com/quicksilver/Quicksilver/assets/154676/9c79ba34-f442-4e2e-8b6e-c9c8c13e9502">

n8henrie:
I wouldn't be surprised if it was my commit here: https://github.com/quicksilver/Quicksilver/commit/260eee39000984b5f93ed15c0781ab7cba98204e

Might be a good place to start looking at least.

n8henrie:
I've reverted that commit [here](https://github.com/quicksilver/Quicksilver/actions/runs/6363050382) -- wanted to see if that fixes it?

skurfer:
That doesn’t seem to be it. I’ll revert back to 2.4.0 for a minute to make sure I’m not crazy.

pjrobertson:
Fixed by [9422d63](https://github.com/quicksilver/Quicksilver/commit/9422d6331c4fcf7d12472f8b98f7fc5ad6a3a435)

Issue #2917, opened 2022-08-09
[Bug]: Names of running background applications show incorrectly


g1oria:
### Before submitting your bug report, please confirm you have completed the following steps

- [X] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [X] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [X] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

Why is it that when I "show contents" for "Running Applications & Processes" it shows several running Firefox apps in the results?

![unnamed](https://user-images.githubusercontent.com/17694864/183699966-614b81ae-f3cf-4b3b-8c76-777cceb6aa81.png)

![unnamed-1](https://user-images.githubusercontent.com/17694864/183700016-932a4124-e2b0-4207-92e9-702c3099ee74.png)

The weird thing is if I uncheck the box: Include background applications, Firefox doesn't show up at all.

![screenshot-08-08-2022-09-34-00](https://user-images.githubusercontent.com/17694864/183700111-48d87991-390e-460e-bdd5-684f9702f0f1.png)

You can see the conversation about this issue in the QS Google Group [here](https://groups.google.com/g/blacktree-quicksilver/c/2yQso_U647E).

### Steps to reproduce

1. Type "Running Applications" . . .
2. Choose "Show Contents" in 2nd pane
3. Multiple instances of Firefox show up in the results.

### Expected behavior

1 instance of Firefox, rather than many. 1 instance of Firefox even when "include background applications" is unchecked.

### MacOS Version

macOS 10.15

### Quicksilver Version

2.4.0 (4039)

pjrobertson:
Thanks Gloria for reporting. This has now been fixed, see 8f1eb8a617952185abd7a921528dc7eb53a6b9be for full information

g1oria:
Ok. Where do I get the update? Because, according to the releases on the
Github page, the last update was 29 days ago.

pjrobertson:
We'll build and release a new version of Quicksilver in the coming days. If you really want to run it now, you can download a pre-release version from: https://github.com/quicksilver/Quicksilver/actions/runs/2825754664

g1oria:
Ah, ok. Thank you.Message ID:
***@***.***>

>

PR #2992, opened 2023-12-02
Don't override names for running processes

### LINKED ISSUE #2975: [Bug]: Since 2.4.1 localized application names are not recognized anymore

- [X] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [X] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [X] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

After the update from 2.4.0 to 2.4.1 localized app names are not recognized anymore when typing.
In 2.4.0 i can type "Vorschau" (preview in german) to select Preview.app. Though i could also type "preview", where Preview.app would be selected.

Now if i type "Vorschau" it isn't recognized at all anymore. Only if i type "preview", then Preview.app gets selected.

I downgraded (after a month of using 2.4.1) back to 2.4.0 and the localized names were working right out of the box again.

### Steps to reproduce

In MacOS with german localization try to select an app via its localized name. The localized name seems to be unknown to Quicksilver 2.4.1
Working fine in 2.4.0

### Expected behavior

Quicksilver indexes localized App strings

### MacOS Version

macOS 13

### Quicksilver Version

_No response_

### Relevant Plugins

_No response_

### Crash Logs or Spindump

_No response_

### Screenshots

_No response_

### Additional info

_No response_

pjrobertson:
Fixes issue #2970

Note: This *re-introduces the bug fixed in #2917 - that bug is actually a Finder issue and not a QS issue

I thought I'd make a PR for this, so @skurfer @n8henrie can test. This should also fix #2975

Note that: this 'fix' re-introduces the bug fixed in #2917 - but upon further investigation, that is a Finder/macOS bug. The example used in the bug report was for Firefox. The example I found was for WeChat:
Screenshot 2023-12-02 at 16 06 25

This .app bundle inside the main WeChat.app shows up as `WeChat` in Finder. However if you open the .app package and look at the Info.plist, the actual name should be `WeChatAppX` and the actual path in terminal is `WeChatAppX.app` not the supposed `WeChat.app` as shown in Finder:

Screenshot 2023-12-02 at 16 06 21

Screenshot 2023-12-02 at 16 06 43

This is a problem with macOS itself, more specifically the `NSURLLocalizedNameKey` resource key on NSURL. I'm inclined to re-open #2917 but then mark it as "won't fix" because yeah, macOS is buggy.

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSProcessMonitor.m
+++ b/Quicksilver/Code-QuickStepCore/QSProcessMonitor.m
@@ -181,7 +181,6 @@ OSStatus appTerminated(EventHandlerCallRef nextHandler, EventRef theEvent, void 
             // the process isn't an app
             newObject = [QSObject fileObjectWithPath:[dict objectForKey:@"CFBundleExecutable"]];
         }
-		[newObject setLabel:[dict objectForKey:@"CFBundleName"]];
     }
 
 	[newObject setObject:dict forType:QSProcessType];


pjrobertson:
P.S. - to test this make sure you delete `~/Library/Caches/Quicksilver` first before launching Quicksilver.

Otherwise QS will still use the 'old' cached objects from before.

n8henrie:
Test would be great!

Sometimes hard to pick a good target though, as the application paths have changed over time, and certain paths require clicking the "QS would like to access files on your ~/{Desktop,Documents}" privacy popup.

One approach is to use the QS itself, as we can guarantee that it exists by the time a test is run, and we control its path, and have permission to read rules in `/tmp/QS/build/Testing` or what have you.

n8henrie:
No worries!So far one of my least favorite parts of testing QS is the global nature of NSUserDefaults and how dependency injection for tests is just not a thing. Would probably require snaking extra parameters through a million lines of code to fix!On Dec 2, 2023, at 7:29 AM, Patrick Robertson ***@***.***> wrote:﻿

Test would be great!

This bug was only happeing for running applications and if you had the 'running applications and processes' catalog entry checked. I think there's already a test for the name of a certain app, but adding a test for this would require:
a) make sure 'running applications and processes' is enabled
b) test the name before app X is launched
c) launch app X
d) make sure QS has re-scanned the catalog
e) test the name after app X is launched
That's going to give me a headache :'(

n8henrie:
> P.S. - to test this make sure you delete `~/Library/Caches/Quicksilver` first before launching Quicksilver.

This seems like an easy footgun to fix. Maybe we should `[[NSFileManager defaultManager] removeItemAtPath:[@"~/Library/Caches/Quicksilver" stringByStandardizingPath] error:&err]` in a setup function for tests?

n8henrie:
Hmm, looking at #2970, I can't reproduce the issue.

`rm -rf ~/Library/Caches/com.blacktree.Quicksilver/ ~/Library/Caches/Quicksilver/ /tmp/QS` and build QS from current main (f5f207ad057551fed5241816a60d0af3c228d41a), and I see the following:

Screenshot 2023-12-02 at 11 16 04

Tried launching Mail.app and I see no difference in behavior.

I don't have WeChat installed. Is there another example app?

pjrobertson:
Try launching Mail.app before you start Quicksilver. Also make sure you
don't have any other versions if Quicksilver running, and that you delete
the caches *after* quitting any running QS instances

n8henrie:
> Try launching Mail.app before you start Quicksilver. 

That did it -- and the fix works. Thanks!

PR #2996, opened 2023-12-09
Force-disable superfluous effects on startup

n8henrie:
I think this is a reasonable move to automatically help people experiencing
show-stopper bugs on Sonoma.

The `Use Visual Effects` seems to even make first-time setup and installation
impossible on Sonoma in some cases; hopefully I put this early enough in
startup to catch these cases.


n8henrie:
@pjrobertson @skurfer I'd like to cut a bugfix release with this temporary workaround. Thoughts?

PR #3002, opened 2024-02-06
remove private apis

n8henrie:
- Remove the private APIs entirely
- Revert "#FIXME: Force-disable superfluous effects on startup"

Attempt at removing private APIs entirely which have previously been used for
animations.

I believe there is a *lot* more relevant code that could be removed, but after
removing the culprit files this seems to be the minimum needed to get QS to
build and run.


pjrobertson:
LGTM. Goodbye QGCPrivate :'(

pjrobertson:
@n8henrie - please merge if this is ready. Only thing I can think of is to do a check in all the plugins for calls to `CGSxxxxx` and remove from plugins as well.

n8henrie:
There is some residual GUI funkiness after this PR, and I've got a long way before I understand much of the GUI stuff, but I think it's worth merging anyway, as everything else from this point onward will require a ton of rebasing otherwise.

Issue #3008, opened 2024-05-15
[Bug]: QSProcessMonitor crashes Quicksilver


n8henrie:
### Before submitting your bug report, please confirm you have completed the following steps

- [X] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [X] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [X] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

As per the title -- we have a number of crashes in QSProcessMonitor. This will be a meta-issue with links to a few relevant issues.

For example, I get a `EXC_BAD_ACCESS` at https://github.com/quicksilver/Quicksilver/blob/f4f679f3a9a0dc377214e150761ff633a9807a13/Quicksilver/Code-QuickStepCore/QSObject.m#L412

My 

### Steps to reproduce

1. Start QS 
2. Enable `Running Applications & Processes` in `Settings` -> `Catalog` -> `Applications`
3. Wait a few minutes to hours

### Expected behavior

No crash

### MacOS Version

Other

### Quicksilver Version

2.4.2

### Crash Logs or Spindump

The callstack during a crash:

```
#1	0x0000000184f77d90 in -[__NSDictionaryM objectForKey:] ()
#2	0x0000000105efa56c in -[QSObject _safeObjectForType:] at /Users/n8henrie/git/qs/quicksilver/Quicksilver/Code-QuickStepCore/QSObject.m:416
#3	0x0000000105efa610 in -[QSObject objectForType:] at /Users/n8henrie/git/qs/quicksilver/Quicksilver/Code-QuickStepCore/QSObject.m:429
#4	0x0000000105f2a6a0 in __54-[QSProcessMonitor processObjectWithPSN:fromSnapshot:]_block_invoke at /Users/n8henrie/git/qs/quicksilver/Quicksilver/Code-QuickStepCore/QSProcessMonitor.m:207
#5	0x0000000184f98904 in __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__ ()
#6	0x0000000184fbe2f0 in __NSDictionaryEnumerate ()
[...]

PR #3010, opened 2024-06-02
process monitor crashes

n8henrie:
- **Simplify locks with @synchronized**
- **Use a threadsafe dict for data as well as cache**

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSObject.m
+++ b/Quicksilver/Code-QuickStepCore/QSObject.m
@@ -53,8 +53,8 @@ NSSize QSMaxIconSize;
 - (id)init {
 	if (self = [super init]) {
 
-		data = [NSMutableDictionary dictionaryWithCapacity:0];
-		meta = [NSMutableDictionary dictionaryWithCapacity:0];
+		data = [QSThreadSafeMutableDictionary dictionaryWithCapacity:0];
+		meta = [QSThreadSafeMutableDictionary dictionaryWithCapacity:0];
 		cache = [QSThreadSafeMutableDictionary dictionaryWithCapacity:0];
 		name = nil;
 		label = nil;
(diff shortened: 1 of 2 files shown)


n8henrie:
This is intended to fix https://github.com/quicksilver/Quicksilver/issues/3008

pjrobertson:
Hey Nate 👋 :-)

Does this actually fix the problem? Have you been running with this and do you still experience crashes? If not - then it looks good to me. My only concern is: does it cause any locks anywhere else in the code?

n8henrie:
Hi Pat!

> Does this actually fix the problem?

No, unfortunately. I think I have mostly / partially reverted it in my local branch, will convert this to a draft to reduce confusion.

I've finally discovered xcode's thread sanitizer, which I think will be helpful in pinpointing areas of concern, but I've also been busy with a move and an upcoming board exam, so it's been slow going.

It also seem that thread sanitizer may have some false positives, as I've spent way too much time trying to hunt down the reported race condition at https://github.com/quicksilver/Quicksilver/blob/f4f679f3a9a0dc377214e150761ff633a9807a13/Quicksilver/Code-QuickStepCore/QSLibrarian.m#L362, but all I've figured out is that it goes away when I remove `NSEnumerationConcurrent` (and returns afterward). Either that or it's somewhere in https://github.com/quicksilver/Quicksilver/blob/f4f679f3a9a0dc377214e150761ff633a9807a13/Quicksilver/Code-QuickStepCore/QSCatalogEntry.m#L734 -- I think..

pjrobertson:
Hey hey!

Darn, that's a shame it doesn't work. Although I didn't think this would work - we used thread safe dicts in the past for QSProcessMonitor, but I actually took them out as they weren't helping :(

Good to hear you've got the thread sanitizer working! I wonder if ChatGPT can analyze and fix the code for us...? ;-)

n8henrie:
> I wonder if ChatGPT can analyze and fix the code for us...?

I've tried! Unfortunately doesn't seem to know ObjC very well either. I tried (repeatedly) to get it to recreate a toy program that would crash with `EXC_BAD_ACCESS` due to a race condition, but the programs it generated all ran without errors or crashing.

n8henrie:
@pjrobertson actually, I take it back -- this seems to mitigate the issue at least somewhat. Without this I'm getting crashes within an hour or two, with this it's been running for 2 days without a crash. Let's merge this for now and see if it at least mitigates the issue.

PR #3025, opened 2025-02-01
[WIP] Fixes for Sequoia

pjrobertson:
* Fix main thread issue + better message when tryig to copy to (protected) /Applications folder
* Fix dragging images into apps (e.g. Whatsapp) (Todo: switch to NSPasteboardItem)
* Remove superfluous visual effects, fixes #3025
* Fix making accessibility prompt focusable


Quicksilver/Code-QuickStepCore/QSObject_FileHandling.m
```diff
@@ -243,9 +248,16 @@ - (BOOL)loadIconForObject:(QSObject *)object {
 	NSFileManager *manager = [NSFileManager defaultManager];
     
 	// the object isn't a file/doesn't exist, so return. shouldn't actually happen
-	if (![manager fileExistsAtPath:path]) {
+	__block BOOL exists = YES;
+	QSGCDMainSync(^{
```
n8henrie, line 252: I assume this change makes this run on the main thread instead of a lower priority background thread?

Quicksilver/Code-QuickStepCore/QSObject_Pasteboard.m
```diff
@@ -221,125 +270,27 @@ - (void)guessName {
...
-        NSArray *fileURLs = [[self validPaths] arrayByEnumeratingArrayUsingBlock:^NSString *(NSString *path) {
-            return [[NSURL fileURLWithPath:path] absoluteString];
-        }];
-        [self setObject:fileURLs forType:@"public.file-url"];
-    }
-	// last case: no other useful types: return a basic string
-	if (!includeTypes) {
-		includeTypes = @[NSStringPboardType, QSTextType];
-	}
+	NSLog(@"WARNING: This method is deprecated, ignoring the declareTypes and includeTypes parameters");
```
n8henrie, line 280: Should we just get rid of the additional instances? Or is the issue that it might be called from plugins? (In which case maybe another benefit of monorepo, being able to more confidently change our API.)

Quicksilver/Code-QuickStepCore/QSObject_Pasteboard.m
```diff
@@ -221,125 +270,27 @@ - (void)guessName {
...
-		[pboard addTypes:[NSArray arrayWithObject:QSPasteboardObjectIdentifier] owner:self];
-        [self writeToPasteboard:pboard data:[self identifier] forType:QSPasteboardObjectIdentifier];
-	}
-	
-	[pboard addTypes:[NSArray arrayWithObject:QSPasteboardObjectAddress] owner:self];
-    QSLib.pasteboardObject = self;
-	//  NSLog(@"types %@", [pboard types]);
-	return YES;
+- (BOOL)putOnPasteboard:(NSPasteboard *)pboard includeDataForTypes:(NSArray *)includeTypes {
+	NSLog(@"WARNING: This method is deprecated, ignoring the includeTypes parameter");
```
n8henrie, line 285: I don't see any other calls to this in the main QS codebase at first glance.

n8henrie (approved):
Wow, tremendous effort. Thanks Paddy!

Quicksilver/Code-QuickStepCore/QSObject_Pasteboard.m
pjrobertson: Exactly - I originally just deleted them but then clipboard plugin crashed, because it was using it. If we have a monorepo we don't need to worry too much because we can just do a global search and replace, but still - we shouldn't just delete methods (like I did!) because if someone is running an old QS with a new plugin or vice versa it will cause a crash.

Quicksilver/Code-QuickStepCore/QSObject_Pasteboard.m
pjrobertson: Clipboard was using it (but I fixed that). I don't know about other plugins...

pjrobertson:
Thanks for approving! I'd like you to run it for a while and see how things go, then we can do a pre-release!

Quicksilver/Code-QuickStepCore/QSObject_Pasteboard.m
pjrobertson: (Oh and another example of why deleting methods is bad - when the `[QSWindow flare]` method was deleted as part of the CGSPrivate, because it no longer existed it broke the yosemite interface (now fixed)

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSObjectActions.m
```diff
@@ -41,7 +41,7 @@ - (NSArray *)validActionsForDirectObject:(QSObject *)dObject indirectObject:(QSO
 
 		[newActions addObject:kQSObjectShowChildrenAction];
 	}
-    if ([[QSLibrarian sharedInstance] firstEntryContainingObject:dObject])
+	if ([dObject objectForMeta:QSCatalogStructureChanged] || [[QSLibrarian sharedInstance] firstEntryContainingObject:dObject])
```
n8henrie, line 44: I'm wondering if this is why I'm not seeing `show source in catalog` for those odd catalog entries (that look like they're bullet points from some kind of note file)?

n8henrie: My results for `asdf` (which is my synonym for clipboard history, but none of these are on my clipboard history):

Screenshot 2025-02-19 at 20 54 13

You can barely see, but the "echo" item is selected in the left pane, and I've searched for `source` in the second:

Screenshot 2025-02-19 at 20 55 30

`Show source in catalog` has completely disappeared from my actions:

Screenshot 2025-02-19 at 20 56 20

n8henrie:
Still having lots of weird catalog issues with this branch.

My catalog pref pane running 2.4.3:

Screenshot 2025-02-21 at 18 33 40

vs from this branch:

Screenshot 2025-02-21 at 18 36 59

PR #3035, opened 2025-03-24
Codesign local

pjrobertson:
Quite simple really. You only ever need to run this script once (if you run it again, it'll generate a new cert and you'll need to setup the permissions again)

Run:

`bash Quicksilver/Tools/codesign/setup_cert.sh`


n8henrie:
Realizing I need to give this a shot, I'm unclear on why this works.

We are creating a cert named `Local Self-Signed`, trusting it, then using it to sign.

But the binary is changing with every build -- shouldn't macos re-prompt us if it's a different binary?

Just because I trust / accept permissions for one app by a specific developer shouldn't imply that I'm okay with *any* app by that developer, right? (Even if it claims to have the same bundleid.)

Also, why would we want this in CI? I thought it was just to skip all the extra permissions popups while developing locally?

pjrobertson:
Just working on this right now, it's still WIP. We don't really want the local cert in CI, but it was complaining that the certs weren't set up properly, and we're building the debug version on CI so...

Ideally, the debug version on CI would be signed using the proper developer cert as well. But I'm not sure how to change `CODE_SIGN_IDENTITY` based on `env` variable.

pjrobertson:
> But the binary is changing with every build -- shouldn't macos re-prompt us if it's a different binary?

I guess not. It's a different binary, but signed by the same cert so it's OK :) It's not a different app, it's the same app (bundle ID is the same)

pjrobertson:
@n8henrie   - you happy for this to be merged in as is for now, it's driving me crazy that I have to re-allow all the permissions every time I build QS locally.

n8henrie:
@pjrobertson -- XCode updated a day or two ago (Version 16.4 (16F6)), and now local signing is broken:

```
 /usr/bin/codesign --force --sign [REDACTED] --entitlements /Users/n8henrie/Library/Developer/Xcode/DerivedData/Quicksilver-ansgkardtljcsedwuirmtbxgmigg/Build/Intermediates.noindex/Quicksilver.build/Debug/Quicksilver.build/Quicksilver.app.xcent --timestamp\=none --generate-entitlement-der /Users/n8henrie/Library/Developer/Xcode/DerivedData/Quicksilver-ansgkardtljcsedwuirmtbxgmigg/Build/Products/Debug/Quicksilver.app

[REDACTED]: no identity found
Command CodeSign failed with a nonzero exit code
```

No luck with the usual clean / rebuild.

Found https://github.com/lwouis/alt-tab-macos/issues/4023, but I only have a single copy of `Local Self-Signed` in my Keychain, and it remains trusted.

Any issues for you?

pjrobertson:
I'm on Xcode 16.3, still works fine here. Haven't updated to 16.4 yet, but let me try it

pjrobertson:
Update: Works fine for me on Xcode 16.4 (build from within Xcode – are you using the build scripts?). Maybe try a restart of your computer?

n8henrie:
Weird. Yes, running from Xcode.

- No luck with xcode clean + `rm -rf /tmp/QS`
- No luck with restarting Xcoce
- No luck with reboot
- No luck with removing from keychain and re-running `setup_cert.sh` (appears to have worked, cert is trusted)

Same error, both in Xcode build log and when run directly in Terminal:

```
$ /usr/bin/codesign --force --sign [REDACTED] --timestamp\=none --preserve-metadata\=identifier,entitlements,flags --generate-entitlement-der /Users/n8henrie/Library/Developer/Xcode/DerivedData/Quicksilver-ansgkardtljcsedwuirmtbxgmigg/Build/Products/Debug/Quicksilver.app/Contents/Frameworks/QSEffects.framework/Versions/A
[REDACTED]: no identity found
```

Wait... just before submitting, I noticed that after the new `setup_cert.sh`, `security find-identity -p codesigning` was now showing a different `[REDACTED]` from above (not sure if this is actually worth redacting?); "full clean" and build seems to have resolved the issue!

pjrobertson:
If you can put this in the dev docs, that’d be useful! 
https://docs.qsapp.com/documentation/advanced-instrutions/building_quicksilver 
 → https://github.com/quicksilver/Documentation/blob/main/Building_Quicksilver.md

n8henrie:
Great idea! On the road for a week or so but will mark this down for my return.

n8henrie:
@pjrobertson this is not working for me again. I've tried removing all the signing certs and re-running the script. Still have to sign QS with every build.

Still working on your end?

pjrobertson:
Still working my end. It wasn't working today when I opened up the app, but then I re-ran the script and it started working

`bash Quicksilver/Tools/codesign/setup_cert.sh`

I added a commit to double check at build time to make life a bit easier. Managed to sneak it in here: https://github.com/quicksilver/Quicksilver/pull/3066/commits/7873c8ac53cc9c7935ecd1c1f33ed02e4ad4f60f

n8henrie:
Yeah, I keep getting the pop-up no matter what. ¯\\\_(ツ)_/¯

n8henrie:
I continue to be plague by issues here.

Today my builds started failing again with codesign errors.

With a little investigation, I found that my `CODE_SIGN_IDENTITY` somehow got a duplicate value (though this may have happened this morning while I was trying to sort out the problem, so might not be the cause).

```
$ git diff
Quicksilver/Quicksilver.xcodeproj/project.pbxproj

                                 CLANG_ENABLE_OBJC_ARC = YES;
                                 CODE_SIGN_ENTITLEMENTS = Quicksilver.entitlements;
                                 CODE_SIGN_IDENTITY = "Local Self-Signed";
+                                "CODE_SIGN_IDENTITY[sdk=macosx*]" = "-";
                                 DEVELOPMENT_TEAM = "";
                                 FRAMEWORK_SEARCH_PATHS = (
                                         "$(inherited)",
```

Resetting to main, `rm -rf /tmp/QS`, xcode clean, and rebuilding, I am met with an error in my build log:

```
/bin/sh -c /tmp/QS/build/Quicksilver.build/Debug/Quicksilver.build/Script-66CAE627153EF9AD0021BC65.sh  ERROR: Local Self-Signed certificate not found! Please run the following command to set up code signing:
```

This comes from https://github.com/quicksilver/Quicksilver/commit/7873c8ac53cc9c7935ecd1c1f33ed02e4ad4f60f, which adds the check

```console
$ security find-identity -v -p codesigning | grep -q "Local Self-Signed"
```

Sure enough, `security find-identity -v -p codesigning` finds two certificates, neither of which is the `Local Self-Signed`.

Where did it go? Why does this keep happening?

AHA! When I open up Keychain, I find `Local Self-Signed` is still there. However, it just expired:

Screenshot 2025-09-22 at 07 41 27

Seems odd, since I just went through this again recently. @pjrobertson have you not been having issues with yours expiring?

Command line version to get expiry ([source](https://apple.stackexchange.com/a/296682/27415)):

```
$ security find-certificate -c "Local Self-Signed" -p |
    openssl x509 -text |
    grep "Not After"
Not After : Sep 20 20:33:36 2025 GMT
```

Digging through a few SO threads leads to [this document](https://images.apple.com/certificateauthority/pdf/Developer_ID_CPS_v1.0.pdf) detailing the required elements for creation of the codesigning certificate, including `Validity period`.

I haven't found a way to extend the expiration of an existing self-signed certificate from the CLI, but as an alternative it seems that adding `-days` to one of the `openssl` calls should allow me to extend the expiration. Testing locally this seems to work and keychain confirms a much longer-lived certificate.

PR incoming.

Issue #3038, opened 2025-04-02
[Bug]: 2.5.0 won't launch


Calorion:
### Before submitting your bug report, please confirm you have completed the following steps

- [x] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [x] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [x] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

When I launch 2.5.0, nothing happens. It shows the system "opening" animation, and then…nothing. It doesn't show up in Activity Monitor. 

2.4.3 seems to work fine.

### Steps to reproduce

1. Double-click Quicksilver in the Finder.

### Expected behavior

It launches.

### MacOS Version

macOS 10.15.7

### Quicksilver Version

2.5.0

### Crash Logs or Spindump

[Quicksilver_2025-04-01-192100_Stephen-Byerley.txt](https://github.com/user-attachments/files/19594474/Quicksilver_2025-04-01-192100_Stephen-Byerley.txt)

n8henrie:
Huh, sorry to hear this!

Usually when log files are too long people paste them into a GitHub "gist" and then link to it in the issue. Can you do that?

Calorion:
> Usually when log files are too long people paste them into a GitHub "gist" and then link to it in the issue. Can you do that?

Easier to just copy and paste it to a TextEdit document. Silly that Console doesn't allow export to text.

Anyway, updated with the crash log.

n8henrie:
```
Dyld Error Message:
  Symbol not found: _CGPreflightScreenCaptureAccess
  Referenced from: /Applications/Quicksilver.app/Contents/MacOS/Quicksilver
  Expected in: /System/Library/Frameworks/CoreGraphics.framework/Versions/A/CoreGraphics
```

Looks like probably related to [this recent commit](https://github.com/quicksilver/Quicksilver/commit/88b441e8007707bc3d232293fa4f02fe08e454e3) 

It looks like it's properly guarded by `if (@available(macOS 10.15, *))`, and you're on 10.15. Any thoughts @pjrobertson?

PR #3042, opened 2025-04-05
Only check for screenshot permissions on macos11+

pjrobertson:
That's a strange one, since `CGPreflightScreenCaptureAccess` is explicitly listed as working on macos10.15+, but I've changed the implementation here slightly to use the `isXXX` methods, and then just use `isBigSur` instead of `isCatalina`.


n8henrie:
https://developer.apple.com/forums/thread/683860?answerId=684400022#684400022

pjrobertson:
This fixes #3038

PR #3055, opened 2025-05-31
Fix default bindings command string order

### LINKED ISSUE #3054: Grab and Drop does not work as expected
Long time user and I thought I would post about it. I tried searching but I don't think I have ever found any solid write-up about this feature. It worked in version 1.6.1, but the latest release still does not have this feature.

My usual workflow for activating this shortcut is as follows. Let's say I want to open a file in a different application than the default. I will click on the file in Finder, let's say a file on the desktop. I will then open Quicksilver and type in the name of my application. I will use TextEdit. Then I will press the keyboard shortcut, command-option-g and allow me to open the file on the desktop with TextEdit.

I believe [this YouTube video](https://www.youtube.com/watch?v=E0Nho0BtCi4) showcases the feature. Here is a [link to the documentation](https://github.com/quicksilver/Documentation/blob/main/Grab_'n_Drop.md) where it talks about this feature.

pjrobertson:
Fixes #3054

### DIFF
--- a/Quicksilver/Code-QuickStepInterface/DefaultBindings.qskeys
+++ b/Quicksilver/Code-QuickStepInterface/DefaultBindings.qskeys
@@ -54,19 +54,19 @@
 		<string>explodeCombinedObject</string>
 		<key>⌃⌘D</key>
 		<string>defineMnemonic:</string>
-		<key>⌃⌥⌘D</key>
+		<key>⌘⌥⌃D</key>
 		<string>removeMnemonic:</string>
 		<key>⌥,</key>
 		<string>uncollectLast:</string>
 		<key>⌥/</key>
 		<string>selectRoot:</string>
-		<key>⌥⌘G</key>
+		<key>⌘⌥G</key>
 		<string>dropSelection:</string>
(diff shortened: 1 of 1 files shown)


n8henrie:
Weird -- why does the order matter here, and how did you figure it out? Did you find some documentation somewhere?

pjrobertson:
I think probably at some point we  (I) updated NDHotkey which maybe change the orders that hotkey strings get written (I have a vague memory). The place to go for all these key bindings is that `DefaultBindings.qskeys`file. If you look at the top, the key is `QSSearchObjectView` which means they are key bindings for all `QSSearchObjectView`  (i.e. they work for `QSSearchObjectView` views. So then you go to `QSSearchObjectView.m` and find:

`- (BOOL)handleBoundKey:(NSEvent *)theEvent {`

In there, there's a check for `	NSString *selectorString = [bindingsDict objectForKey:theEventString];` which is doing a string comparison with the keys (keyboard shortcut strings like `⌘C` ). I debugged that and found `theEventString` was coming back in a different order, so I flipped it :)

PR #3056, opened 2025-06-08
Fix labels for cloud files (Google Drive, Dropbox etc.)

### LINKED ISSUE #2939: File's path appears as "iCloud" instead of it's real location, when browsing Google Drive files
I search for Google Drive and is path is something like /Library/CloudStorage/GoogleDrive...

As soon as I go inside that folder, the path changes to "iCloud" and remains like that for every file inside Google Drive.

MacOS Ventura

pjrobertson:
Fixes #2939


ernstki:
@pjrobertson Thanks for looping back around to this one, sir, it's appreciated.

PR #3063, opened 2025-08-21
Avoid QoS inversion when loading scripts

n8henrie:
@pjrobertson @skurfer does this look like the right way to resolve this warning?

Screenshot 2025-08-21 at 14 01 15

### DIFF
--- a/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSAppleScriptActions.m
+++ b/Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSAppleScriptActions.m
@@ -414,9 +414,12 @@
         int pid = [[NSProcessInfo processInfo] processIdentifier];
         NSAppleEventDescriptor* targetAddress = [[NSAppleEventDescriptor alloc] initWithDescriptorType:typeKernelProcessID bytes:&pid length:sizeof(pid)];
         
-        NSDictionary *errorDict = nil;
-        NSAppleScript *script = [[NSAppleScript alloc] initWithContentsOfURL:[NSURL fileURLWithPath:scriptPath] error:&errorDict];
+        NSDictionary __block *errorDict = nil;
+        NSAppleScript __block *script;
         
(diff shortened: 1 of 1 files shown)


n8henrie:
Apparently not, it initially seemed to resolve the warning but now it's back.

Screenshot 2025-08-21 at 14 17 22

pjrobertson:
I think this is kind of a non-issue. Are we actually seeing any issues because of this, or is it just xcode complaining?

The 'right' way to 'fix' (using 'fix' loosely, because I don't think this is broken) is to change the thread that `argumentCountForScript` is called on. Since we can't change the thread the `initWithContentsOfURL:` method is called on to make it higher, the only thing we can do is move our actual call 'down' to a lower priority thread.

Going all the way up the chain, the 'right' way to fix would therefore be to move the call to `[QSExec loadFileActions];` in QSController.m to be on a background thread. Think of it like this: Xcode is telling us that since we're doing file-related stuff, that could be slow, we shouldn't be doing it on the main thread. If we move `[QSExec loadFileActions];` to a background thread, then we won't block the UI.

Caveat: I haven't tested this / we may be doing UI stuff in `[QSExec loadFileActions];` somewhere, which would cause a crash unless it's put back on the main thread (with QSGCDMainSync)

n8henrie:
@pjrobertson thanks as always for your time and the explanation. This makes sense; I couldn't tell what part was running on the lower priority thread or why (as I didn't see any explicit GCD calls). Having IO be done in an async context makes sense though, as does the mountain of headaches with regards to cancellation, callbacks, etc.

No, this was just trying to fix an XCode warning, and trying to learn. Some success with the latter if nothing else ;)

pjrobertson:
No worries, this is actually something you could try out - moving `[QSExec loadFileActions]` to be run on a background thread with `QSGCDAsync` - it may have a side-benefit of speeding up QS launch, but again - may also have some unwanted side effects if we're actually doing main thread stuff there!

PR #3067, opened 2025-08-23
Use isMainThread to check if we're on the main thread

pjrobertson:
This is an alternative to using `dispatch_queue_get_label(queue) == dispatch_queue_get_label(DISPATCH_CURRENT_QUEUE_LABEL)` which is unreliable for the main thread. 

See here: https://stackoverflow.com/a/23957282/305324

Attached is the crash logs I've been seeing, FYI

[global-selection-crash.txt](https://github.com/user-attachments/files/21946881/global-selection-crash.txt)


pjrobertson:
Merging this to get a new release out

PR #3075, opened 2025-09-22
Extend the self-signed cert expiration to 10 years

n8henrie:
See also: https://github.com/quicksilver/Quicksilver/pull/3035#issuecomment-3319249830


pjrobertson:
LGTM!

PR #3076, opened 2025-09-27
Silence unhelpful log when object not found for self

n8henrie:
During long-running debug sessions my log fills up with
`No object found for process 0`.

I'm fairly confident that GetEventParameter(... kEventParamProcessID)
returning PSN of 0 refers to the current process (though having trouble
finding this in official Apple documentation).

This silences this log spam when running in debug mode.

Alternatively could map `0` to the Quicksilver PID when scanning for running
processes, but that would add a little overhead every time processes were
scanned and might be problematic if someone had multiple QS processes running
(for whatever reason).


pjrobertson:
LGTM. The error is actually a higher level problem (I think) with the fact that we now can no longer get info on the process because it's already been killed (I think the 'old' way of doing things with the carbon API meant we could still get the info).

But if processes are working, then this is all good!

PR #3080, opened 2025-10-22
Use `cp -r` instead of "Copy Files"

n8henrie:
The 'Copy Files' never properly let XCode know about `Quicksilver.pch`,
so we've required a separate "empty" step just for dependency resolution
for quite some time.

For some reason when building a Release build (with `Archive`), this
step wasn't working properly.

Instead, we can replace both steps with just a `cp -r` which seems to
serve both purposes and allows a Release / Archive build from scratch.


pjrobertson:
Build seems to be working fine on CI. Do you have a GH issue or relevant link for context on what this is solving?

n8henrie:
Try doing a local Product -> Archive, after cleaning everything (including /tmp/QS).

For me this consistently fails, /tmp/QS/Configuration is empty.

n8henrie:
Also, if you look at the contents of this step on current `main`, you'll see it's always been a bit of a kludge (the entire script is just a comment explaining why the empty script exists).

pjrobertson:
OK, sounds good to me!

PR #3090, opened 2026-01-30
Further tweaks to process checks

pjrobertson:
This adds a few extra tweaks to the previous PR. We don't *only* want to show NSBundle type apps (because command line apps are legitimate 'foreground apps' - especially if they have GUIs. But we also don't want to show background apps (esp the safari xpc apps). So I've added an additional check for XPC here and removed the previous check that *only* allowed NSBundles through.

| shows HelloQS | doesn't show safari XPC apps |
| --- | --- |
| Screen Region 2026-01-30 at 16 42 27 |  Screen Region 2026-01-30 at 16 42 51 |

### DIFF
--- a/Quicksilver/Code-QuickStepCore/QSProcessMonitor.m
+++ b/Quicksilver/Code-QuickStepCore/QSProcessMonitor.m
@@ -165,13 +165,12 @@ OSStatus appTerminated(EventHandlerCallRef nextHandler, EventRef theEvent, void 
 }
 
 - (QSObject *)imbuedFileProcessForDict:(NSDictionary *)dict {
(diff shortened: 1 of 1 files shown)


n8henrie:
Seems reasonable. It's possible someone out there is using QS to relaunch hung XPC processes or kill Microsoft / Chrome AutoUpdater or what have you, but seems pretty unlikely; I don't see any issues mentioning "xpc" that don't seem to be about crashes or hangs, so this seems like a net win.

If there are complaints, I imagine we could add a setting to reenable this, but I think the default should be off (80/20 rule).

Thanks as always for quick action!

pjrobertson:
Thanks! Sounds good. I think if people really want to relaunch XPC apps, then they can just enable 'show background processes'. No need for an extra setting ;-)

PR #3100, opened 2026-03-15
Update FMDB subproject to latest commit d3abf748

gliptak:
Correcting
https://github.com/quicksilver/Quicksilver/actions/runs/22161973552/job/64080631160
```
/Users/runner/work/Quicksilver/Quicksilver/Quicksilver/Code-External/FMDB/fmdb.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.11, but the range of supported deployment target versions is 10.13 to 15.5.99. (in target 'FMDB MacOS' from project 'fmdb')
```
https://github.com/ccgus/fmdb/pull/885


pjrobertson:
Thanks @gliptak for looking in to this! It looks like your updated code is still failing to pass CI.

The warning at the end of the logs was just a warning:
> /Users/runner/work/Quicksilver/Quicksilver/Quicksilver/Code-External/FMDB/fmdb.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.11, but the range of supported deployment target versions is 10.13 to 15.5.99. (in target 'FMDB MacOS' from project 'fmdb')

However the actual reason for the build error was caused by an encoding error in one of the newly added translation files:

> /Users/runner/work/Quicksilver/Quicksilver/Quicksilver/Localized/pt-BR.lproj/Localizable.strings:1:1: error: could not decode input file using specified encoding: Unicode (UTF-8), and the file contents appear to be encoded in Unicode (UTF-16) (in target 'Quicksilver' from project 'Quicksilver')

Xcode logs can be very cryptic sometimes. Opening the project in the Xcode UI can sometimes make it easier to debug.

This issue should now be fixed in: https://github.com/quicksilver/Quicksilver/commit/842de244137fb563abf46d85250c715031bf7c4d

gliptak:
thank you @pjrobertson might bringing FMDB current be still of value?

pjrobertson:
Good question @gliptak  - I think it might be, yes! Do you have a list of what's changed between the version we currently have and the version you provided? One risk I see is the mismatch in deployment target, did you try it and does it still work?

gliptak:
I'm not sure how to generate a meaningful list of changes between
-Subproject commit 61e51fde7f7aab6554f30ab061cc588b28a97d04
+Subproject commit d3abf748a2788471535993286603322ccdd02c3d

pjrobertson:
Is there anything in the FMDB [release notes?](https://github.com/ccgus/fmdb/releases) One way would be to look through the commit messages, but there may be a lot.

If you're up for it, one thing to do may be to pin the submodule to a git tag version instead of to a commit hash. E.g. we could pin to [v2.7.8](https://github.com/ccgus/fmdb/releases/tag/v2.7.8) (assuming that works properly with Quicksilver)

PR #3102, opened 2026-03-21
Add unit tests check to CI

pjrobertson:
I've just realised that CI wasn't running unit tests. Maybe they can't be run in a headless state... let's see

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,19 @@ on:
       - 'translations_**'
 
 jobs:
+  test:
+    runs-on: macos-latest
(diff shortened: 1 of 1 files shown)


pjrobertson:
Turns out qsrelease does run the tests. Ideally this should be split into its own job for clarity, but the way CI currently runs makes this a bit tricky. Closing for now - main thing is to ensure that CI is running tests

n8henrie:
I should have some time next week to look at splitting CI into something like build, test, and optionally release phases.

Yes the build process is a little funky -- both CLI and from within Xcode 😆

PR #3119, opened 2026-05-31
Do not block startup on optional permissions

krahd:
This changes the startup permission gate so Quicksilver does not become unusable when optional permissions are unavailable or misreported by macOS.

On Tahoe/macOS 26, Quicksilver can repeatedly show the permissions window even after permissions appear granted in System Settings. In the current implementation, normal startup is blocked when either Accessibility or Full Disk Access is not authorised, and the permissions window only auto-closes when Accessibility, Full Disk Access, Contacts, Calendars, Reminders, and Screenshot state are all satisfied.

This patch keeps Accessibility as the essential launch-blocking permission, but avoids blocking startup on Full Disk Access and lets the permissions window close once Accessibility is available. Optional permissions can still be requested from the permissions UI or preferences, but they should not prevent the app from launching.

This does not attempt to fix the separate Primer/interface file-descriptor issue.

### DIFF
--- a/Quicksilver/Code-App/QSController.m
+++ b/Quicksilver/Code-App/QSController.m
@@ -822,8 +822,7 @@ static QSController *defaultController = nil;
 	return;
 #endif
 
-	MPAuthorizationStatus fullDisk = [MPPermissionsKit authorizationStatusForPermissionType:MPPermissionTypeFullDiskAccess];
-	if(![self hasAccessibilityPermission] || fullDisk != MPAuthorizationStatusAuthorized) {
+	if(![self hasAccessibilityPermission]) {
 		if (![accessibilityPermissionWindow isVisible]) {
 			[self showAccessibilityPrompt:nil];
 			return;
@@ -857,9 +856,9 @@ static QSController *defaultController = nil;
 		[calendarsButton setEnabled:!hasCalendars || !hasReminders];
 		[screenshotButton setEnabled:!hasScreenshot];
 
-		[closeAccessibilityWindowButton setEnabled:(hasAccessibility && hasFullDisk)];
+		[closeAccessibilityWindowButton setEnabled:hasAccessibility];
 		
-		if (hasAccessibility && hasFullDisk && hasCalendars && hasContacts && hasReminders && hasClickedScreenshotButton) {
+		if (hasAccessibility) {
 				[accessibilityPermissionWindow close];
 		} else if (!accessibilityChecker) {
 				accessibilityChecker = [NSTimer scheduledTimerWithTimeInterval:0.5 repeats:YES block:^(NSTimer * _Nonnull timer) {


krahd:
Context from my Tahoe/macOS 26.5 setup:

- macOS 26.5
- Quicksilver 2.6.0, installed via Homebrew
- Accessibility and Full Disk Access were granted manually in System Settings
- Quicksilver still repeatedly showed the permissions prompt and did not proceed normally
- Clean uninstall/reinstall, reboot, quarantine removal, code-signature verification, and TCC resets did not resolve the prompt loop
- Code signature verified successfully with `codesign --verify --deep --strict --verbose=2 /Applications/Quicksilver.app`
- Bundle identifier was `com.blacktree.Quicksilver`
- A separate issue also appeared with Primer/interface loading and `errno=24` / too many open files; switching to Bezel appeared to reduce or avoid that separate problem

This PR is intended only to address the permission-gate/startup-blocking behaviour, not the separate Primer/file-descriptor issue.

n8henrie:
Hello -- thanks for your time. I'm not sure this behavior is desirable; I'd rather Quicksilver "fail loudly" when permissions aren't working; I imaging having it continue to half-function would lead to many more difficult-to-debug issues.

I think we should focus on getting your permissions issue fixed first. Would you mind opening an issue (which will help other users find this discussion in the expected place)?

Across a couple open-source MacOS projects I've worked with, the approaches you've listed above (thank you for trying so many!) have never worked for me. The solution here is usually to:

1. Leave the System Settings permission pane open
2. Quit Quicksilver
3. Use the `-` in the bottom left corner of the permissions preference pane to *remove* the existing Quicksilver entry entirely
4.  Re-open Quicksilver
5. Hit the toggle button to enable the permission in question

krahd:
Thanks, that makes sense. I agree this PR changes the failure policy rather than only fixing a mechanical bug, so it may not be the right upstream approach.

I will open a separate issue with the full Tahoe/macOS 26.5 reproduction details so the discussion is discoverable.

One important update from further local testing: my TCC database eventually showed both permissions as granted:

- `kTCCServiceAccessibility|com.blacktree.Quicksilver|0|2`
- `kTCCServiceSystemPolicyAllFiles|com.blacktree.Quicksilver|0|2`

The patched local build only worked once I built the `Quicksilver Distribution` scheme rather than the plain `Quicksilver` scheme; the plain scheme produced an incomplete app bundle lacking bundled plugins. With the distribution build and Bezel selected, Quicksilver opened.

I also agree that a better upstream fix here may be narrower than this PR: keep failing loudly, but make the permissions window/message distinguish Accessibility from Full Disk Access so users are not sent down the wrong debugging path.

I will open the issue and link it here.

krahd:
https://github.com/quicksilver/Quicksilver/issues/3121

PR #3120, opened 2026-05-31
Fix Bezel fallback after interface load failure

krahd:
Fixes a typo in the command-interface fallback path.

When the preferred command interface cannot be loaded, Quicksilver shows a notification saying it is switching to Bezel. The fallback code then asks the registry for `QSBezelnterfaceController`, which is missing the `I` in `Interface`.

That typo means the intended Bezel fallback can fail, leaving Quicksilver without a usable command interface after an interface/plugin load failure.

This changes the fallback key to `QSBezelInterfaceController`.

Context:
- This is relevant to recent macOS Tahoe reports where Quicksilver launches but has no usable interface, or shows “Interface could not be loaded. Switching to Bezel” and then remains crippled.
- This PR does not address file descriptor limits directly.
- This PR is separate from the optional-permissions startup gate PR.


krahd:
Context from my Tahoe/macOS 26.5 setup:

- Quicksilver 2.6.0 showed interface/plugin failure symptoms around Primer and Bezel
- Logs included interface/plugin load failures and `errno=24` / too many open files
- Switching manually to Bezel appeared to reduce or avoid the immediate interface failure
- The source appears to intend a Bezel fallback, but the key was misspelled as `QSBezelnterfaceController`

This PR only fixes the broken fallback key. It does not claim to solve the separate file-descriptor-limit problem directly.

n8henrie:
Great catch, thanks!

This seems like the kind of bug that a stronger type system (or perhaps a language that makes its type system easier to leverage to "make invalid states unrepresentable") would help catch. My kingdom for a Rust (or maybe even Swift) rewrite of QS's core!

n8henrie:
Thank you!

PR #3128, opened 2026-08-20
Show application locations matching Finder's behaviour

### LINKED ISSUE #3125: [Bug]: Safari is displayed using Cryptex backing path instead of /Applications/Safari.app

- [x] Searched the issue is not already listed in the [Issue List](https://github.com/quicksilver/Quicksilver/issues?q=is%3Aissue) (I've searched it)
- [x] Searched in the [Quicksilver Users' Group](https://groups.google.com/forum/#!forum/blacktree-quicksilver) for questions relating to tech support.
- [x] Downloaded the [latest version of Quicksilver](https://qsapp.com/) for your system and confirmed the issue still persists.

### Bug description

When searching for Safari, it shows its base location at `/System/Volumes/Preboot/Cryptexes/App/System/Applications/Safari.app` rather than the canonical/overlay location in `/Applications`.

### Steps to reproduce

1. Search for Safari in Quicksilver
2. View the path returned

### Expected behavior

Safari is shown with a location at /Applications

### MacOS Version

26.5.1

### Quicksilver Version

2.6.0

### Relevant Plugins

_No response_

### Crash Logs or Spindump

_No response_

### Screenshots

Image

### Additional info

There is a similar issue with apps like "Find My" (/System/Applications/Find My.app), Music, etc.

Related to https://github.com/quicksilver/Quicksilver/issues/3082.

danielcompton:
macOS shows applications in the /Applications folder that may not actually exist there on disk.

There are two cases:
1. Cryptex backed applications like Safari - these are in a path like /System/Volumes/Preboot/Cryptexes/App/System/Applications/Safari.app. This path isn't genuinely visible to users. /Applications appears to contain Safari.app but it is just a symlink to the cryptex.

```console
$ ls -l /Applications/Safari.app
lrwxr-xr-x@ 1 root  wheel  54 23 Jul 09:57 /Applications/Safari.app@ -> ../System/Cryptexes/App/System/Applications/Safari.app
$ ls -l /System/Cryptexes/App/System/Applications/Safari.app
total 0
drwxr-xr-x  11 root  wheel  352 16 Jul 18:22 Contents/
```

2. System applications like Find My - these are in a path like /System/Applications/FindMy.app. There is no symlink to /Applications, but Finder will display these system applications in /Applications.

E.g.

```console
$ ls -l /Applications/FindMy.app
ls: /Applications/FindMy.app: No such file or directory
$ ls -l /System/Applications/FindMy.app
total 0
drwxr-xr-x  11 root  wheel  352 23 Jul 09:57 Contents/
```

This PR fixes both of these issues. It merges applications found at multiple locations, e.g. the symlinked /Applications/Safari.app + Cryptex Safari.app if they both resolve to the same user visible location.

See https://developer.apple.com/forums/thread/745673 for more discussion on cryptexes.

**Changes**

- **Add NSURL category for user-visible paths**
- **Catalog apps at their user-visible path instead of cryptex backing path**
- **Show system applications at their Finder location**
- **Document the Additional Applications preset as a cryptex fallback**

Fixes #3125

I checked how Alfred handles this and it appears to just use hardcoded string comparisons to replace certain applications with the user visible path. I think these changes are more principled and correct.

### DIFF
--- a/Quicksilver/Code-QuickStepFoundation/NSURL_BLTRExtensions.m
+++ b/Quicksilver/Code-QuickStepFoundation/NSURL_BLTRExtensions.m
@@ -101,6 +101,81 @@ NSString *QSPasswordForHostUserType(NSString *host, NSString *user, SecProtocolT
 
 @end
 
+/* The system domain's Applications directories in every path form: as reported
+ * by NSSearchPathForDirectoriesInDomains, plus their symlink-resolved forms
+ * (e.g. /System/Cryptexes/App is a symlink to /System/Volumes/Preboot/Cryptexes/App,
+ * the form directory enumerators report) */
+static NSArray *QSSystemApplicationsDirectories(void) {
+    static NSArray *directories = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        NSMutableSet *set = [NSMutableSet set];
+        for (NSString *dir in NSSearchPathForDirectoriesInDomains(NSApplicationDirectory, NSSystemDomainMask, YES)) {
+            [set addObject:dir];
+            [set addObject:[dir stringByResolvingSymlinksInPath]];
+        }
+        directories = [set allObjects];
+    });
+    return directories;
+}
+
+/* YES if any directory between root and the last path component is a bundle */
+static BOOL QSPathHasBundleAncestor(NSString *root, NSArray *components, NSUInteger fromIndex) {
+    NSString *ancestor = root;
+    for (NSUInteger i = fromIndex; i + 1 < [components count]; i++) {
+        ancestor = [ancestor stringByAppendingPathComponent:components[i]];
+        NSNumber *isPackage = nil;
+        [[NSURL fileURLWithPath:ancestor] getResourceValue:&isPackage forKey:NSURLIsPackageKey error:NULL];
+        if ([isPackage boolValue]) return YES;
+    }
+    return NO;
+}
+
+@implementation NSURL (QSCanonicalPath)
+
+- (NSURL *)URLByMappingSystemApplicationsToLocalDomain {
+    if (![self isFileURL]) return self;
+
+    NSString *path = [self path];
+    NSArray *components = nil;
+    for (NSString *systemApps in QSSystemApplicationsDirectories()) {
+        if (![path hasPrefix:systemApps]) continue;
+        if (!components) components = [path pathComponents];
+
+        // Must be a descendant on a component boundary: /System/ApplicationsFoo
+        // shares the prefix string but isn't inside /System/Applications
+        NSArray *systemAppsComponents = [systemApps pathComponents];
+        NSUInteger prefixCount = [systemAppsComponents count];
+        if ([components count] <= prefixCount) continue;
+        if (![[components subarrayWithRange:NSMakeRange(0, prefixCount)] isEqualToArray:systemAppsComponents]) continue;
+
+        // Finder only merges the folder hierarchy; paths inside a bundle are not remapped
+        if (QSPathHasBundleAncestor(systemApps, components, prefixCount)) return self;
+
+        NSString *localApps = [NSSearchPathForDirectoriesInDomains(NSApplicationDirectory, NSLocalDomainMask, YES) firstObject];
+        NSArray *relativeComponents = [components subarrayWithRange:NSMakeRange(prefixCount, [components count] - prefixCount)];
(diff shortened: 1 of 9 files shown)


Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSDirectoryParser.m
```diff
@@ -138,7 +139,15 @@ - (NSArray *)objectsFromPath:(NSString *)path depth:(NSInteger)depth types:(NSAr
...
         if (include) {
-            QSObject *obj = [QSObject fileObjectWithFileURL:theURL];
+            NSURL *objectURL = theURL;
+            if (QSTypeConformsTo(type, (NSString *)kUTTypeApplication)) {
+                // Applications can be scanned via a cryptex backing path
+                // (e.g. /System/Cryptexes/App/System/Applications/Safari.app);
+                // catalog them at their user-visible path so every source
+                // yields the same object (#3125)
+                objectURL = [objectURL URLByResolvingToUserVisiblePath];
+            }
```
skurfer, line 149: Is there any need to load this object from the user-visible path? It seems like getting it using the actual path would be fine (if not preferable). We’re really just concerned with what happens in the UI, and that should be handled by the changes in `detailsOfObject:`, right?

Quicksilver/PlugIns-Main/QSCorePlugIn/QSCorePlugIn-Info.plist
```diff
@@ -50,6 +50,11 @@
 					<key>name</key>
 					<string>Applications</string>
 				</dict>
+				<!-- The App cryptex's Applications directory. Apps here (currently
+				     just Safari) also have an /Applications symlink and are picked up
+				     by the Applications preset above, and both scans produce the same
+				     object, so this is a fallback for a future cryptex app shipped
+				     without a symlink. See issues #2932 and #3125. -->
```
skurfer, line 57: This comment is fine, just be aware it will probably get blown away if we ever edit this file in Xcode.

Quicksilver/Code-QuickStepFoundation/NSURL_BLTRExtensions.m
```diff
@@ -101,6 +101,81 @@ - (NSURL *)URLByReallyResolvingSymlinksInPath {
...
 
+/* The system domain's Applications directories in every path form: as reported
+ * by NSSearchPathForDirectoriesInDomains, plus their symlink-resolved forms
+ * (e.g. /System/Cryptexes/App is a symlink to /System/Volumes/Preboot/Cryptexes/App,
+ * the form directory enumerators report) */
+static NSArray *QSSystemApplicationsDirectories(void) {
+    static NSArray *directories = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        NSMutableOrderedSet *set = [NSMutableOrderedSet orderedSet];
```
skurfer: Is there a reason for this to be ordered? (I think regular sets are lighter weight, but we’re obviously not talking about a huge collection here, so not a big deal either way. Just curious.)

skurfer (commented):
Looks good, but I had a couple of questions.

Quicksilver/Code-QuickStepFoundation/NSURL_BLTRExtensions.m
danielcompton: No real reason. Claude seemed to want to do it this way for debugging, but I agree that a regular set is a better choice. Fixed now.

Quicksilver/PlugIns-Main/QSCorePlugIn/Code/QSDirectoryParser.m
danielcompton: I could see arguments either way and am not stuck on doing it this way, but if we did change it to make the object URL the base one then things like "Get File Location" will return `Preboot:Cryptexes:App:System:Applications:Safari.app` instead of `Applications:Safari.app` and similarly for FindMy.app.

But yes, I'm mostly concerned with the UI display of the path, so happy to change it if you'd like?

skurfer: Oh, we want those actions to return the user-friendly path, so I’m sold on leaving this in.

skurfer:
I tested this out locally and pulled up Safari. The details look correct, but the **Get Path** action doesn’t actually grab the user-friendly path. Does that work on your end?

PR #3132, opened 2026-09-18
MACOSX_DEPLOYMENT_TARGET -> 12

n8henrie:
Allows building on macos27


skurfer:
I think we just need to change line 13 in the Common config and all those targets will pull from that.

n8henrie:
Will try that this evening

n8henrie:
That works for QS an our deps, but isn't inherited by e.g. FMDB, which is still a somewhat active project but ironically its most recent commit (6m ago) was to pin this exact setting to even *older* than ours: <https://github.com/ccgus/fmdb/commit/00d6a39fdc7045bd0d6fc8fac6e76f840a5b3a1e>

